PR #12, opened 2021-05-03
Adds support for "as" usage in var declarations

### LINKED ISSUE #9: Improper "as" handling
TG/Bee contains the following code:
```
/procpath
	// Although these variables are effectively const, if they are marked const
	// below, their accesses are optimized away.

	/// A text string of the verb's name.
	var/name as text
	/// The verb's help text or description.
	var/desc as text
	/// The category or tab the verb will appear in.
	var/category as text
	/// Only clients/mobs with `see_invisibility` higher can use the verb.
	var/invisibility as num
```

It causes an "Expected dedent" exception unless you remove every `as (whatever)`. While the usage of `as` is pointless, the BYOND compiler doesn't care about its existence.

ike709:
Fixes #9 

AFAIK the BYOND compiler just ignores this.

### DIFF
--- a/OpenDreamShared/Compiler/DM/DMParser.cs
+++ b/OpenDreamShared/Compiler/DM/DMParser.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using OpenDreamShared.Dream;
 using DereferenceType = OpenDreamShared.Compiler.DM.DMASTDereference.DereferenceType;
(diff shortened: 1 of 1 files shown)


wixoaGit (changes requested):
This doesn't correctly handle more complex types (like `as num|text|file`). It should be replaced with a call to `AsTypes`

ike709:
Fixed.

wixoaGit (approved):

PR #25, opened 2021-05-09
Ignore spaces after multiline comments

### LINKED ISSUE #11: Expected EOF exception in DMParser
```
Unhandled exception. System.Exception: Expected EOF
   at OpenDreamShared.Compiler.Parser.Consume(TokenType type, String errorMessage) in C:\Users\Ike\Documents\GitHub\OpenDream\OpenDream\OpenDreamShared\Compiler\Parser.cs:line 62
   at OpenDreamShared.Compiler.DM.DMParser.File() in C:\Users\Ike\Documents\GitHub\OpenDream\OpenDream\OpenDreamShared\Compiler\DM\DMParser.cs:line 19
   at DMCompiler.Program.Main(String[] args) in C:\Users\Ike\Documents\GitHub\OpenDream\OpenDream\DMCompiler\Program.cs:line 38
```
Tested the compiler against BeeStation after doing a placeholder `#if` implementation (just `SkipIfBody()`) and fixing the other open issues. Can't figure out which file is the issue after poking around with the debugger.

ike709:
Fixes #11 

Not sure if there's other chars I should add.

### DIFF
--- a/OpenDreamShared/Compiler/DMPreprocessor/DMPreprocessorLexer.cs
+++ b/OpenDreamShared/Compiler/DMPreprocessor/DMPreprocessorLexer.cs
@@ -63,7 +63,7 @@ namespace OpenDreamShared.Compiler.DMPreprocessor {
                                     {
                                         Advance();
                                     }
-                                    if(!AtEndOfSource) break;
+                                    break;
                                 }
                                 else if (AtEndOfSource) throw new Exception("Expected \"*/\" to end multiline comment");
                                 else if (!isStar) Advance();
--- a/OpenDreamShared/Compiler/DMPreprocessor/DMPreprocessorLexer.cs
+++ b/OpenDreamShared/Compiler/DMPreprocessor/DMPreprocessorLexer.cs
@@ -56,7 +56,15 @@ namespace OpenDreamShared.Compiler.DMPreprocessor {
                             while (true) {
                                 bool isStar = GetCurrent() == '*';
 
-                                if (isStar && Advance() == '/') break;
+                                if (isStar && Advance() == '/')
+                                {
+                                    Advance();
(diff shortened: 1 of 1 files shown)


wixoaGit (changes requested):
Tabs should also be ignored

ike709:
Done.

OpenDreamShared/Compiler/DMPreprocessor/DMPreprocessorLexer.cs
```diff
@@ -56,7 +56,15 @@ class DMPreprocessorLexer : TextLexer {
...
 
-                                if (isStar && Advance() == '/') break;
+                                if (isStar && Advance() == '/')
+                                {
+                                    Advance();
+                                    while (GetCurrent() == ' ' || GetCurrent() == '\t')
+                                    {
+                                        Advance();
+                                    }
+                                    if(!AtEndOfSource) break;
```
wixoaGit, line 66: Wouldn't we want to always break, regardless of if we're at the end of the source?

ike709: Wouldn't that prevent the exception from throwing because the `else if(AtEndOfSource)` doesn't fire?

wixoaGit: At this point the comment already ended, so it shouldn't be throwing.

ike709: An excellent point.

ike709, line 66:
```suggestion
                                    break;
```

PR #26, opened 2021-05-09
Adds support for as()

ike709:
Doing 
```
for(var/item/thing/T as() in some_list)
``` 
is the same as 
```
for(var/item/thing/T as anything in some_list)
```

Both define a loop without internal typechecking.

### DIFF
--- a/OpenDreamShared/Compiler/DM/DMParser.cs
+++ b/OpenDreamShared/Compiler/DM/DMParser.cs
@@ -356,8 +356,6 @@ namespace OpenDreamShared.Compiler.DM {
                 if (procStatement == null) procStatement = While();
                 if (procStatement == null) procStatement = DoWhile();
                 if (procStatement == null) procStatement = Switch();
-                Whitespace();
-                Check(TokenType.Skip);
 
                 if (procStatement != null)
                 {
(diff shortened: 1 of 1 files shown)


ZeWaka (approved):

OpenDreamShared/Compiler/DM/DMParser.cs
```diff
@@ -1633,8 +1633,13 @@ class DMParser : Parser<Token> {
...
 
             if (Check(TokenType.DM_As)) {
+                if (Check(TokenType.DM_LeftParenthesis) && Check(TokenType.DM_RightParenthesis))
+                {
+                    return DMValueType.Anything;
+                }
+                
                 Whitespace();
-
+
```
ike709, line 1642:
```suggestion
```

OpenDreamShared/Compiler/DM/DMParser.cs
```diff
@@ -1633,8 +1633,12 @@ class DMParser : Parser<Token> {
             DMValueType type = DMValueType.Anything;
 
             if (Check(TokenType.DM_As)) {
+                if (Check(TokenType.DM_LeftParenthesis) && Check(TokenType.DM_RightParenthesis))
```
wixoaGit, line 1636: Whitespace needs to be checked for before the left parenthesis, and `Consume` should be used for the right parenthesis

wixoaGit (changes requested):
This works for `as()` but doesn't work if you give it any types, like `as(num|text)`.

ike709:
Didn't know you could insert types like that, should be good now.

ike709:
This works against beestation. `as()` and `as(text|num)` are valid but `as(` errors as expected. `as text|num` seems unaffected

wixoaGit (approved):

PR #41, opened 2021-05-11
Small compiler tweaks

ZeWaka:
each commit details the tweaks, just small stuff

* Fixed typo
* Did some null propagation and checks
* Index from end expressions vs manual length decrease

### DIFF
--- a/DMCompiler/DM/DMProc.cs
+++ b/DMCompiler/DM/DMProc.cs
@@ -243,7 +243,9 @@ namespace DMCompiler.DM {
                     _bytecodeWriter.Write((byte)parameterType);
 
                     if (parameterType == DreamProcOpcodeParameterType.Named) {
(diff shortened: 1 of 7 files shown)


DMCompiler/DM/DMProc.cs
```diff
@@ -243,7 +243,7 @@ private class DMProcScope {
                     _bytecodeWriter.Write((byte)parameterType);
 
                     if (parameterType == DreamProcOpcodeParameterType.Named) {
-                        WriteString(parameterNames[namedParameterIndex++]);
+                        WriteString(parameterNames?[namedParameterIndex++]);
```
wixoaGit, line 246: I would prefer an exception be thrown if `parameterNames` is null

wixoaGit (changes requested):

wixoaGit (approved):

PR #48, opened 2021-05-13
Prompt window Title setting

ZeWaka:
If not set, defaults to `OpenDream` for the Titles.

~~Still trying to figure out how to implement it on `Prompt`'s creators.~~

image

### DIFF
--- a/OpenDreamServer/Net/DreamConnection.cs
+++ b/OpenDreamServer/Net/DreamConnection.cs
@@ -209,7 +209,7 @@ namespace OpenDreamServer.Net {
                     promptWaitHandle.Set();
                 };
 
-                SendPacket(new PacketPrompt(promptId, types, "DreamOpen", message));
+                SendPacket(new PacketPrompt(promptId, types, "NonNull", message));
                 promptWaitHandle.WaitOne();
                 return promptResponse;
             });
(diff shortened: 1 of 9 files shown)


OpenDreamServer/Net/DreamConnection.cs
```diff
@@ -209,7 +209,7 @@ class DreamConnection {
                     promptWaitHandle.Set();
                 };
 
-                SendPacket(new PacketPrompt(promptId, types, message));
+                SendPacket(new PacketPrompt(promptId, types, "NonNull", message));
```
ZeWaka, line 212: This is specifically where I'm having trouble - how to get the title set to this point?

wixoaGit: `input()` works using an opcode called `Prompt`. If you want to change its behavior, you do that [here](https://github.com/wixoaGit/OpenDream/blob/55efe282984927cc27981525a715cbbf2eeb2722/OpenDreamServer/Dream/Procs/DreamProcInterpreterOpcodes.cs#L1087).

Line 1094 is where it tries to understand arguments `input()` was given. Since title is the third argument if you give the recipient, and the second if not, you'll want to do something like
`title = arguments.OrderedArguments[2].GetValueAsString();` in the if, and
`title = arguments.OrderedArguments[1].GetValueAsString();` in the else.

Once you get the title from the arguments you can pass this to the Prompt method on line 1104.

TestGame/code.dm
```diff
@@ -21,6 +21,11 @@
 
 		for (var/mob/viewer in viewers)
 			viewer << "[ckey] says: \"[message]\""
+	
+	verb/say_loud()
+		var/msg = input("Please put the message you want to say loudly.", "Say Loud")
+		for (var/mob/M in viewers())
```
wixoaGit, line 27: Maybe instead do `world << ...` to differentiate between say and say_loudly?

wixoaGit (changes requested):

PR #55, opened 2021-05-14
Adds world.address/port/url implementation

ZeWaka:
No setting yet - plan to do setting of port/address via DreamServer arguments in separate PR

### DIFF
--- a/OpenDreamServer/Dream/Objects/MetaObjects/DreamMetaObjectWorld.cs
+++ b/OpenDreamServer/Dream/Objects/MetaObjects/DreamMetaObjectWorld.cs
@@ -18,7 +18,7 @@ namespace OpenDreamServer.Dream.Objects.MetaObjects {
             if (fps.Value != null) {
                 dreamObject.SetVariable("tick_lag", new DreamValue(10.0f / fps.GetValueAsInteger()));
             }
-            
+
             //New() is not called here
         }
 
@@ -59,7 +59,7 @@ namespace OpenDreamServer.Dream.Objects.MetaObjects {
             } else if (variableName == "maxz") {
                 return new DreamValue(Program.DreamMap.Levels.Count);
             } else if (variableName == "address") {
-                return DreamValue.Null;
+                return new(Program.DreamServer.Address.ToString());
             } else if (variableName == "system_type") {
                 //system_type value should match the defines in Defines.dm
                 if (Environment.OSVersion.Platform is PlatformID.Unix or PlatformID.MacOSX or PlatformID.Other) {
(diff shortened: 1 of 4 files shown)


wixoaGit (commented):
`world.address` seems to always be 0.0.0.0, while in byond it's my computer's local IP address.

ZeWaka:
> the reason it's showing up as`0.0.0.0` is because `IPAddress.Any` which the default set, goes to `0.0.0.0` instead of BYOND, which finds a different one - ex. for me `172.20.48.0`

I think this is correct. It's supposed to get your local IP, and the world is being hosted on `0.0.0.0` by default currently.

Also, `world.internet_address` will be a pain, since we'll have to open a socket to an external server (say 8.8.8.8 or something).

OpenDreamServer/Dream/Objects/MetaObjects/DreamMetaObjectWorld.cs
```diff
@@ -59,7 +59,11 @@ class DreamMetaObjectWorld : DreamMetaObjectRoot {
             } else if (variableName == "maxz") {
                 return new DreamValue(Program.DreamMap.Levels.Count);
             } else if (variableName == "address") {
-                return DreamValue.Null;
+                return new(Program.DreamServer.Address.ToString());
+            } else if (variableName == "port") {
+                return new(Program.DreamServer.Port.ToString());
+            } else if (variableName == "url") {
+                return new("byond://" + Program.DreamServer.Address + ":" + Program.DreamServer.Port);
```
wixoaGit, line 66: I'm hesitant to have this return a `byond://` url

OpenDreamServer/Dream/Objects/MetaObjects/DreamMetaObjectWorld.cs
```diff
@@ -59,7 +59,11 @@ class DreamMetaObjectWorld : DreamMetaObjectRoot {
             } else if (variableName == "maxz") {
                 return new DreamValue(Program.DreamMap.Levels.Count);
             } else if (variableName == "address") {
-                return DreamValue.Null;
+                return new(Program.DreamServer.Address.ToString());
+            } else if (variableName == "port") {
+                return new(Program.DreamServer.Port.ToString());
```
wixoaGit, line 64: `/world.port` is a number

wixoaGit (changes requested):

OpenDreamServer/Dream/Objects/MetaObjects/DreamMetaObjectWorld.cs
ZeWaka: fair, was just replicating existing behavior - we should probably specify our own thing

OpenDreamServer/Dream/Objects/MetaObjects/DreamMetaObjectWorld.cs
ZeWaka: 😹

wixoaGit (approved):

PR #56, opened 2021-05-14
Implementation of z-level removal (`world.maxz--`)

ZeWaka:
Unsure on exact GC behavior regarding the deletion of the levels, please advise.

The actual z-removal works though. You're no longer able to traverse to the level in the test project.

If you are on a Z-level when it gets removed, your client just stays stuck and can't move. This is similar to BYOND's behavior, where it just crashes your client.

### DIFF
--- a/OpenDreamRuntime/DreamRuntime.cs
+++ b/OpenDreamRuntime/DreamRuntime.cs
@@ -123,6 +123,7 @@ namespace OpenDreamRuntime
 
             Map = new DreamMap(this);
             Map.LoadMap(CompiledJson.Maps[0]);
+            WorldInstance.SetVariable("maxz", new DreamValue(Map.Levels.Count));
 
             WorldInstance.SpawnProc("New");
         }
--- a/OpenDreamRuntime/DreamRuntime.cs
+++ b/OpenDreamRuntime/DreamRuntime.cs
@@ -123,8 +123,7 @@ namespace OpenDreamRuntime
 
             Map = new DreamMap(this);
             Map.LoadMap(CompiledJson.Maps[0]);
-            WorldInstance.SetVariable("maxz", new DreamValue(Map.Levels.Count));
-
+            
             WorldInstance.SpawnProc("New");
         }
 
--- a/OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectWorld.cs
+++ b/OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectWorld.cs
@@ -32,7 +32,7 @@ namespace OpenDreamRuntime.Objects.MetaObjects {
             } else if (variableName == "maxz") {
                 int newMaxZ = variableValue.GetValueAsInteger();
 
-                if (newMaxZ < oldVariableValue.GetValueAsInteger()) {
+                if (newMaxZ < Runtime.Map.Levels.Count) {
                     while (Runtime.Map.Levels.Count > newMaxZ) {
                         Runtime.Map.RemoveLevel();
                     }
--- a/OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectWorld.cs
+++ b/OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectWorld.cs
@@ -32,9 +32,15 @@ namespace OpenDreamRuntime.Objects.MetaObjects {
             } else if (variableName == "maxz") {
                 int newMaxZ = variableValue.GetValueAsInteger();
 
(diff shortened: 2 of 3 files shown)


TestGame/code.dm
```diff
@@ -15,6 +15,7 @@
 
 	verb/tell_location()
 		usr << "You are at ([x], [y], [z])"
+		world.maxz--
```
ZeWaka, line 18: Temporary.

ZeWaka:
Looks like `world.maxz` isn't getting set at all @ world init.

OpenDreamServer/Dream/DreamMap.cs
```diff
@@ -99,7 +99,10 @@ public class MapLevel {
 
         public void RemoveLevel() {
             MapLevel toRemove = Levels[^1];
-            toRemove.Cells = null;
+            foreach (MapCell cell in toRemove.Cells) {
+                cell.Turf.Delete();
+                cell.Area.Delete();
```
wixoaGit, line 102: This will delete an area, even if it still contains other turfs or is referenced somewhere in the DM code.

wixoaGit (changes requested):

OpenDreamServer/Dream/DreamMap.cs
ZeWaka: Here's what I have - can't find a way to compensate for higher z-levels though: 
```cs
public void RemoveLevel() {
    int idx = Levels.Count - 1;
    MapLevel toRemove = Levels[idx];
    foreach (MapCell cell in toRemove.Cells) {
        cell.Turf.Delete();
        // Area.z represents the lowest Z level a comprised turf is on, therefore we can delete if it's >= our idx
        // DOES NOT CONSIDER TURFS ON A HIGHER Z-LEVEL!
        if (cell.Area.GetVariable("z").GetValueAsInteger() >= idx) 
            cell.Area.Delete();
    }
    Levels.Remove(toRemove);
}
```

wixoaGit: Using the following code to test, it doesn't appear that the area _should_ be deleted after all of its turfs are gone. Its `x` `y` and `z` become null, but that's not something we can replicate without keeping track of an area's turfs, which we don't currently do.

```dm
world.maxz++

var/area/newarea = new /area/test()
newarea.contents += locate(1, 1, world.maxz)

world << "[newarea], [newarea.x], [newarea.y], [newarea.z]"
world.maxz--
world << "[newarea], [newarea.x], [newarea.y], [newarea.z]"
```

Output:
```
The test, 1, 1, 2
The test, , , 
```

OpenDreamRuntime/DreamRuntime.cs
```diff
@@ -123,6 +123,7 @@ public class DreamRuntime
 
             Map = new DreamMap(this);
             Map.LoadMap(CompiledJson.Maps[0]);
+            WorldInstance.SetVariable("maxz", new DreamValue(Map.Levels.Count));
```
wixoaGit, line 126: Why is this needed?

OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectWorld.cs
```diff
@@ -32,9 +32,15 @@ public DreamMetaObjectWorld(DreamRuntime runtime)
             } else if (variableName == "maxz") {
                 int newMaxZ = variableValue.GetValueAsInteger();
 
-                if (newMaxZ < oldVariableValue.GetValueAsInteger()) throw new NotImplementedException("Cannot set maxz lower than previous value");
-
-                while (Runtime.Map.Levels.Count < newMaxZ) Runtime.Map.AddLevel();
+                if (newMaxZ < oldVariableValue.GetValueAsInteger()) {
```
wixoaGit, line 35: Use `Runtime.Map.Levels.Count` instead of `oldVariableValue`
```suggestion
                if (newMaxZ < Runtime.Map.Levels.Count) {
```

wixoaGit (changes requested):

OpenDreamRuntime/DreamRuntime.cs
ZeWaka: ...To set the value of the variable?

wixoaGit: It shouldn't need to be set, `DreamMetaObjectWorld`'s `OnVariableGet` handles `maxz`

ZeWaka: I completely forgot the special variable handlers were a thing 🤦‍♂️

wixoaGit (approved):

PR #59, opened 2021-05-14
Implement world.system_type

ike709:
http://www.byond.com/docs/ref/#/world/var/system_type

It checks for unix/macos rather than windows because the enum contains like 4 different entries for windows.

### DIFF
--- a/OpenDreamServer/Dream/Objects/MetaObjects/DreamMetaObjectWorld.cs
+++ b/OpenDreamServer/Dream/Objects/MetaObjects/DreamMetaObjectWorld.cs
@@ -19,15 +19,6 @@ namespace OpenDreamServer.Dream.Objects.MetaObjects {
                 dreamObject.SetVariable("tick_lag", new DreamValue(10.0f / fps.GetValueAsInteger()));
             }
             
-            //system_type value should match the defines in Defines.dm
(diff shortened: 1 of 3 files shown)


ZeWaka (approved):

ZeWaka:
lmao 
image

OpenDreamServer/Dream/Objects/MetaObjects/DreamMetaObjectWorld.cs
```diff
@@ -18,6 +18,15 @@ class DreamMetaObjectWorld : DreamMetaObjectRoot {
             if (fps.Value != null) {
                 dreamObject.SetVariable("tick_lag", new DreamValue(10.0f / fps.GetValueAsInteger()));
             }
+            
+            //system_type value should match the defines in Defines.dm
+            if (Environment.OSVersion.Platform is PlatformID.Unix or PlatformID.MacOSX or PlatformID.Other) {
```
wixoaGit, line 23: Setting the variable when creating the world object will end up leaving `system_type` writable instead of constant. Moving this to `OnVariableGet` would cause this to always return the same value, even after attempting to set it.

wixoaGit (changes requested):

ike709:
Done.

wixoaGit (approved):

PR #60, opened 2021-05-14
Implement goto and label in DMVisitorProcBuilder and DMProc

ike709:
**Disclaimer: I have no idea if this works.**

I have not actually tested this with sample DM code.

### DIFF
--- a/DMCompiler/DM/DMProc.cs
+++ b/DMCompiler/DM/DMProc.cs
@@ -225,6 +225,10 @@ namespace DMCompiler.DM {
         public void ContinueIfFalse() {
             JumpIfFalse(_loopStack.Peek() + "_continue");
         }
+        
+        public void Goto(string label) {
+            Jump(label + "_goto");
+        }
 
         public void PushProcArguments() {
             WriteOpcode(DreamProcOpcode.PushProcArguments);
--- a/DMCompiler/DM/DMProc.cs
+++ b/DMCompiler/DM/DMProc.cs
@@ -227,7 +227,7 @@ namespace DMCompiler.DM {
         }
         
         public void Goto(string label) {
-            Jump(label);
+            Jump(label + "_codelabel");
         }
 
         public void PushProcArguments() {
--- a/DMCompiler/DM/DMProc.cs
+++ b/DMCompiler/DM/DMProc.cs
@@ -227,7 +227,7 @@ namespace DMCompiler.DM {
         }
         
         public void Goto(string label) {
-            Jump(label + "_goto");
+            Jump(label);
         }
 
         public void PushProcArguments() {
--- a/DMCompiler/DM/Visitors/DMVisitorProcBuilder.cs
+++ b/DMCompiler/DM/Visitors/DMVisitorProcBuilder.cs
@@ -64,7 +64,7 @@ namespace DMCompiler.DM.Visitors {
         }
 
         public void VisitProcStatementLabel(DMASTProcStatementLabel statementLabel) {
-            _proc.AddLabel(statementLabel.Name);
+            _proc.AddLabel(statementLabel.Name + "_codelabel");
         }
 
         public void VisitProcStatementBreak(DMASTProcStatementBreak statementBreak) {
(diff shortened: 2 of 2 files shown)


DMCompiler/DM/Visitors/DMVisitorProcBuilder.cs
```diff
@@ -59,6 +59,14 @@ class DMVisitorProcBuilder : DMASTVisitor {
             _proc.Continue();
         }
 
+        public void VisitProcStatementGoto(DMASTProcStatementGoto statementGoto) {
+            _proc.Goto(statementGoto.Label.Identifier);
+        }
+
+        public void VisitProcStatementLabel(DMASTProcStatementLabel statementLabel) {
+            _proc.AddLabel(statementLabel.Name + "_goto");
```
ZeWaka, line 67: why _goto - you can `break` to labels.

wixoaGit (commented):
This causes issues when there are nested for loops enumerating a list

```dm
var/list/L = list(list(1, 2), list(3, 4))
for(var/item1 in L)
	world.log << "[item1]"
	if (!istype(item1, /list)) break

	for (var/item2 in item1)
		goto label
	label:
```

What this should do is print "/list" twice, but it instead prints "/list" and "2". This is because when a loop starts enumerating a list it pushes an enumerator onto a special stack, which should be popped when the loop is over, but the goto skips over that.

To fix this I'd rather change the behavior of loops than the implementation of goto.

DMCompiler/DM/Visitors/DMVisitorProcBuilder.cs
ike709: Good point

ike709:
> [snip]
> To fix this I'd rather change the behavior of loops than the implementation of goto.

This is above my level of comprehension for this project, so should I close this or leave it open until you address that?

wixoaGit:
Leave it open

DMCompiler/DM/Visitors/DMVisitorProcBuilder.cs
```diff
@@ -59,6 +59,14 @@ class DMVisitorProcBuilder : DMASTVisitor {
             _proc.Continue();
         }
 
+        public void VisitProcStatementGoto(DMASTProcStatementGoto statementGoto) {
+            _proc.Goto(statementGoto.Label.Identifier);
+        }
+
+        public void VisitProcStatementLabel(DMASTProcStatementLabel statementLabel) {
+            _proc.AddLabel(statementLabel.Name);
```
wixoaGit, line 67: This could end up conflicting with labels generated by other pieces of code. For example, a `label0:` label would conflict with the first label created by `_proc.NewLabelName()`

wixoaGit (changes requested):

wixoaGit:
After that's fixed, I'm going to just merge this and create a new issue for the loop problem.

DMCompiler/DM/Visitors/DMVisitorProcBuilder.cs
ike709: I don't know what exactly I should do about this.

wixoaGit: You could add something like `"_codelabel"` to the end, like you had before.

DMCompiler/DM/DMProc.cs
```diff
@@ -225,6 +225,10 @@ private class DMProcScope {
         public void ContinueIfFalse() {
             JumpIfFalse(_loopStack.Peek() + "_continue");
         }
+        
+        public void Goto(string label) {
+            Jump(label);
```
ike709, line 230:
```suggestion
            Jump(label + "_codelabel");
```

DMCompiler/DM/Visitors/DMVisitorProcBuilder.cs
ike709, line 67:
```suggestion
            _proc.AddLabel(statementLabel.Name + "_codelabel");
```

ike709:
Done

wixoaGit (approved):

PR #61, opened 2021-05-14
Implements md5()

ike709:
image

Since md5() can take a file or a text we first try to load the arg as a file converted to a string (largely copying file2text), then if that's null we just hash the arg string directly.

### DIFF
--- a/DMCompiler/DMStandard/_Standard.dm
+++ b/DMCompiler/DMStandard/_Standard.dm
@@ -72,6 +72,7 @@ proc/view(Dist = 4, Center = usr)
 proc/viewers(Depth, Center = usr)
 proc/walk(Ref, Dir, Lag = 0, Speed = 0)
 proc/walk_to(Ref, Trg, Min = 0, Lag = 0, Speed = 0)
+proc/md5(T)
 
 #include "Defines.dm"
 #include "Types\Client.dm"
--- a/DMCompiler/DMStandard/_Standard.dm
+++ b/DMCompiler/DMStandard/_Standard.dm
@@ -39,6 +39,7 @@ proc/length(E)
 proc/log(X, Y)
 proc/lowertext(T)
 proc/max(A)
+proc/md5(T)
 proc/min(A)
 proc/num2text(N, Digits, Radix)
 proc/oview(Dist = 5, Center = usr)
@@ -72,7 +73,6 @@ proc/view(Dist = 4, Center = usr)
 proc/viewers(Depth, Center = usr)
 proc/walk(Ref, Dir, Lag = 0, Speed = 0)
 proc/walk_to(Ref, Trg, Min = 0, Lag = 0, Speed = 0)
-proc/md5(T)
 
 #include "Defines.dm"
 #include "Types\Client.dm"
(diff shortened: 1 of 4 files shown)


ike709:
Turns out those dashes are because of BitConverter so give me a sec to replace it with an old-fashioned StringBuilder

DMCompiler/DMStandard/_Standard.dm
```diff
@@ -72,6 +72,7 @@ proc/view(Dist = 4, Center = usr)
 proc/viewers(Depth, Center = usr)
 proc/walk(Ref, Dir, Lag = 0, Speed = 0)
 proc/walk_to(Ref, Trg, Min = 0, Lag = 0, Speed = 0)
+proc/md5(T)
```
wixoaGit, line 75: Keep this in alphabetic order

OpenDreamServer/Dream/Procs/Native/DreamProcNative.cs
```diff
@@ -77,6 +77,7 @@ static class DreamProcNative {
             root.SetNativeProc(DreamProcNativeRoot.NativeProc_viewers);
             root.SetNativeProc(DreamProcNativeRoot.NativeProc_walk);
             root.SetNativeProc(DreamProcNativeRoot.NativeProc_walk_to);
+            root.SetNativeProc(DreamProcNativeRoot.NativeProc_md5);
```
wixoaGit, line 80: Same here

OpenDreamServer/Dream/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -1231,5 +1233,28 @@ public static DreamValue NativeProc_viewers(DreamObject instance, DreamObject us
 
             return DreamValue.Null;
         }
+        
+        [DreamProc("md5")]
+        [DreamProcParameter("T", Type = DreamValueType.String | DreamValueType.DreamResource)]
+        public static DreamValue NativeProc_md5(DreamObject instance, DreamObject usr, DreamProcArguments arguments) {
```
wixoaGit, line 1239: And here

wixoaGit (changes requested):

OpenDreamServer/Dream/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -1231,5 +1233,28 @@ public static DreamValue NativeProc_viewers(DreamObject instance, DreamObject us
...
+            if(arguments.ArgumentCount > 1) throw new Exception("md5() only takes one argument");
+            DreamValue arg = arguments.GetArgument(0, "T");
+            DreamResource resource;
+
+            if (arg.Type == DreamValueType.String) {
+                resource = Program.DreamResourceManager.LoadResource(arg.GetValueAsString());
```
wixoaGit: Giving `md5()` a string always hashes the string, it doesn't attempt to load it as a file.

OpenDreamServer/Dream/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -1231,5 +1233,28 @@ public static DreamValue NativeProc_viewers(DreamObject instance, DreamObject us
...
+            } else {
+                resource = arg.GetValueAsDreamResource();
+            }
+
+            string text = resource.ReadAsString();
+            if (text == null) text = arg.GetValueAsString();
```
wixoaGit: Lines 1242 to 1251 could be replaced with
```cs
string text;
if (arg.TryGetValueAsDreamResource(out DreamResource resource)) {
    text = resource.ReadAsString();
} else if (!arg.TryGetValueAsString(out text)) {
    return DreamValue.Null; //Return null if arg is neither a file or a string
}
```

wixoaGit (changes requested):

ike709:
@wixoaGit Addressed, the format of the output also matches byond exactly now.

wixoaGit (approved):

PR #62, opened 2021-05-14
Adds a batch script to copy DMStandard to the bin

ike709:
This allows IDEs like Rider to execute it in the build process, as an external tool.


ike709:
Tested and working.

ZeWaka (approved):

wixoaGit:
What if the directories `bin\Debug\net5.0` or `bin\Release\net5.0` don't exist?

ike709:
@wixoaGit Before my latest commit it would ask for confirmation to make them, I just modified it to make them automatically if they don't exist.

PR #66, opened 2021-05-15
implements GetValues() for datum.vars list

ZeWaka:
I think this is all that was needed?

according to http://www.byond.com/docs/ref/#/datum/var/vars

### DIFF
--- a/OpenDreamServer/Dream/Objects/DreamObject.cs
+++ b/OpenDreamServer/Dream/Objects/DreamObject.cs
@@ -76,6 +76,14 @@ namespace OpenDreamServer.Dream.Objects {
             }
         }
 
+        public List<DreamValue> GetVariables() {
+            List<DreamValue> list = new(_variables.Count);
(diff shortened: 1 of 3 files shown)


OpenDreamServer/Dream/Objects/DreamObject.cs
```diff
@@ -76,6 +76,14 @@ class DreamObject {
             }
         }
 
+        public List<DreamValue> GetVariables() {
```
wixoaGit, line 79: The name of this seems like it could be misleading, maybe `GetVariableNames()`?

wixoaGit (changes requested):

OpenDreamServer/Dream/Objects/DreamObject.cs
ZeWaka: I was just going off the other method, will change

wixoaGit (approved):

PR #68, opened 2021-05-15
Adds islist() and list2params()

ike709:
Both survived basic testing, but it's probable that the latter has cases I haven't considered.


OpenDreamServer/Dream/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -550,6 +557,40 @@ static class DreamProcNativeRoot {
...
+        [DreamProc("list2params")]
+        [DreamProcParameter("List")]
+        public static DreamValue NativeProc_list2params(DreamObject instance, DreamObject usr, DreamProcArguments arguments)
+        {
+            if (!arguments.GetArgument(0, "List").TryGetValueAsDreamList(out DreamList list)) throw new ArgumentException("list2params was passed a non-list");
+            
+            StringBuilder paramBuilder = new StringBuilder();
+            
+            if (list.IsAssociative()) {
+                Dictionary<DreamValue, DreamValue> associativeValues = list.GetAssociativeValues();
```
ZeWaka, line 570: This looks like it has the same issue as #92.

wixoaGit (changes requested):
The keys and values need to be URL encoded using `HttpUtility.UrlEncode`

Currently `list2params(list("@"="<"))` returns `@=<` when it should return `%40=%3c`

OpenDreamServer/Dream/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -550,6 +557,40 @@ static class DreamProcNativeRoot {
 
             throw new Exception("Cannot check length of " + value + "");
         }
+        
+        [DreamProc("list2params")]
+        [DreamProcParameter("List")]
+        public static DreamValue NativeProc_list2params(DreamObject instance, DreamObject usr, DreamProcArguments arguments)
+        {
+            if (!arguments.GetArgument(0, "List").TryGetValueAsDreamList(out DreamList list)) throw new ArgumentException("list2params was passed a non-list");
```
wixoaGit, line 565: Return an empty string instead of throwing an exception.

wixoaGit (changes requested):

OpenDreamServer/Dream/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -550,6 +557,40 @@ static class DreamProcNativeRoot {
...
+                        paramBuilder.Append('&');
+                    }
+                }
+            } else {
+                foreach (DreamValue val in values)
+                {
+                    paramBuilder.Append(val.Value);
+                    paramBuilder.Append('&');
+                }
+            }
```
wixoaGit, line 588:
```suggestion
            Dictionary<DreamValue, DreamValue> associativeValues = list.GetAssociativeValues();
            foreach (DreamValue entry in values) {
                if (associativeValues.ContainsKey(entry))
                {
                    paramBuilder.Append($"{entry.Value}={associativeValues[entry].Value}");
                    paramBuilder.Append('&');
                } else {
                    paramBuilder.Append(entry.Value);
                    paramBuilder.Append('&');
                }
            }
```

wixoaGit (changes requested):

ike709:
Addressed.

wixoaGit (approved):

PR #84, opened 2021-05-15
Adds support for the null index in DreamLists

ZeWaka:
title, cursed behaivor
image

### DIFF
--- a/OpenDreamServer/Dream/Objects/DreamList.cs
+++ b/OpenDreamServer/Dream/Objects/DreamList.cs
@@ -67,9 +67,6 @@ namespace OpenDreamServer.Dream.Objects {
         }
 
         public virtual DreamValue GetValue(DreamValue key) {
-            if (key == DreamValue.Null) // x = List[null]
-                key = new(String.Empty);
-
             if (key.Type == DreamValue.DreamValueType.Integer) {
                 return _values[key.GetValueAsInteger() - 1]; //1-indexed
             }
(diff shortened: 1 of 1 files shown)


OpenDreamServer/Dream/Objects/DreamList.cs
```diff
@@ -67,26 +67,25 @@ class DreamList : DreamObject {
         }
 
         public virtual DreamValue GetValue(DreamValue key) {
-            if (key == DreamValue.Null) return DreamValue.Null; //TODO: Null key
+            if (key == DreamValue.Null) // x = List[null]
+                key = new(String.Empty);
```
wixoaGit, line 71: This makes `list[null]` and `list[""]` the same thing, which is incorrect.

OpenDreamServer/Dream/Objects/DreamList.cs
```diff
@@ -119,10 +118,10 @@ class DreamList : DreamObject {
         }
 
         public static bool IsValidAssociativeKey(DreamValue key) {
-            return key.Value != null && key.IsType( DreamValue.DreamValueType.String |
-                                                    DreamValue.DreamValueType.DreamPath |
-                                                    DreamValue.DreamValueType.DreamObject |
-                                                    DreamValue.DreamValueType.DreamResource);
+            return key != null && (key == DreamValue.Null || key.IsType(DreamValue.DreamValueType.String |
```
wixoaGit, line 121: DreamValue is a struct, which cannot be null.
```suggestion
            return (key == DreamValue.Null || key.IsType(DreamValue.DreamValueType.String |
```

wixoaGit (changes requested):

OpenDreamServer/Dream/Objects/DreamList.cs
ZeWaka: What do you think then? We'd have to use some cursed placeholder value?

wixoaGit (approved):

PR #87, opened 2021-05-15
Window title / `world.name`

ZeWaka:
title

it goes `world.name` -> the filename of the .dme -> i don't think its possible to go past that but i put a fallback

### DIFF
--- a/OpenDreamRuntime/DreamRuntime.cs
+++ b/OpenDreamRuntime/DreamRuntime.cs
@@ -121,7 +121,7 @@ namespace OpenDreamRuntime
 
             Map = new DreamMap(this);
             Map.LoadMap(CompiledJson.Maps[0]);
-            
+
             WorldInstance.SpawnProc("New");
         }
 
@@ -177,6 +177,8 @@ namespace OpenDreamRuntime
                 return null;
             }
 
+            CompiledJson.SourceFileName = Path.GetFileNameWithoutExtension(path);
+
             return compiledJson;
         }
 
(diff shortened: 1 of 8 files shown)


ZeWaka:
Ideally the worldName finding stuff would live somewhere else and be stored in some global state, but we're trying to get rid of global states so I'm not sure what I would do to remedy.

OpenDreamRuntime/DreamRuntime.cs
```diff
@@ -222,16 +224,27 @@ public class DreamRuntime
             connection.SendPacket(new PacketInterfaceData(_clientInterface));
             connection.SendPacket(new PacketFullGameState(StateManager.FullState));
 
+            // Set to <world> part of <world>.json unless we define world.name - final fallback to "OpenDream World"
+            String worldName;
+            DreamValue worldInstanceName = WorldInstance.GetVariable("name");
+            if (worldInstanceName.Value != null ) {
```
wixoaGit, line 230:
```suggestion
            if (worldInstanceName != DreamValue.Null) {
```

wixoaGit (commented):
This doesn't handle `/world.name` changing while the client is already connected

ike709:
...How's this going?

ZeWaka:
Haven't gotten to it - feel free to take it over, IRL has been busy and I have a deadline for goon

I also don't know how I would handle this for the stated error above, however I forget beyond behavior, but personally I don't think that the window title would change upon update unless you reconnect in which case my PR would work

ZeWaka:
Probably irrelevant with RobustToolbox

PR #92, opened 2021-05-15
Properly encodes assoc. lists with both associated and non-associated values

### LINKED ISSUE #90: When converting from a normal list to an associative list, elements are lost
title
```cs
var/list/L = list("1", 2, usr)
world << json_encode(L) // ["1",2,"atom"] vs. ["1",2,"atom"]
L["yeet"] = 69
world << json_encode(L) // {"yeet":69} vs. {"1":null,"":null,"atom":null,"yeet":69}
```

ZeWaka:
Properly handles json encoding of lists that don't have an associated value with values.

Fixes #90


ZeWaka:
Can also add a method to DreamList along the lines of `DreamList.ContainsAssocativeValue(val)` if that implementation is preferred - don't know your preference.

wixoaGit (approved):

PR #96, opened 2021-05-17
Removes boxing for DreamPath equality

ZeWaka:
title

See #95 for reasoning

relevant: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/how-to-define-value-equality-for-a-type

### DIFF
--- a/OpenDreamShared/Dream/DreamPath.cs
+++ b/OpenDreamShared/Dream/DreamPath.cs
@@ -193,7 +193,7 @@ namespace OpenDreamShared.Dream {
 
         public static bool operator ==(DreamPath lhs, DreamPath rhs) => lhs.Equals(rhs);
         
-        public static bool operator !=(DreamPath lhs, DreamPath rhs) => (lhs == rhs);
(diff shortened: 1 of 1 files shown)


ZeWaka:
should be tested with stripped TG.

OpenDreamShared/Dream/DreamPath.cs
```diff
@@ -175,24 +175,26 @@ public enum PathType {
...
+            return true;
+        }
+        
         public override int GetHashCode() {
             return PathString.GetHashCode();
         }
 
+        public static bool operator ==(DreamPath lhs, DreamPath rhs) => lhs.Equals(rhs);
+        
+        public static bool operator !=(DreamPath lhs, DreamPath rhs) => (lhs == rhs);
```
wixoaGit, line 196: Is this suppose to be `!(lhs == rhs)`?

ZeWaka, line 196:
```suggestion
        public static bool operator !=(DreamPath lhs, DreamPath rhs) => !(lhs == rhs);
```

wixoaGit (approved):

PR #98, opened 2021-05-17
fixes stupid reference equality in #95

ZeWaka:
see title, fixes #95 / #97 

https://web.archive.org/web/20100317143155/http://msdn.microsoft.com/en-us/library/ms173147(VS.80).aspx / https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/how-to-define-value-equality-for-a-type

> By default, the operator `==` tests for reference equality by determining if two references indicate the same object...

> The `==` and `!=` operators can be used with classes even if the class does not overload them. However, the default behavior is to perform a reference equality check.

now is also faster: 80ds

### DIFF
--- a/OpenDreamServer/Dream/DreamValue.cs
+++ b/OpenDreamServer/Dream/DreamValue.cs
@@ -270,14 +270,11 @@ namespace OpenDreamServer.Dream {
             }
         }
 
-        public bool Equals(DreamValue other) {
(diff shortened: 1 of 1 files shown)


ZeWaka:
still getting weirdness - will work on tomorrow

wixoaGit (approved):

PR #101, opened 2021-05-18
execution context for managing/scheduling proc call

willox:
it's in working order, just wanted to get this up so people can see it before i finish some tweaks (there's a few TODOs about.)

sleep and spawn both work, but in a half-assed way that won't support `-1` or `0` waits properly (and probably more.)

will fix https://github.com/wixoaGit/OpenDream/issues/94


willox:
`DreamProc` is a base class for any proc. We instantiate one per proc as it only contains static information such as `Name`, `SuperProc`, `ArgumentNames`.

`NativeProc : DreamProc` is an implementation that supports calling the native procs you had written previously (except for sleep, which is no longer one of those.)

`DMProc : DreamProc` is an implementation that contains the DM interpreter, so it is used for procs written in DM.

`SleepProc : DreamProc` is an implementation used only for `sleep()`.

Each of the procs has an associated State type, an instance of the state will exist for each current call to those procs.

willox:
I've just pushed basic support for retrieving return values asyncronously in C# (for cases where you want the return value even if the proc has paused during execution.)

If you sleep in `/client/New()` before `. = ..()`, you'll see it works. Whether or not the rest of the code supports that isn't something I'd know.

I still want to write something a little more robust for situations like verbs where we need to wait on multiple procs to execute (such as when a verb has two prompts to show.)

OpenDreamServer/Dream/Procs/DMProc.cs
```diff
@@ -0,0 +1,281 @@
...
+    class DMProcState : ProcState
+    {
+        delegate ProcStatus? OpcodeHandler(DMProcState state);
+
+        // TODO: This pool is not returned to if the proc runtimes
+        private static ArrayPool<DreamValue> _dreamValuePool = ArrayPool<DreamValue>.Shared;
+
+        #region Opcode Handlers
+        //In the same order as the DreamProcOpcode enum
+        private static OpcodeHandler[] _opcodeHandlers = new OpcodeHandler[] {
```
ZeWaka, line 33:
```suggestion
        private static readonly OpcodeHandler[] _opcodeHandlers = {
```
can be marked readonly - immutable
also avoid object allocation and use array initialization instead

OpenDreamServer/Dream/Procs/DMProc.cs
```diff
@@ -0,0 +1,281 @@
...
+        public DreamValue PopDreamValue() {
+            object value = _stack.Pop();
+
+            if (value is IDreamProcIdentifier identifier) {
+                return identifier.GetValue();
+            } else if (value is DreamValue dreamValue) {
+                return dreamValue;
+            } else {
+                throw new Exception("Last object on stack was not a dream value or identifier");
+            }
```
ZeWaka, line 247:
```suggestion
            return value switch {
                IDreamProcIdentifier identifier => identifier.GetValue(),
                DreamValue dreamValue => dreamValue,
                _ => throw new Exception("Last object on stack was not a dream value or identifier")
            };
```
brrrr

OpenDreamServer/Dream/Procs/ExecutionContext.cs
```diff
@@ -0,0 +1,165 @@
...
+    enum ProcStatus {
+        Returned,
+        Deferred,
+        Called,
+    }
+
+    abstract class DreamProc {
+        public string Name { get; }
+
+        // This is currently publically settable because the loading code doesn't know what our super is until after we are instantiated
```
ZeWaka, line 18:
```suggestion
        // This is currently publicly settable because the loading code doesn't know what our super is until after we are instantiated
```

:^)

willox:
image
probably not pushing this tonight, but i got this working by defining our own TaskScheduler and SynchronizationContext.

willox:
Ready for review, although it could use some documentation.

Async builtins are defined like this:
```csharp
        [DreamProc("sleep")]
        [DreamProcParameter("Delay", Type = DreamValueType.Number)]
        public static async Task<DreamValue> NativeProc_sleep(AsyncNativeProc.State state) {
            float delay = state.Arguments.GetArgument(0, "Delay").GetValueAsNumber();
            int delayMilliseconds = (int)(delay * 100);

            // This is obviously not the proper behaviour
            await Task.Delay(delayMilliseconds);
            return DreamValue.Null;
        }
```

You can spawn a new thread (vm coroutine) for them using:
```chsarp
DreamProc proc = ...; // This can represent a sync/async native proc or a DM proc
DreamThread.Run(proc, src, usr, arguments);
```

You can also spawn a new thread with an anonymous async native proc:
```csharp
DreamThread.Run(async (state) => {
    DreamProc proc = ...; 
    var result = await state.Call(proc, src, usr, arguments);
    // We're always still on the main-thread here. 
    // If the proc we called did not sleep, this whole operation has been synchronous!
    UseResult(result);
    return DreamValue.Null;
});
```

willox:
@wixoaGit ok i went a little overboard, hopefully it's ok, i moved a bunch of stuff around to get the code into a testable state

i'm not touching this for a while unless you want something changed

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -854,7 +856,7 @@ public static DreamValue NativeProc_oviewers(DreamObject instance, DreamObject u
         public static DreamList params2list(string queryString) {
             queryString = queryString.Replace(";", "&");
             NameValueCollection query = HttpUtility.ParseQueryString(queryString);
-            DreamList list = new DreamList();
+            DreamList list = new DreamList(CurrentRuntime);
```
wixoaGit, line 859: params2list is used outside of proc calls (`HandlePacketTopic`), so `CurrentRuntime` may be null.

OpenDreamRuntime/DreamRuntime.cs
```diff
@@ -0,0 +1,239 @@
...
+            ObjectTree.SetMetaObject(DreamPath.Mob, new DreamMetaObjectMob(this));
+            DreamProcNative.SetupNativeProcs(ObjectTree);
+
+            ListDefinition = ObjectTree.GetObjectDefinitionFromPath(DreamPath.List);
+
+            WorldInstance = ObjectTree.CreateObject(DreamPath.World);
+            ObjectTree.GetObjectDefinitionFromPath(DreamPath.Root).GlobalVariables["world"].Value = new DreamValue(WorldInstance);
+
+            Map = new DreamMap(this);
+            Map.LoadMap(CompiledJson.Maps[0]);
```
wixoaGit, line 115: Loading the map should be done after the global init proc is called, and before `/world.New()` is called.

OpenDreamRuntime/Procs/NativeProc.cs
```diff
@@ -0,0 +1,84 @@
...
+                _proc = proc;
+                Src = src;
+                Usr = usr;
+                Arguments = arguments;
+            }
+
+            protected override ProcStatus InternalResume()
+            {
+                if (DreamProcNativeRoot.CurrentRuntime != null) {
+                    throw new InvalidOperationException();
```
wixoaGit, line 35: This causes an issue with nested native proc calls. An example of one of these is the native proc `image()`, which creates an `/image` object. This calls `/image.New()`, which calls the native proc `istext()`.

wixoaGit (changes requested):

OpenDreamRuntime/Procs/NativeProc.cs
willox: This `DreamProcNativeRoot.CurrentRuntime` is a bit of a hack I was hoping won't be permanent (but it definitely beats editing every single native proc even more inside this single PR.)

I can change this static var to a stack for now - but ideally that `/image/New` call will happen inside the same DreamThread as the `/image()` proc. This is probably best achieved by changing `image` to an AsyncNativeProc because... that's where all the complex functionality like that lives (even if the proc ends up never yielding.)

OpenDreamRuntime/DreamRuntime.cs
willox: Ah, I messed around with this when porting the Main from OpenDreamServer and must have missed this bit when I tried to keep the ordering the same.

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
willox: Good point. I'll change HandlePacketTopic to create a (synchronous) DreamThread instead of directly calling the native proc.

If you wanted to avoid the DreamThread (which isn't exactly expensive), the solution would be to just move the functionality of params2list out of the native proc.

OpenDreamRuntime/DMMParser.cs
```diff
@@ -1,13 +1,15 @@
...
+using OpenDreamRuntime.Objects;
 using OpenDreamShared.Compiler.DM;
 using OpenDreamShared.Dream;
 using System;
 using System.Collections.Generic;
 
+// Why is this in the wrong project?
 namespace OpenDreamShared.Compiler.DMM {
     class DMMParser : DMParser {
+        public DreamRuntime Runtime { get; }
```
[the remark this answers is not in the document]
willox: Do you have something that points this out automatically? Would be useful to know.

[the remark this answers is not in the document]
wixoaGit: No, these were all the issues I ran into while attempting to run my copy of tg

OpenDreamRuntime/Procs/NativeProc.cs
wixoaGit: Currently, the `New()` call on every object creates a new DreamThread. This is because `New()` is called in the meta objects' `OnObjectCreated` using `SpawnProc`.

willox: Yeah, that behaviour can stay for now. It's not ideal only because it'll lead to worse stack traces in errors and won't allow New to be asynchronous (is it ever?)

I've fixed all the other issues locally, just working on this last one now.

willox:
Ok those are sorted

OpenDreamRuntime/DreamConnection.cs
```diff
@@ -1,42 +1,59 @@
...
+    public abstract class DreamConnection {
+        // Interface
+        public abstract byte[] ReadPacketData();
+        public abstract void SendPacket(IPacket packet);
+
+        // Implementation
+        public string CKey;
+        public IPAddress Address;
+        public readonly List<int> PressedKeys = new();
+        public DreamRuntime Runtime { get; }
```
wixoaGit, line 30: Does `Runtime` need to be a public variable on all of these classes?

OpenDreamRuntime/DMMParser.cs
```diff
@@ -1,13 +1,15 @@
...
-using OpenDreamServer.Dream.Objects;
+﻿using OpenDreamRuntime;
+using OpenDreamRuntime.Objects;
 using OpenDreamShared.Compiler.DM;
 using OpenDreamShared.Dream;
 using System;
 using System.Collections.Generic;
 
+// Why is this in the wrong project?
 namespace OpenDreamShared.Compiler.DMM {
```
wixoaGit, line 9: This can be moved to the OpenDreamRuntime namespace.

OpenDreamServer/Server.cs
```diff
@@ -0,0 +1,136 @@
...
+            if (!_ckeyToConnection.ContainsKey(pRequestConnect.CKey)) {
+                connection.CKey = pRequestConnect.CKey;
+
+                DreamConnectionRequest.Invoke(connection);
+            } else {
+                connection.SendPacket(new PacketConnectionResult(false, "A connection with your ckey already exists"));
+            }
+        }
+	}
+}
```
wixoaGit, line 136: This file also has a mixture of tabs and spaces.

wixoaGit (changes requested):

willox:
Can I create a .editorconfig file to deal with the indentation? I'd prefer to not be changing the defaults of my editors.

https://editorconfig.org/

I'm going to bed now so don't expect any changes soon.

wixoaGit:
> Can I create a .editorconfig file to deal with the indentation?

Yes, an editorconfig would be nice to have

.editorconfig
```diff
@@ -0,0 +1,16 @@
...
+[*.{cs, xaml}]
+indent_style = space
+indent_size = 4
+
+[*.csproj]
+indent_style = space
+indent_size = 2
+
+[*.{dm, dme, dmm, dmf}]
+indent_style = tab
```
ZeWaka, line 16: pog

.editorconfig
```diff
@@ -0,0 +1,16 @@
+root = true
+
+[*.{cs, xaml, csproj, dm, dme, dmm, dmf}]
+end_of_line = lf
+insert_final_newline = true
```
ZeWaka, line 5: pog 2

willox:
anything else to do?

OpenDreamRuntime.Tests/UnitTest1.cs
```diff
@@ -0,0 +1,149 @@
...
+using OpenDreamRuntime.Objects;
+using System.Collections.Generic;
+using System.Net;
+using System.Threading.Tasks;
+using System.Diagnostics;
+using System.IO;
+using OpenDreamRuntime.Procs;
+using OpenDreamShared.Dream;
+
+namespace OpenDreamServer.Tests
```
wixoaGit, line 13:
```suggestion
namespace OpenDreamRuntime.Tests
```

wixoaGit (changes requested):

wixoaGit:
Beyond that, everything looks good.

wixoaGit (approved):

PR #102, opened 2021-05-18
Converts foreach -> for List loops for `ProcNativeList`/`DreamList`

ZeWaka:
These were iterating on `List`s. The other one in `DreamProcNativeList` are iterating on arguments, so I don't care about them.

Profiling for `ContainsValue` (called via `Set`):
```
DreamList.ContainsValue(DreamValue) - Own+System / Total time
foreach: 9,585ms / 22,320ms  (3,338ms was spent by System)
for: 3,544ms / 16,198ms      (No System time spent!)
```

test code:
```cs
var/list/L = list()
for(var/i in 1 to 12500)
    L["[i]"] = 1
```

Reading:
https://codingsight.com/foreach-or-for-that-is-the-question
https://stackoverflow.com/questions/365615/in-net-which-loop-runs-faster-for-or-foreach

### DIFF
--- a/OpenDreamServer/Dream/Objects/DreamList.cs
+++ b/OpenDreamServer/Dream/Objects/DreamList.cs
@@ -21,9 +21,9 @@ namespace OpenDreamServer.Dream.Objects {
 
         private static DreamObjectDefinition _listDefinition = null;
 
-        private List<DreamValue> _values = new();
-        private Dictionary<DreamValue, DreamValue> _associativeValues = new();
-        private object _listLock = new object();
+        private readonly List<DreamValue> _values = new();
+        private readonly Dictionary<DreamValue, DreamValue> _associativeValues = new();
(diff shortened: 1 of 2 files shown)


ike709:
Did you profile in Debug or Release?

ZeWaka:
> Did you profile in Debug or Release?

@ike709 Release bruv

wixoaGit:
Using the test code you gave, this seems to perform slower than before.

```dm
var/start = world.timeofday

var/list/L = list()
for(var/i in 1 to 12500)
	L["[i]"] = 1

var/elapsed = world.timeofday - start
world << "[elapsed] deciseconds"
```

Before
image

After
image

ZeWaka:
> Using the test code you gave, this seems to perform slower than before.

Looks like I messed up on my profiling locally, good catch.

PR #104, opened 2021-05-23
Adds support for indeterminate proc & macro args

ike709:
Fixes https://github.com/wixoaGit/OpenDream/issues/14 and the same issue with procs.

I'm not sure if we do any checking that the number of args in a proc/macro call matches the number in the definition, but if we do show me where and I'll adjust that.

Survives parsing BeeStation but I'm not experienced with advanced preprocessor tricks so look over this carefully.


ike709:
Ugh looks like I goofed parent calls.

ike709:
Should be good now?

ike709:
Conflicts resolved.

OpenDreamShared/Compiler/DM/DMLexer.cs
```diff
@@ -114,10 +114,19 @@ public class DMLexer : Lexer<Token> {
                         case TokenType.DM_Preproc_Punctuator_RightBracket: bracketNesting--; Advance(); token = CreateToken(TokenType.DM_RightBracket, preprocToken.Text); break;
                         case TokenType.DM_Preproc_Punctuator_Comma: Advance(); token = CreateToken(TokenType.DM_Comma, preprocToken.Text); break;
                         case TokenType.DM_Preproc_Punctuator_Period: {
-                            if (Advance().Type == TokenType.DM_Preproc_Punctuator_Period) {
-                                token = CreateToken(TokenType.DM_SuperProc, "..");
-
+                            if (Advance().Type == TokenType.DM_Preproc_Punctuator_Period)
+                            {
                                 Advance();
+                                if (GetCurrent().Type == TokenType.DM_Preproc_Punctuator_Period)
```
wixoaGit, line 120:
```suggestion
                                if (Advance().Type == TokenType.DM_Preproc_Punctuator_Period)
```

OpenDreamShared/Compiler/DM/DMParser.cs
```diff
@@ -921,6 +921,9 @@ public class DMParser : Parser<Token> {
 
                 if (Check(TokenType.DM_Comma)) {
                     Whitespace();
+                    
+                    if (Check(TokenType.DM_IndeterminateArgs)) break;
```
wixoaGit, line 925: This doesn't work with `proc/example(a, ..., b)`

wixoaGit (changes requested):

OpenDreamShared/Compiler/DM/DMParser.cs
ike709: That's valid DM?

wixoaGit: Yes

ike709: While this is technically valid, the docs specify the last parameter can end in `...` and AFAIK nobody uses it anywhere except as the last arg. Do you still want me to change it?

wixoaGit: I'd prefer for the compiler to accept it, for now. Maybe later we can make it a warning too.

ike709:
OpenDream test code:
```
test_args_one(a, ..., b)
    world << "[a], [b], [args.len]"

test_args_two(a, b, ...)
    world << "[a], [b], [args.len]"

verb/test_args()
    test_args_one(1,2,3,4,6)
    test_args_two(1,2,3,4,6)
```
Output:
image

BYOND:
image

Compiler parity resolved.

OpenDreamShared/Compiler/DM/DMParser.cs
```diff
@@ -921,6 +921,14 @@ public class DMParser : Parser<Token> {
 
                 if (Check(TokenType.DM_Comma)) {
                     Whitespace();
+
+                    if (Check(TokenType.DM_IndeterminateArgs) && !Check(TokenType.DM_Comma))
```
wixoaGit, line 925: This breaks if there is whitespace after the IndeterminateArgs token (`proc/example(a, ... , b)`). Rewriting the loop like this would probably be best:
```cs
public DMASTDefinitionParameter[] DefinitionParameters() {
    List<DMASTDefinitionParameter> parameters = new();
    DMASTDefinitionParameter parameter = DefinitionParameter();

    if (parameter != null) parameters.Add(parameter);
    while (Check(TokenType.DM_Comma)) {
        Whitespace();

        parameter = DefinitionParameter();
        if (parameter != null) {
            parameters.Add(parameter);
        } else if (!Check(TokenType.DM_IndeterminateArgs)) {
            Error("Expected parameter definition");
        }
    }

    return parameters.ToArray();
}
```

wixoaGit (changes requested):

ike709:
Done.

OpenDreamShared/Compiler/DM/DMParser.cs
```diff
@@ -913,19 +913,21 @@ public class DMParser : Parser<Token> {
         }
 
         public DMASTDefinitionParameter[] DefinitionParameters() {
-            List<DMASTDefinitionParameter> parameters = new List<DMASTDefinitionParameter>();
+            List<DMASTDefinitionParameter> parameters = new();
             DMASTDefinitionParameter parameter = DefinitionParameter();
 
-            while (parameter != null) {
+            if (parameter != null || Check(TokenType.DM_IndeterminateArgs)) {
                 parameters.Add(parameter);
```
wixoaGit, line 920: I just realized this adds null to parameters if the first argument is DM_IndeterminateArgs
```suggestion
            if (parameter != null || Check(TokenType.DM_IndeterminateArgs)) {
                if (parameter != null) parameters.Add(parameter);
```

wixoaGit (changes requested):

ike709:
Donezo again

wixoaGit (approved):

PR #107, opened 2021-05-23
add test for async proc call & fix for potential resumptions of stale proc states

willox:
The problem I noticed was that if a proc completes synchronously, we'd return `ProcStatus.Returned` on the first call to InternalResume but the task's continuation would still be scheduled (and would later call Thread.Resume again.)

Seeing as we've already got the result, we don't need to schedule the task at all and can just discard it.

### DIFF
--- a/OpenDreamRuntime.Tests/UnitTest1.cs
+++ b/OpenDreamRuntime.Tests/UnitTest1.cs
@@ -146,7 +146,7 @@ namespace OpenDreamRuntime.Tests
             Assert.Zero(runtime.ExceptionCount);
         }
 
-        [Test, Timeout(1000)]
+        [Test, Timeout(10000)]
         public void AsyncCall() {
             var runtime = CreateRuntime();
             var result = DreamValue.Null;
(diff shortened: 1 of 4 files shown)


OpenDreamRuntime.Tests/UnitTest1.cs
```diff
@@ -145,5 +145,24 @@ public class Tests {
             Assert.AreEqual(imageDefinition, obj.ObjectDefinition);
             Assert.Zero(runtime.ExceptionCount);
         }
+
+        [Test, Timeout(1000)]
```
wixoaGit, line 149: I ran this test a few times, it tends to take about 900ms and I even had it fail once. Should the timeout be increased?

wixoaGit (changes requested):

OpenDreamRuntime.Tests/UnitTest1.cs
willox: Yes, I'll web edit it to a generous 10 seconds as it'll ideally not fail most of the time

wixoaGit (approved):

PR #108, opened 2021-05-23
CRASH proc ceases execution of caller

willox:
I made this work by having the CRASH proc return a special kind of exception. I also swapped around the args to Assert.AreEqual in my tests because I didn't realise it mattered before.

### DIFF
--- a/OpenDreamRuntime/DreamThread.cs
+++ b/OpenDreamRuntime/DreamThread.cs
@@ -44,8 +44,8 @@ namespace OpenDreamRuntime {
         }
     }
 
-    class ProcRuntime : Exception {
-        public ProcRuntime(string message)
+    class PropagatingRuntime : Exception {
+        public PropagatingRuntime(string message)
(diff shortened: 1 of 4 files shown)


willox:
Just edited this to also make the stack overflow exception stop every proc in the stack, instead of just the top-most.

OpenDreamRuntime/DreamThread.cs
```diff
@@ -154,6 +176,13 @@ public class DreamThread {
             _current = state;
         }
 
+        // Returns true if the thread still contains any proc states
```
wixoaGit, line 179: This doesn't return anything

wixoaGit (changes requested):

OpenDreamRuntime/DreamThread.cs
willox: bah

wixoaGit (approved):

PR #109, opened 2021-05-24
waitfor support in OpenDreamRuntime

willox:
~This doesn't add support for compiling `set waitfor = <expr>`, but handles everything else. For now I just added support for `set OPENDREAM_WAITFOR_FALSE = <ignored expr>` to the compiler.~

~The compiler part is going to be a pain because it requires evaluating constant expressions at compile-time. If you don't want this merged until that is done... maybe I can figure it out.~

adds `set waitfor = ...` support, more details below


OpenDreamRuntime/Procs/AsyncNativeProc.cs
```diff
@@ -96,23 +96,26 @@ protected override ProcStatus InternalResume()
...
-                // We need to call a proc
-                if (_callProcNotify != null) {
-                    var callProcNotify = _callProcNotify;
-                    _callProcNotify = null;
-
-                    Thread.PushProcState(callProcNotify);
-                    return ProcStatus.Called;
-                }
+                // We need to call a proc.
+                while (_callProcNotify != null || _callResult != null)
```
willox, line 100: ^ I also added this loop as AsyncNativeProcs with multiple (synchronous) external calls in a row would deadlock.

willox:
Okay I went and did it but there's quite a bit of room for clean-up. The change is mostly moving logic for emitting expressions from DMVisitorProcBuilder into a separate class (which uses another visitor internally.)

DMCompiler/DM/Visitors/DMVisitorExpression.cs
```diff
@@ -0,0 +1,422 @@
...
+        public void VisitCallableProcIdentifier(DMASTCallableProcIdentifier procIdentifier) {
+            Result = new Expressions.Proc(procIdentifier.Identifier);
+        }
+
+        public void VisitProcCall(DMASTProcCall procCall) {
+            // arglist hack
+            if (procCall.Callable is DMASTCallableProcIdentifier ident) {
+                if (ident.Identifier == "arglist") {
+                    if (procCall.Parameters.Length != 1) throw new Exception("arglist must be the only argument");
+                    if (procCall.Parameters.Length != 1) throw new Exception("arglist must have 1 argument");
```
wixoaGit, line 109: You're making the same check twice here.

DMCompiler/DM/Visitors/DMVisitorExpression.cs
```diff
@@ -0,0 +1,422 @@
...
+        }
+
+        public void VisitNewPath(DMASTNewPath newPath) {
+            var args = new ArgumentList(_dmObject, _proc, newPath.Parameters);
+            Result = new Expressions.NewPath(newPath.Path.Path, args);
+        }
+
+        public void VisitNewInferred(DMASTNewInferred newInferred) {
+            if (_inferredPath is null) {
+                throw new Exception("An inferred new requires a type!");
```
wixoaGit, line 341: This is throwing with the line `/datum/point/proc/copy_to(datum/point/p = new)`

DMCompiler/DM/Visitors/DMVisitorExpression.cs
```diff
@@ -0,0 +1,422 @@
...
+
+        public void VisitConstantNull(DMASTConstantNull constant) {
+            Result = new Expressions.Null();
+        }
+
+        public void VisitConstantInteger(DMASTConstantInteger constant) {
+            Result = new Expressions.Number(constant.Value);
+        }
+
+        public void VisitConstantFloat(DMASTConstantInteger constant) {
```
wixoaGit, line 33:
```suggestion
        public void VisitConstantFloat(DMASTConstantFloat constant) {
```

DMCompiler/DM/Visitors/DMVisitorExpression.cs
```diff
@@ -0,0 +1,422 @@
+using OpenDreamShared.Compiler.DM;
+using OpenDreamShared.Dream;
+using OpenDreamShared.Dream.Procs;
+using System;
+using System.Collections.Generic;
+
+namespace DMCompiler.DM.Visitors {
+    class DMVisitorExpression : DMASTVisitor {
```
wixoaGit, line 8: This is missing VisitIn (`value in list`), and VisitCall (`call()()`)

DMCompiler/DM/DMExpression.cs
```diff
@@ -0,0 +1,1318 @@
...
+
+            public override void EmitPushValue(DMObject dmObject, DMProc proc) {
+                if (_path != null) {
+                    _path.EmitPushValue(dmObject, proc);
+                } else {
+                    throw new Exception("implicit locate() not implemented");
+                }
+
+                if (_container != null) {
+                    _container.EmitPushProc(dmObject, proc);
```
wixoaGit, line 1163:
```suggestion
                    _container.EmitPushValue(dmObject, proc);
```

wixoaGit (changes requested):

DMCompiler/DM/Visitors/DMVisitorExpression.cs
willox: It's kind of annoying that we can't be sure every expression visitor is implemented. Perhaps the interface should be split into 2 instead of having default implementations. (Not right now though)

willox:
That's all fixed

DMCompiler/DM/Visitors/DMVisitorExpression.cs
```diff
@@ -0,0 +1,460 @@
...
+        }
+
+        public void VisitNewPath(DMASTNewPath newPath) {
+            var args = new ArgumentList(_dmObject, _proc, newPath.Parameters);
+            Result = new Expressions.NewPath(newPath.Path.Path, args);
+        }
+
+        public void VisitNewInferred(DMASTNewInferred newInferred) {
+            if (_inferredPath is null) {
+                throw new Exception("An inferred new requires a type!");
```
wixoaGit, line 340: This is throwing with the line `var/datum/parsed_map/parsed = cached_map || new(file(mappath))`

wixoaGit (changes requested):
A `..()` call with zero arguments is not supplying the super proc with the current proc's args.

```dm
proc/test()
	world << "A [args.len]"

test()
	world << "B [args.len]"
	..()

verb/test_verb()
	test(1, 2, 3)
```

Expected output:
```
B 3
A 3
```
Actual output:
```
B 3
A 0
```

DMCompiler/DM/Visitors/DMVisitorExpression.cs
willox: I should be able to make the inferred type propagate down to sub-expressions

willox: oh god byond even supports `new()` inside the condition part of a ternary expression

wixoaGit: I believe `new()` works anywhere in a variable assign's value. I know `var/datum/d = new(new())` works.

willox:
```
	var/list/L = list()
	L[new()] = new()
```

lmfao it builds

willox: i'm going to... decently replicate this. it's a super simple change on our end so we can change it at a later date

willox:
done

OpenDreamRuntime.Tests/DMProject/code.dm
```diff
@@ -24,6 +24,82 @@
...
+	. = 5
+	sleep(1)
+	. = 6
+
+//
+
+/world/proc/default_test(datum/arg = new())
+	return arg
+
+/world/proc/crazy_inferred_types()
```
wixoaGit, line 75: This isn't being used in any tests

wixoaGit (changes requested):

OpenDreamRuntime.Tests/DMProject/code.dm
willox: i was happy with this as long as it manages to compile (meaning that the new() is inferring _something_)

wixoaGit (approved):

PR #111, opened 2021-05-24
Adds `world.cpu`

### LINKED ISSUE #78: `world.cpu` is not implemented
http://www.byond.com/docs/ref/#/world/var/cpu

ike709:
Not sure if this is the best way to handle it, but it seems to work.

Fixes #78

### DIFF
--- a/DMCompiler/DMStandard/Types/World.dm
+++ b/DMCompiler/DMStandard/Types/World.dm
@@ -12,7 +12,7 @@
 	var/timeofday
 	var/realtime
 	var/tick_lag = 1
-	var/cpu = 100
+	var/cpu = 0
 	var/fps = null
 	var/tick_usage
 
--- a/DMCompiler/DMStandard/Types/World.dm
+++ b/DMCompiler/DMStandard/Types/World.dm
@@ -12,6 +12,7 @@
 	var/timeofday
 	var/realtime
 	var/tick_lag = 1
+	var/cpu = 100
 	var/fps = null
 	var/tick_usage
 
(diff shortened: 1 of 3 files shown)


DMCompiler/DMStandard/Types/World.dm
```diff
@@ -12,6 +12,7 @@
 	var/timeofday
 	var/realtime
 	var/tick_lag = 1
+	var/cpu = 100
```
ike709, line 15: Didn't seem like the best idea to leave it null on the first tick, so instead lets presume the first tick is costly.

wixoaGit: `world.cpu` starts off as a 0

OpenDreamRuntime/DreamRuntime.cs
```diff
@@ -157,6 +157,7 @@ public class DreamRuntime
                 int elapsedTime = (int)(new DateTimeOffset(DateTime.Now).ToUnixTimeMilliseconds() - TickStartTime);
                 int tickLength = (int)(100 * WorldInstance.GetVariable("tick_lag").GetValueAsFloat());
                 int timeToSleep = tickLength - elapsedTime;
+                WorldInstance.SetVariable("cpu", new DreamValue(100 - timeToSleep));
```
wixoaGit, line 160: This assumes a `world.tick_lag` of 1. `world.cpu` should be a ratio of `elapsedTime` and `tickLength`.
```suggestion
                WorldInstance.SetVariable("cpu", new DreamValue((float)elapsedTime / tickLength * 100));
```

TestGame/code.dm
```diff
@@ -49,6 +49,9 @@
 	    for(var/item in out3)
 	    	usr << "The output should be between 1 and 10: [item]"
 
+	verb/show_cpu()
+		usr << "CPU: [world.cpu]"
```
wixoaGit, line 53: Could this be moved to the `Stat` proc?

wixoaGit (changes requested):

DMCompiler/DMStandard/Types/World.dm
ike709:
```suggestion
	var/cpu = 0
```

ike709:
Done.

wixoaGit (approved):

PR #112, opened 2021-05-25
Adds logo

ZeWaka:
Adds a logo (probably not the final one) to the repo and to the README.
I originally took it in a more BYOND-inspired direction, but that's probably not the best idea:
<details>
  <summary>Old Draft Logo</summary>
</img>
</details>
Here it is, the grey text is so it can be read on white/black/darkgrey:

OpenDream

I just think we should have at least a temporary one at the very least.


wixoaGit:
The SVG doesn't look much like the PNG (in my browser, at least). The font is different, and the letters are clipping into each other.

image

ZeWaka:
Guess the font embedding is broken, i'll do it as paths then.

wixoaGit (approved):

PR #116, opened 2021-05-26
conditional access operators

willox:
This adds `?.`.

When used on the LHS of an assignment (`=`, `+=`, etc.)`, the entire assignment will evaluate to null without the RHS executing.

`x?.y.z` will always result in a runtime if x is null (as `null.z` is performed.) This behaviour varies depending on which version of BYOND you are using: http://www.byond.com/forum/post/2667038?hla=Willox#comment25786061. I chose the sane one.

closes https://github.com/wixoaGit/OpenDream/issues/106


willox:
I also touched up a few of the opcode handlers which were pushing/popping different amounts of values depending on which branch they took. Caused problems in situations like `return L[2] *= 15 `, where null would be returned instead of the list's new value.

ike709:
Conflicts

willox:
conflicts fixed

OpenDreamShared/Compiler/Token.cs
```diff
@@ -27,9 +27,11 @@ public enum TokenType {
         DM_Preproc_Punctuator_Comma,
         DM_Preproc_Punctuator_LeftBracket,
         DM_Preproc_Punctuator_LeftParenthesis,
+        DM_Preproc_Punctuator_Colon,
```
wixoaGit, line 30: Keep these in alphabetic order

wixoaGit (changes requested):

willox:
Would you like me to pull my const_expr branch onto this before merging it? It cleans up the expressions namespace into separate files and adds support for (most) constant expressions in switch statement conditions.

https://github.com/willox/OpenDream/compare/conditionals...const_expr

Otherwise I'll just PR it after this is merged.

OpenDreamShared/Compiler/Token.cs
```diff
@@ -46,6 +48,7 @@ public enum TokenType {
         DM_BarEquals,
         DM_Break,
         DM_Call,
+        DM_QuestionColon,
```
wixoaGit, line 51: These too

wixoaGit (changes requested):

wixoaGit:
Make that another PR

wixoaGit (approved):

Issue #120, opened 2021-05-26
`|` / `BitOr()` only works for integers


ZeWaka:
https://github.com/wixoaGit/OpenDream/blob/25a6348434a3e76411f67c1cda3feb446bcf7def/OpenDreamRuntime/Procs/DMOpcodeHandlers.cs#L528-L532

PR #123, opened 2021-05-26
Implements DreamList unions (aka `L1 | L2`)

ZeWaka:
title

partially solves #120 - won't be doing the icon bit now

```cs
var/list/L = list(1,2,3)
var/list/G = list(3,4,5)
var/x = 7
var/y = 8
world.log << json_encode(L | G)
world.log << json_encode(G | L)
world.log << json_encode(L | x)
world.log << json_encode(x | y)
world.log << json_encode(L | null)
world.log << json_encode(null | L)
```
to
```
[1,2,3,4,5]
[3,4,5,1,2]
[1,2,3,7]
15
[1,2,3,null]
null
```

### DIFF
--- a/OpenDreamRuntime/Objects/DreamList.cs
+++ b/OpenDreamRuntime/Objects/DreamList.cs
@@ -140,6 +140,15 @@ namespace OpenDreamRuntime.Objects {
         public int GetLength() {
             return _values.Count;
         }
+
+        public DreamList Union(DreamList other) {
+            DreamList copy = CreateCopy();
+            copy._values.AddRange(other.GetValues());
+            foreach ((DreamValue key, DreamValue value) in other.GetAssociativeValues()) {
+                copy._associativeValues[key] = value;
+            }
+            return copy;
+        }
     }
 
     // /datum.vars list
--- a/OpenDreamRuntime/Objects/DreamList.cs
+++ b/OpenDreamRuntime/Objects/DreamList.cs
@@ -1,5 +1,6 @@
 ﻿using OpenDreamRuntime.Procs;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace OpenDreamRuntime.Objects {
     delegate void DreamListValueAssignedEventHandler(DreamList list, DreamValue key, DreamValue value);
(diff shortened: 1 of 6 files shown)


OpenDreamRuntime/Objects/DreamList.cs
```diff
@@ -144,6 +144,15 @@ public class DreamList : DreamObject {
         public int GetLength() {
             return _values.Count;
         }
+
+        public DreamList Union(DreamList other) {
```
ZeWaka, line 148: could also override the `|` operator if you want

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -526,10 +526,16 @@ static class DMOpcodeHandlers {
...
         public static ProcStatus? BitOr(DMProcState state) {
-            int second = state.PopDreamValue().GetValueAsInteger();
-            int first = state.PopDreamValue().GetValueAsInteger();
+            DreamValue value = state.PopDreamValue();
+            if (value.TryGetValueAsInteger(out int secondInt)) {
+                int first = state.PopDreamValue().GetValueAsInteger();
+
+                state.Push(new DreamValue(first | secondInt));
+            } else if (value.TryGetValueAsDreamList(out DreamList secondList)) {
+                DreamList first = state.PopDreamValue().GetValueAsDreamList();
```
ike709, line 535: What if someone mixes an integer and a list?

ZeWaka: :sadcat:

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -525,11 +525,35 @@ static class DMOpcodeHandlers {
...
+        public static ProcStatus? BitOr(DMProcState state) {                                // x | y
+            DreamValue secondValue = state.PopDreamValue();
 
-            state.Push(new DreamValue(first | second));
+            if (secondValue.TryGetValueAsInteger(out int secondInt)) {                      // x | 2
+                DreamValue firstValue = state.PopDreamValue();
+
+                if (firstValue.TryGetValueAsInteger(out int firstInt)) {                    // 1 | 2
+                    state.Push(new DreamValue(firstInt | secondInt));
+                } else if (firstValue.TryGetValueAsDreamList(out DreamList firstList)) {    // List | 2
```
wixoaGit, line 536: How this is done with other operators (`+`, `-`, etc) is there's an operator method in `IDreamMetaObject`. If the first value is a `DreamObject`, that method gets called. The override of `OperatorOr` in `DreamMetaObjectList` would handle the union.

wixoaGit (changes requested):

wixoaGit (approved):

PR #124, opened 2021-05-27
constant expressions support more types & are used for switch conditions

willox:
This leaves out a few things:
- Still no support for using constant variables (`var/const/x = ...`) inside constant expressions
- Actually, this doesn't touch var/const at all. Switch conditions and `set` are the only things using constant expressions.
- No builtin calls are supported (`md5("...")`, etc.)

You might notice that I haven't implemented spooky behaviour like `1 + "blah"` evaluating to `0` or whatever, this is because BYOND actually doesn't support this if done inside a constant expression.
`var/const/x = "a" + 2` will not compile on BYOND.


DMCompiler/DM/Expressions/Procs.cs
```diff
@@ -0,0 +1,99 @@
...
+            proc.PushSelf();
+        }
+
+        public override ProcPushResult EmitPushProc(DMObject dmObject, DMProc proc) {
+            proc.PushSelf();
+            return ProcPushResult.Unconditional;
+        }
+    }
+
+    // ..
```
ZeWaka, line 43: what about `...` and further friends? :^)

ZeWaka: unironically a method to get the name of the current proc would be great extension in the future

DMCompiler/DM/Expressions/Constant.cs
```diff
@@ -0,0 +1,262 @@
...
+#endregion
+    }
+
+    // null
+    class Null : Constant {
+        public override void EmitPushValue(DMObject dmObject, DMProc proc) {
+            proc.PushNull();
+        }
+
+        public override bool IsTruthy() => true;
```
wixoaGit, line 89: Should this return false?

wixoaGit (changes requested):

DMCompiler/DM/Expressions/Constant.cs
willox: yeah i think i copied this from a now removed Not() implementation

wixoaGit (approved):

PR #131, opened 2021-05-30
object initialization is now a proc

willox:
`new DreamList` has been swapped for `DreamList.Create` in many places (which will call the list/New stuff in an immediate manner.) I think in an ideal future we just won't require any DM to run for list initialization.

This change allows datum/New to sleep. Anything using `InitSpawn` will just call the initialization method in a new DreamThread and throw an exception if the initializations were not completed synchronously.

In an absolutely ideal future we'd remove InitSpawn


wixoaGit:
> I think in an ideal future we just won't require any DM to run for list initialization.

We already don't _need_ to run any DM for list initialization. Currently, all the DM that is run when you create a list is this
```dm
proc/New(Size)
	if (Size != null) len = Size
```
This could easily be moved to `DreamMetaObjectList`'s `OnObjectCreated`.

wixoaGit (approved):

Issue #43, opened 2021-05-11
input() doesn't support every type


ike709:
Currently only supports `text`, `num`, and `message`.

See #36

ZeWaka:
the types to add: 
```
password      // un-echoed text (for use with input() only)
command_text  // raw command text from the rest of the input line
icon          // an icon file from the user's computer
sound         // a sound file from the user's computer
file          // any type of file from the user's computer
key           // a key from the user's BYOND key file
color         // a color (see rgb proc)
```

PR #149, opened 2021-06-23
Implemented the color and password input types

DamianX:
See #43

### DIFF
--- a/OpenDreamClient/Interface/Prompts/ColorPrompt.cs
+++ b/OpenDreamClient/Interface/Prompts/ColorPrompt.cs
@@ -1,0 +1,36 @@
+﻿using OpenDreamShared.Dream.Procs;
+using System;
+using System.Windows.Controls;
+using System.Windows.Media;
+using Xceed.Wpf.Toolkit;
+
+namespace OpenDreamClient.Interface.Prompts
+{
+    class ColorPrompt : InputWindow
+    {
+        public ColorPrompt(int promptId, string title, string message, string defaultValue, bool canCancel) : base(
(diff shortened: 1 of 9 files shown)


OpenDreamClient/Interface/Prompts/ColorPrompt.cs
```diff
@@ -0,0 +1,36 @@
+﻿using OpenDreamShared.Dream.Procs;
+using System;
+using System.Windows.Controls;
+using System.Windows.Media;
+using Xceed.Wpf.Toolkit;
```
ZeWaka, line 5: Why are you using WPF? This is not cross-platform.

DamianX: I was under the impression the client was already not cross platform. What should be used instead?

ZeWaka: It's only not cross-platform because WebView2 hasn't been finished for Linux/Darwin. The last comment from Microsoft on this matter was that they intended to do so by the end of this year.

DamianX: Good to know. I tried making this work with Windows.Forms.ColorDialog but I'm getting weird hangs that I don't know how to debug, so I'm closing this.

wixoaGit: WPF is already used for everything UI-related.

ZeWaka: `Microsoft.Web.WebView2.Wpf` that is

PR #150, opened 2021-06-23
Added byond compiler output files to gitignore

DamianX:
So that you can compile with dream maker and not have to worry about accidentally committing them

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,8 +22,20 @@
 /DMCompiler/obj
 /DMCompiler/Properties
 /packages/System.Net.Sockets.4.3.0
+
 /TestGame/environment.json
+/TestGame/*.dmb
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -22,8 +22,20 @@
 /DMCompiler/obj
 /DMCompiler/Properties
 /packages/System.Net.Sockets.4.3.0
+
 /TestGame/environment.json
+/TestGame/*.dmb
+/TestGame/*.int
+/TestGame/*.lk
+/TestGame/*.rsc
+/TestGame/*.rsc.lk
```
ZeWaka, line 31: Why explicitly these directories? Why not everywhere?

wixoaGit (approved):

PR #151, opened 2021-06-27
Implements var/thing[len] list syntax

### LINKED ISSUE #49: Doesn't support `var/foo[size]` list initialization syntax
from old goon code:
```
var/crban_reason[0]	// Banned key+reason
```

results in:
```
Error at code\game\admin\ban.dm:4:19: Expected EOF
```

ike709:
Test code:
```
verb/stupid_list()
    var/thing[5]
    usr << thing.len
```

Fixes #49

### DIFF
--- a/OpenDreamShared/Compiler/DM/DMParser.cs
+++ b/OpenDreamShared/Compiler/DM/DMParser.cs
@@ -421,17 +421,26 @@ namespace OpenDreamShared.Compiler.DM {
 
                 DMASTExpression value = null;
 
-                if (Check(TokenType.DM_LeftBracket) && !Check(TokenType.DM_RightBracket)) //TODO: Multidimensional lists
+                if (Check(TokenType.DM_LeftBracket)) //TODO: Multidimensional lists
                 {
                     //Type information
                     if (path.Path.FindElement("list") != 0)
                     {
-                        path = new DMASTPath(new DreamPath("list/"  + path.Path.PathString));
+                        path = new DMASTPath(new DreamPath("list/" + path.Path.PathString));
                     }
 
-                    var size = Expression();
+                    Whitespace();
 
-                    Consume(TokenType.DM_RightBracket, "Expected ']'");
+                    var size = new DMASTConstantInteger(0);
+                    if (!Check(TokenType.DM_RightBracket))
+                    {
+                        var expr = Expression() as DMASTExpressionConstant;
+                        if (expr is DMASTConstantInteger integer)
+                        {
+                            size = integer;
+                        }
+                        Consume(TokenType.DM_RightBracket, "Expected ']'");
+                    }
 
                     value = new DMASTNewPath(new DMASTPath(DreamPath.List),
                             new[] {new DMASTCallParameter(size)});
--- a/OpenDreamShared/Compiler/DM/DMParser.cs
+++ b/OpenDreamShared/Compiler/DM/DMParser.cs
@@ -447,7 +447,7 @@ namespace OpenDreamShared.Compiler.DM {
                 if (Check(TokenType.DM_Equals)) {
                     if (value is not null)
                     {
-                        Error("list doubly initialized");
+                        Error("List doubly initialized");
                     }
                     else
                     {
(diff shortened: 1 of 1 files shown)


OpenDreamShared/Compiler/DM/DMParser.cs
```diff
@@ -418,13 +418,42 @@ public class DMParser : Parser<Token> {
...
+                if (Check(TokenType.DM_LeftBracket)) //TODO: Multidimensional lists
+                {
+                    //Type information
+                    if (path.Path.PathString.Substring(0, 4) != "list")
+                    {
+                        path = new DMASTPath(new DreamPath("list/"  + path.Path.PathString));
+                    }
+
+                    var size = Expression() as DMASTExpressionConstant;
+                    if (size is not DMASTConstantInteger)
```
wixoaGit, line 433: After some testing, it appears DM will actually take any expression as a valid size for the list (including non-constants). The value of the variable is also `null` when the brackets are empty (`var/list/L[]`).

OpenDreamShared/Compiler/DM/DMParser.cs
```diff
@@ -418,13 +418,42 @@ public class DMParser : Parser<Token> {
...
                 DMASTPath path = Path();
                 if (path == null) Error("Expected a variable name");
-                Whitespace();
 
                 DMASTExpression value = null;
 
+                if (Check(TokenType.DM_LeftBracket)) //TODO: Multidimensional lists
+                {
+                    //Type information
+                    if (path.Path.PathString.Substring(0, 4) != "list")
```
wixoaGit, line 427:
```suggestion
                    if (path.Path.FindElement("list") != 0)
```

wixoaGit (changes requested):

OpenDreamShared/Compiler/DM/DMParser.cs
```diff
@@ -418,13 +418,37 @@ public class DMParser : Parser<Token> {
...
+                    if (path.Path.FindElement("list") != 0)
+                    {
+                        path = new DMASTPath(new DreamPath("list/"  + path.Path.PathString));
+                    }
+
+                    var size = Expression();
+
+                    Consume(TokenType.DM_RightBracket, "Expected ']'");
+
+                    value = new DMASTNewPath(new DMASTPath(DreamPath.List),
```
wixoaGit, line 436: `value` should not be set if there was no expression, and `size` is null.

OpenDreamShared/Compiler/DM/DMParser.cs
```diff
@@ -418,13 +418,37 @@ public class DMParser : Parser<Token> {
...
+                }
+
+                Whitespace();
+
                 if (Check(TokenType.DM_Equals)) {
-                    Whitespace();
-                    value = Expression();
+                    if (value is not null)
+                    {
+                        Error("list doubly initialized");
```
wixoaGit, line 445:
```suggestion
                        Error("List doubly initialized");
```

wixoaGit (changes requested):

OpenDreamShared/Compiler/DM/DMParser.cs
```diff
@@ -418,13 +418,56 @@ public class DMParser : Parser<Token> {
...
+                            size = integer;
+                        }
+                        else
+                        {
+                            size = new DMASTConstantInteger(0);
+                        }
+
+                        Whitespace();
+                        Consume(TokenType.DM_RightBracket, "Expected ']'");
+                    }
```
wixoaGit, line 449: `Expression()` returns null if it couldn't parse an expression, and already consumes any whitespace after it. `var/L[expression]` is also always equal to `var/list/L = new /list(expression)` (from what I can tell). Because of this, this can be simplified down to 2 lines:

```suggestion
                    DMASTExpression size = Expression();
                    Consume(TokenType.DM_RightBracket, "Expected ']'");
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #152, opened 2021-06-27
Adds missing valid escapes

ike709:
See http://www.byond.com/docs/ref/#/DM/text/macros

Fixes https://github.com/wixoaGit/OpenDream/issues/140


OpenDreamShared/Compiler/DM/DMLexer.cs
```diff
@@ -9,19 +9,26 @@ public class DMLexer : Lexer<Token> {
             "t", "n",
             "[", "]",
             "\\", "\"", "'",
+            "<", ">",
 
             "icon",
             "Roman", "roman",
             "The", "the",
             "A", "a", "An", "an",
+            "th",
```
ZeWaka, line 18: cursed

wixoaGit (approved):

PR #155, opened 2021-06-28
Adds `atom/var/text`

ike709:
See http://www.byond.com/docs/ref/#/atom/var/text

### DIFF
--- a/DMCompiler/DMStandard/Types/Atoms/_Atom.dm
+++ b/DMCompiler/DMStandard/Types/Atoms/_Atom.dm
@@ -35,7 +35,7 @@
 	var/gender = "neuter"
 	var/density = FALSE
 
-	proc/New()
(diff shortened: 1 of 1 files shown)


DMCompiler/DMStandard/Types/Atoms/_Atom.dm
```diff
@@ -34,6 +35,9 @@
 	var/gender = "neuter"
 	var/density = FALSE
 
+	New()
```
ZeWaka, line 38: Does this override properly if someone sets `text` in their atom/New() ?

ike709:
Blocked by https://github.com/wixoaGit/OpenDream/issues/158

wixoaGit:
The `/atom.text` default is set at compile time, not in `New()`

PR #169, opened 2021-06-28
Use 32x32 as default icon size

ike709:
Fixes https://github.com/wixoaGit/OpenDream/issues/129


wixoaGit:
The correct thing to do here wouldn't be to assume an icon size of 32x32. The icon size needs to be determined using the dimensions of the png, and the amount of icons in the file.

ike709:
Will reopen if I have time

PR #215, opened 2021-07-07
Adds more info to yielding New() exceptions

ike709:
Includes the typepath: 
`Unhandled exception. System.InvalidOperationException: DreamObject.InitSpawn (/obj/structure/closet/secure_closet/injection) called a yielding proc!`

### DIFF
--- a/OpenDreamRuntime/Objects/DreamObject.cs
+++ b/OpenDreamRuntime/Objects/DreamObject.cs
@@ -29,7 +29,7 @@ namespace OpenDreamRuntime.Objects {
             thread.PushProcState(procState);
 
             if (thread.Resume() == DreamValue.Null) {
-                throw new InvalidOperationException("DreamObject.InitSpawn called a yielding proc!");
+                throw new InvalidOperationException($"DreamObject.InitSpawn ({this.ObjectDefinition.Type}) called a yielding proc!");
             }
         }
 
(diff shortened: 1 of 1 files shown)


OpenDreamRuntime/Objects/DreamObject.cs
```diff
@@ -29,7 +29,7 @@ public class DreamObject {
             thread.PushProcState(procState);
 
             if (thread.Resume() == DreamValue.Null) {
-                throw new InvalidOperationException("DreamObject.InitSpawn called a yielding proc!");
+                throw new InvalidOperationException($"DreamObject.InitSpawn ({this.ObjectDefinition.Type}) called a yielding proc!");
```
wixoaGit, line 32: It would be better to include a DM stack trace. This can be done by calling `thread.HandleException`
```suggestion
                Exception exception = new InvalidOperationException("DreamObject.InitSpawn called a yielding proc!");

                thread.HandleException(exception);
                throw exception;
```
Though I'm not sure if this should throw. Execution should be able to continue despite this.

wixoaGit (changes requested):

OpenDreamRuntime/Objects/DreamObject.cs
ike709: So what should I do?

wixoaGit: Just call `thread.HandleException`, don't throw.
```suggestion
                thread.HandleException(new InvalidOperationException("DreamObject.InitSpawn called a yielding proc!"));
```

ike709:
You got it boss

wixoaGit (approved):

PR #222, opened 2021-07-09
Fixes ckey() and adds ckeyEx()

### LINKED ISSUE #180: `ckeyEx()` not implemented
http://www.byond.com/docs/ref/#/proc/ckeyEx

NTstation

ike709:
Closes #180 and fixes ckey() to match BYOND
I added `[\\^]` because when testing inputs with `^` it was failing to catch it, same for when I included it in the main capture group. But I'm not great at regex so...


DMCompiler/DMStandard/_Standard.dm
```diff
@@ -8,6 +8,7 @@ proc/arcsin(X)
 proc/arctan(A)
 proc/ascii2text(N)
 proc/ckey(Key)
+proc/ckeyEx(Key)
```
wixoaGit, line 11: According to the [reference page](http://www.byond.com/docs/ref/#/proc/ckeyEx), this parameter's name is `Text`
```suggestion
proc/ckeyEx(Text)
```

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -130,7 +130,16 @@ static class DreamProcNativeRoot {
...
+            return new DreamValue(key);
+        }
+
+        [DreamProc("ckeyEx")]
+        [DreamProcParameter("Key", Type = DreamValueType.String)]
+        public static DreamValue NativeProc_ckeyEx(DreamObject instance, DreamObject usr, DreamProcArguments arguments) {
+            string key = arguments.GetArgument(0, "Key").GetValueAsString();
+
+            key = Regex.Replace(key, "[\\^]|[^A-z0-9@_-]", ""); //Remove all punctuation except - and _
             return new DreamValue(key);
```
wixoaGit, line 143: According to the [reference page](http://www.byond.com/docs/ref/#/proc/ckeyEx), this parameter's name is `Text`
```suggestion
        [DreamProcParameter("Text", Type = DreamValueType.String)]
        public static DreamValue NativeProc_ckeyEx(DreamObject instance, DreamObject usr, DreamProcArguments arguments) {
            string text = arguments.GetArgument(0, "Text").GetValueAsString();

            text = Regex.Replace(text, "[\\^]|[^A-z0-9@_-]", ""); //Remove all punctuation except - and _
            return new DreamValue(text);
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #224, opened 2021-07-09
Adds client/connection

### LINKED ISSUE #184: `client/connection` not implemented
http://www.byond.com/docs/ref/#/client/var/connection

NTstation

ike709:
Closes #184 

According to the docs the value is null until the handshake is done: http://www.byond.com/docs/ref/#/client/var/connection

### DIFF
--- a/OpenDreamRuntime/DreamRuntime.cs
+++ b/OpenDreamRuntime/DreamRuntime.cs
@@ -191,6 +191,7 @@ namespace OpenDreamRuntime
             connection.SendPacket(new PacketConnectionResult(true, null, interfaceResource.ReadAsString()));
             connection.SendPacket(new PacketFullGameState(StateManager.FullState, connection.CKey));
 
+            client.SetVariable("connection", new DreamValue("seeker"));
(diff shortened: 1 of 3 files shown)


OpenDreamRuntime/DreamRuntime.cs
```diff
@@ -191,6 +191,7 @@ public class DreamRuntime
             connection.SendPacket(new PacketConnectionResult(true, null, interfaceResource.ReadAsString()));
             connection.SendPacket(new PacketFullGameState(StateManager.FullState, connection.CKey));
 
+            client.SetVariable("connection", new DreamValue("seeker"));
```
wixoaGit, line 194: This leaves it possible for DM code to set `client.connection`. This should be implemented as a variable getter in `DreamMetaObjectClient`

wixoaGit (changes requested):

ike709:
Done. I also converted the `else if` chains while I was in there, since it's more performant.

wixoaGit (approved):

PR #226, opened 2021-07-09
Implements `world.sleep_offline`

ike709:
Closes https://github.com/wixoaGit/OpenDream/issues/194

Doesn't properly implement `sleep_offline = 0` because I have no idea how to handle that (see the docs: http://www.byond.com/docs/ref/#/world/var/sleep_offline)

Draft because I don't see where we actually accurately track the number of connected clients. I had to manually clear the `Connections` list while testing.


ike709:
This will probably need changes with RT so I'll just close it for now.

PR #233, opened 2021-07-10
Implemented icon_states(), moved DMI parser to shared

### LINKED ISSUE #198: `icon_states()` not implemented
http://www.byond.com/docs/ref/#/proc/icon_states

DamianX:
Closes #198

I moved the bit that was parsing the PNG header to Shared, and implemented reading the width of the image from the IHDR chunk by hand rather than relying on GDI+, so the server doesn't need to import that.


OpenDreamShared/Resources/DMIParser.cs
```diff
@@ -54,6 +57,81 @@ public class ParsedDMIFrame {
...
+            return (width, height);
+        }
+
+        public static (string, uint, uint) ReadDMIDescription(byte[] bytes) {
+            MemoryStream stream = new MemoryStream(bytes);
+            BinaryReader reader = new BinaryReader(stream);
+
+            stream.Seek(8, SeekOrigin.Begin);
+
+            var (width, height) = _readIHDRChunk(reader, stream);
```
wixoaGit, line 97: I don't like the IHDR reader being used in the zTXt reader, could you split them up?

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -406,6 +407,17 @@ public static DreamValue NativeProc_hascall(DreamObject instance, DreamObject us
             return new DreamValue(HttpUtility.HtmlEncode(plainText));
         }
 
+        [DreamProc("icon_states")]
+        [DreamProcParameter("Icon", Type = DreamValueType.DreamResource)]
+        [DreamProcParameter("mode", Type = DreamValueType.Float)]
+        public static DreamValue NativeProc_icon_states(DreamObject instance, DreamObject usr, DreamProcArguments arguments)
```
wixoaGit, line 413: This should throw a NotImplementedException if `mode` isn't 0

wixoaGit (changes requested):

OpenDreamShared/Resources/DMIParser.cs
DamianX: I'm not sure what you mean. It's necessary to somehow obtain the width from the IHDR chunk because `ParseDMIDescription` needs it.

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -406,6 +407,26 @@ public static DreamValue NativeProc_hascall(DreamObject instance, DreamObject us
             return new DreamValue(HttpUtility.HtmlEncode(plainText));
         }
 
+        [DreamProc("icon_states")]
+        [DreamProcParameter("Icon", Type = DreamValueType.DreamResource)]
+        [DreamProcParameter("mode", Type = DreamValueType.Float)]
```
wixoaGit, line 412:
```suggestion
        [DreamProcParameter("mode", Type = DreamValueType.Float, DefaultValue = 0)]
```

wixoaGit (changes requested):

OpenDreamShared/Resources/DMIParser.cs
wixoaGit: `ReadDMIDescription` is the zTXt reader, and shouldn't be returning data from other chunks.

With this being moved to shared, I don't see any reason for code using DMIParser to have to use both ReadDMIDescription and ParseDMIDescription to parse a DMI. Maybe instead there could be a `ParseDMI(byte[] data)` method that reads the chunks, and returns a ParsedDMIDescription?

DamianX: `ReadDMIDescription` is back to only dealing with zTXt. `icon_states` now uses the new `GetIconStatesFromDescription`, which exclusively cares about icon states. `ParseDMIDescription` is back to only being used by the client. Is this okay?

wixoaGit: Yes, that's good

wixoaGit (approved):

PR #235, opened 2021-07-10
Removes pointless allocating and iterating from CreateListEnumerator()

ike709:


OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -43,17 +43,16 @@ static class DMOpcodeHandlers {
...
-            if (list != null) {
-                values.Capacity = list.GetLength();
-
-                foreach (DreamValue value in list.GetValues()) {
-                    values.Add(value);
-                }
+            if (list == null)
+            {
+                List<DreamValue> values = new List<DreamValue>();
+                state.EnumeratorStack.Push(values.GetEnumerator());
```
ike709, line 49: Is this part necessary or can I just not `Push()` if it's null?

wixoaGit: An enumerator has to be pushed. You could do `Enumerable.Empty<DreamValue>().GetEnumerator()` instead.

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -43,17 +43,16 @@ static class DMOpcodeHandlers {
...
+            }
+            else
+            {
+                state.EnumeratorStack.Push(list.GetValues().GetEnumerator());
```
wixoaGit, line 53: This breaks loops that modify the list they're enumerating.

wixoaGit (approved):

PR #236, opened 2021-07-12
Adds initial `locs` support

### LINKED ISSUE #178: `atom/movable/var/locs` not implemented
Needed for NTstation

http://www.byond.com/docs/ref/#/atom/movable/var/locs

ike709:
Fixes #178 

Also converts elseif chains to switches.

### DIFF
--- a/OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectMovable.cs
+++ b/OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectMovable.cs
@@ -24,48 +24,78 @@ namespace OpenDreamRuntime.Objects.MetaObjects {
             base.OnObjectDeleted(dreamObject);
         }
 
-        public override void OnVariableSet(DreamObject dreamObject, string variableName, DreamValue variableValue, DreamValue oldVariableValue) {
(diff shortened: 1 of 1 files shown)


OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectMovable.cs
```diff
@@ -24,48 +24,78 @@ public DreamMetaObjectMovable(DreamRuntime runtime)
...
+                    }
                 }
-            } else {
-                return base.OnVariableGet(dreamObject, variableName, variableValue);
+                case "locs":
+                {
+                    //TODO multi-tile support
+                    var list = DreamList.Create(Runtime);
+                    list.AddValue(dreamObject.GetVariable("loc"));
+                    return new DreamValue(list);
```
ZeWaka, line 95: Why do we need to wrap the list in a new dreamvalue? Isn't a list already one?

wixoaGit: A list is a DreamObject, which has to be wrapped by a DreamValue

wixoaGit (approved):

PR #238, opened 2021-07-15
Removes every `HasFlag()`

ike709:
Correct me if I'm wrong, but `HasFlag()` uses boxing, which is slow.


wixoaGit:
Do any of these usages of `HasFlag` run often enough for it to be a performance concern?

ike709:
I haven't checked, but is there any actual reason *not* to just change it?

ZeWaka:
Do you know that it boxes or are you just guessing? That's how the PR reads to me.

ike709:
@ZeWaka https://medium.com/@arphox/the-boxing-overhead-of-the-enum-hasflag-method-c62a0841c25a

ike709:
https://stackoverflow.com/questions/7368652/what-is-it-that-makes-enum-hasflag-so-slow

wixoaGit (approved):

PR #242, opened 2021-07-20
CreateObject now attempts to convert a string to a path

ike709:
Solves a lot of exceptions running NTstation

### DIFF
--- a/OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
+++ b/OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
@@ -68,7 +68,15 @@ namespace OpenDreamRuntime.Procs {
 
         public static ProcStatus? CreateObject(DMProcState state) {
             DreamProcArguments arguments = state.PopArguments();
-            DreamPath objectPath = state.PopDreamValue().GetValueAsPath();
+            var val = state.PopDreamValue();
+            if (!val.TryGetValueAsPath(out var objectPath) && val.TryGetValueAsString(out var pathString))
+            {
+                objectPath = new DreamPath(pathString);
+                if (!state.Runtime.ObjectTree.HasTreeEntry(objectPath))
+                {
+                    throw new Exception($"Cannot create unknown object {val.Value}");
+                }
+            }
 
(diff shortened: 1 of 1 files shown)


wixoaGit:
The `CreateObject` opcode should only ever be given a `DreamPath`. If it's being given a string, there's some other issue going on. What code is causing the exceptions?

ike709:
```
/obj/effect/spawner/lootdrop
	icon = 'icons/mob/screen_gen.dmi'
	icon_state = "x2"
	var/lootcount = 1		//how many items will be spawned
	var/lootdoubles = 0		//if the same item can be spawned twice
	var/list/loot			//a list of possible items to spawn e.g. list(/obj/item, /obj/structure, /obj/effect)

/obj/effect/spawner/lootdrop/initialize()
	if(loot && loot.len)
		for(var/i = lootcount, i > 0, i--)
			if(!loot.len) break
			var/lootspawn = pick(loot)
			if(!lootdoubles)
				loot.Remove(lootspawn)

			new lootspawn(get_turf(src))
	qdel(src)
```

The `loot` var is varedited in the DMM to be a list of types as strings.

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -68,7 +68,15 @@ static class DMOpcodeHandlers {
 
         public static ProcStatus? CreateObject(DMProcState state) {
             DreamProcArguments arguments = state.PopArguments();
-            DreamPath objectPath = state.PopDreamValue().GetValueAsPath();
+            var val = state.PopDreamValue();
+            if (!val.TryGetValueAsPath(out var objectPath) && val.TryGetValueAsString(out var pathString))
```
wixoaGit, line 72: This will end up with `objectPath` being `DreamPath.Root` if `val` is neither a path or a string, and will go on to create it. It should error instead.

wixoaGit (changes requested):

ike709:
Should be good

wixoaGit (approved):

PR #255, opened 2021-07-29
Adds some basic info to the console when compiling

ike709:
Just mentions the DME name when you start compiling and mentions how long it takes

### DIFF
--- a/DMCompiler/Program.cs
+++ b/DMCompiler/Program.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Reflection;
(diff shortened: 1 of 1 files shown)


DMCompiler/Program.cs
```diff
@@ -18,6 +18,8 @@ class Program {
 
         static void Main(string[] args) {
             if (!VerifyArguments(args)) return;
+            Console.WriteLine($"Compiling {GetDMEName(args)}");
```
wixoaGit, line 21: Could this print every file that's being compiled, instead of just the DME?

wixoaGit (changes requested):

wixoaGit (approved):

PR #263, opened 2021-08-01
Adds a program arg to dump the preprocessor

ike709:
And generally makes the compiler more arg-friendly.

Just toss in `--dump-preprocessor` and it Just Works:tm:

### DIFF
--- a/DMCompiler/Program.cs
+++ b/DMCompiler/Program.cs
@@ -33,7 +33,7 @@ namespace DMCompiler {
                     result.Append(t.Text);
                 }
 
-                string output = Path.ChangeExtension(CompiledFiles[0], "txt") ?? Path.Join(System.AppDomain.CurrentDomain.BaseDirectory, "preprocessor_dump.txt");
+                string output = Path.ChangeExtension(CompiledFiles[0], "txt") ?? Path.Join(System.AppDomain.CurrentDomain.BaseDirectory, "preprocessor_dump.dm");
                 File.WriteAllText(output, result.ToString());
                 Console.WriteLine($"Preprocessor output dumped to {output}");
             }
--- a/DMCompiler/Program.cs
+++ b/DMCompiler/Program.cs
@@ -1,7 +1,9 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Reflection;
+using System.Text;
 using System.Text.Json;
 using DMCompiler.Compiler.DMM;
 using DMCompiler.DM;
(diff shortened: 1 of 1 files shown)


DMCompiler/Program.cs
```diff
@@ -14,13 +16,29 @@
...
-            DMPreprocessor preprocessor = Preprocess(args);
+            DMPreprocessor preprocessor = Preprocess(CompiledFiles);
+            if (HasArgument("--dump-preprocessor"))
+            {
+                StringBuilder result = new();
+                foreach (Token t in preprocessor.GetResult()) {
+                    result.Append(t.Text);
+                }
+
+                string output = Path.ChangeExtension(CompiledFiles[0], "txt") ?? Path.Join(System.AppDomain.CurrentDomain.BaseDirectory, "preprocessor_dump.txt");
```
wixoaGit, line 36: Is there a situation where `Path.ChangeExtension(CompiledFiles[0], "txt")` is null? Also, the extension should instead be changed to `.dm`.

wixoaGit (changes requested):

DMCompiler/Program.cs
ike709:
```suggestion
                string output = Path.ChangeExtension(CompiledFiles[0], "txt") ?? Path.Join(System.AppDomain.CurrentDomain.BaseDirectory, "preprocessor_dump.dm");
```

ike709: Practically speaking, no. But the IDE will complain about a possible null without this.

ike709:
Good to go

DMCompiler/Program.cs
```diff
@@ -14,13 +16,29 @@
...
-            DMPreprocessor preprocessor = Preprocess(args);
+            DMPreprocessor preprocessor = Preprocess(CompiledFiles);
+            if (HasArgument("--dump-preprocessor"))
+            {
+                StringBuilder result = new();
+                foreach (Token t in preprocessor.GetResult()) {
+                    result.Append(t.Text);
+                }
+
+                string output = Path.ChangeExtension(CompiledFiles[0], "txt") ?? Path.Join(System.AppDomain.CurrentDomain.BaseDirectory, "preprocessor_dump.dm");
```
wixoaGit, line 36:
```suggestion
                string output = Path.ChangeExtension(CompiledFiles[0], "dm") ?? Path.Join(System.AppDomain.CurrentDomain.BaseDirectory, "preprocessor_dump.dm");
```

wixoaGit (changes requested):

ike709:
Woopsie

wixoaGit (approved):

PR #322, opened 2021-08-09
[Robust] Add localhost congroup controller

Zumorica:
This allows clients connected through localhost to have all admin permissions when the server is compiled in DEBUG, therefore allowing you to access server-side ViewVariables and SCSI (once #321 is merged) when testing the server & client locally.
Ideally, in the future a more advanced system can be implemented that allows hosts to give some accounts certain permissions.
Also adds VV attributes to a few DMISpriteComponent properties.

### DIFF
--- a/Content.Server/DMISpriteComponent.cs
+++ b/Content.Server/DMISpriteComponent.cs
@@ -5,6 +5,7 @@ using Robust.Shared.GameObjects;
 using Robust.Shared.Maths;
 using Robust.Shared.Players;
 using Robust.Shared.Utility;
+using Robust.Shared.ViewVariables;
 
 namespace Content.Server {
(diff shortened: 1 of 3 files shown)


wixoaGit:
What's the difference between this and the console.loginlocal cvar?

Zumorica:
console.loginlocal is [specific to SS14](https://github.com/space-wizards/space-station-14/blob/b2322864e9b4423451c1d921a887953926d09bfc/Content.Shared/CCVar/CCVars.cs#L135), not RobustToolbox.
The `server_config.toml` file is a bit outdated, yeah...

Zumorica:
I guess, if you wanted, I could make this use a CVar just like SS14, and remove the `#if DEBUG`

PR #331, opened 2021-08-20
Adds score and medal procs

### LINKED ISSUE #287: `world/GetMedal()` not implemented
http://www.byond.com/docs/ref/#/world/proc/GetMedal

### LINKED ISSUE #300: `world/SetScores()` not implemented
http://www.byond.com/docs/ref/#/world/proc/SetScores

ike709:
Closes #287 and closes #300


ike709:
According to the docs these all return null if they fail so we shouldn't need to crash them.

wixoaGit:
Should this close the issues if it doesn't actually provide any implementation?

ike709:
Well considering it's not really something we can actually implement, I'd say so.

PR #335, opened 2021-08-22
Implemented matrix/Invert()

### LINKED ISSUE #329: `matrix/Invert()` is not implemented
http://www.byond.com/docs/ref/#/matrix/proc/Invert

DamianX:
Closes #329

The test is currently failing because it looks like numbers in OpenDream are more precise than BYOND's.

Inverting a `matrix(1,2,3,4,5,6)` in BYOND results in:
`-1.66667 0.666667 1 1.33333 -0.333333 -2`
while in OpenDream it results in:
`-1.6666667 0.6666667 1 1.3333334 -0.33333334 -2`

Do I change the test to expect the output we get in OpenDream currently?
Do we want to mimic BYOND numbers exactly?


DMCompiler/DMStandard/Types/Matrix.dm
```diff
@@ -36,6 +36,26 @@
 		f += Matrix2.f
 		return src
 
+	proc/Invert()
+		var/determinant = a*e - d*b
+		if(!determinant)
+			CRASH("Invalid matrix")
```
martinlyra, line 42: Is this intended or does this also exist in Byond too? Fairly sure the official documentation says the matrix won't be changed by this proc when the matrix is not invertible.
```suggestion
			return src
```

DamianX: My implementation seems consistent with BYOND's behavior. It returns null if `Invert` runtimes, leaving the matrix itself unchanged.

wixoaGit:
> Do I change the test to expect the output we get in OpenDream currently?
Do we want to mimic BYOND numbers exactly?

Have the test expect OpenDream's output. I wouldn't expect a slightly higher precision to cause any problems so there's no need to exactly mimic byond here.

DamianX:
There's a problem: Inverting the matrix twice doesn't return the original matrix. Starting with `matrix(1,2,3,4,5,6)`, the test inverts it and then inverts it again. The result is: `0.99999994 1.9999999 2.9999998 3.9999998 5 5.9999995`, which is not quite the original. I'm not very well-versed in floating-point magic so I'm not sure how to proceed. Is it because the operations I use inside `Invert` introduce too many errors?

martinlyra:
> There's a problem: Inverting the matrix twice doesn't return the original matrix. Starting with `matrix(1,2,3,4,5,6)`, the test inverts it and then inverts it again. The result is: `0.99999994 1.9999999 2.9999998 3.9999998 5 5.9999995`, which is not quite the original. I'm not very well-versed in floating-point magic so I'm not sure how to proceed. Is it because the operations I use inside `Invert` introduce too many errors?

Honestly? There is not much else you can do about this. I think this might be the reason why Byond has less precision than usual, to make those irrational .9999999x's round up to integers. 

And because the error is only less than `10^-6`, I think it is negligible at best. But if it is going to be a problem. How about we add a `Round` proc to matrix for the intents and purposes of rounding the elements to integers to a desired precision?

DamianX:
I feel like we'd have to round everything if we do that, not just matrix operations, so that everything is consistent with each other and BYOND. Right?

wixoaGit:
The reason that the invert test is failing is due to a difference in how floats are handled in OpenDream. This issue would be best handled in another PR, could the test be left out for now?

DamianX:
I've commented out the test.

wixoaGit (approved):

PR #336, opened 2021-08-22
Adds static typing for proc returns

ike709:
Example snippet:
```
world/proc/fizz() as text
    if(prob(50))
        return "buzz"
    else
        return 5
```
`Error at (unknown location): /world.fizz(): Invalid return type Num, expected Text`

Opening as a draft so wixoa can tell me if I'm on the right track or not with how I'm implementing this.


ike709:
Shoving this here so it isn't lost in discord:
> Instead of having ActualReturnType, could you check the return value against ReturnType in VisitProcStatementReturn and error in there instead?
Would make it possible to error on the return token instead of the proc's definition in the future
And instead of switching on  statement.value you could have an override-able DMValueType Type property on DMExpression

ike709:
Suggestions applied

OpenDreamShared/Compiler/Parser.cs
```diff
@@ -80,7 +81,8 @@ public class Parser<SourceType> {
         }
 
         protected void Error(string message) {
-            Errors.Add(new CompilerError(_currentToken, message));
+            //rrors.Add(new CompilerError(_currentToken, message));
```
ZeWaka, line 84: :waa:

ike709: shhh it's for debugging

ike709:
TODO: Implement implicit return types on overrides here: https://github.com/wixoaGit/OpenDream/blob/master/DMCompiler/DM/Visitors/DMVisitorObjectBuilder.cs#L101

OpenDreamShared/Compiler/Parser.cs
ZeWaka: no the typo

ike709: oh I know yeah

OpenDreamShared/Compiler/CompilerError.cs
```diff
@@ -59,6 +59,7 @@ public class CompileErrorException : Exception {
 
         public CompileErrorException(string message) : base(message) {
             Error = new CompilerError(null, message);
+            //throw new Exception(message);
```
ike709, line 62:
```suggestion
```

DMCompiler/DM/Visitors/DMVisitorObjectBuilder.cs
```diff
@@ -105,7 +107,26 @@ class DMVisitorObjectBuilder : DMASTVisitor {
...
-                if (!procDefinition.IsOverride && dmObject.HasProc(procName)) {
+                if (procDefinition.IsOverride)
+                {
+                    if (dmObject is not null)
+                    {
+                        var parentProcs = dmObject.GetProcs(procName);
+                        if (parentProcs is not null && parentProcs.Count >= 1)
+                        {
+                            var parent = parentProcs[dmObject.GetProcs(procName).Count - 1];
+                            if (procDefinition.ReturnTypes.Equals(DMValueType.Anything))
```
wixoaGit, line 118: So this would cause the following code to error?
```
proc/return_num() as num
	return 9

return_num() as anything
	return "A"
```
I personally think this is a little confusing and shouldn't be allowed.

DMCompiler/DM/Visitors/DMVisitorObjectBuilder.cs
```diff
@@ -105,7 +107,26 @@ class DMVisitorObjectBuilder : DMASTVisitor {
...
 
-                if (!procDefinition.IsOverride && dmObject.HasProc(procName)) {
+                if (procDefinition.IsOverride)
+                {
+                    if (dmObject is not null)
+                    {
+                        var parentProcs = dmObject.GetProcs(procName);
+                        if (parentProcs is not null && parentProcs.Count >= 1)
+                        {
+                            var parent = parentProcs[dmObject.GetProcs(procName).Count - 1];
```
wixoaGit, line 117:
```suggestion
                            var parent = parentProcs[^1];
```

DMCompiler/DM/Visitors/DMVisitorProcBuilder.cs
```diff
@@ -137,9 +141,21 @@ class DMVisitorProcBuilder : DMASTVisitor {
         }
 
         public void VisitProcStatementReturn(DMASTProcStatementReturn statement) {
+
+
             if (statement.Value != null) {
-                DMExpression.Emit(_dmObject, _proc, statement.Value);
-            } else {
+                var expr = DMExpression.Emit(_dmObject, _proc, statement.Value);
+                if (!_proc.ReturnTypes.Equals(DMValueType.Anything) && expr.Path == new DreamPath("/list"))
```
wixoaGit, line 148:
```suggestion
                if (!_proc.ReturnTypes.HasFlag(DMValueType.Anything) && expr.Path == DreamPath.List)
```

OpenDreamRuntime/DreamValue.cs
```diff
@@ -98,6 +99,55 @@ public enum DreamValueType {
...
+                        type = DMValueType.Turf;
+                    }
+                    else if(TryGetValueAsDreamObjectOfType(new DreamPath("/area"), out _))
+                    {
+                        type = DMValueType.Area;
+                    }
+                    else
+                    {
+                        type = DMValueType.Anything;
+                    }
```
wixoaGit, line 137:
```suggestion
                    DreamObject dreamObject = GetValueAsDreamObject();

                    if(dreamObject.IsSubtypeOf(DreamPath.Mob))
                    {
                        type = DMValueType.Mob;
                    }
                    else if(dreamObject.IsSubtypeOf(DreamPath.Obj))
                    {
                        type = DMValueType.Obj;
                    }
                    else if(dreamObject.IsSubtypeOf(DreamPath.Turf))
                    {
                        type = DMValueType.Turf;
                    }
                    else if(dreamObject.IsSubtypeOf(DreamPath.Area))
                    {
                        type = DMValueType.Area;
                    }
                    else
                    {
                        type = DMValueType.Anything;
                    }
```

OpenDreamRuntime/DreamValue.cs
```diff
@@ -98,6 +99,55 @@ public enum DreamValueType {
             throw new Exception("Value " + this + " was not the expected type of " + type + "");
         }
 
+        public DMValueType GetDMValueType(out DMValueType type)
```
wixoaGit, line 102: Why does this both set an `out` parameter and return the value?

wixoaGit (changes requested):

ike709:
We don't need to deal with maintaining this right now, but I do intend to reimplement it when OpenDream is more usable.

It still serves as a valid proof of concept.

PR #343, opened 2021-08-25
Associative lists will no longer be allocated for non-associative lists

ike709:
Fixes a part of https://github.com/wixoaGit/OpenDream/issues/342

I didn't profile the memory usage but it is presumably better. Also might have gone overboard with the null checking but w/e.

### DIFF
--- a/OpenDreamRuntime/Objects/DreamList.cs
+++ b/OpenDreamRuntime/Objects/DreamList.cs
@@ -162,10 +162,7 @@ namespace OpenDreamRuntime.Objects {
             DreamList newList = new DreamList(Runtime);
             newList._values = _values.Union(other.GetValues()).ToList();
             var otherVals = other.GetAssociativeValues();
-            if (newList._associativeValues is null)
-            {
-                newList._associativeValues = new Dictionary<DreamValue, DreamValue>(otherVals?.Count ?? 0);
-            }
+            newList._associativeValues ??= new Dictionary<DreamValue, DreamValue>(otherVals?.Count ?? 0);
 
             if (otherVals != null)
                 foreach ((DreamValue key, DreamValue value) in otherVals)
--- a/OpenDreamRuntime/Objects/DreamList.cs
+++ b/OpenDreamRuntime/Objects/DreamList.cs
@@ -11,7 +11,7 @@ namespace OpenDreamRuntime.Objects {
         internal event DreamListBeforeValueRemovedEventHandler BeforeValueRemoved;
 
         private List<DreamValue> _values = new();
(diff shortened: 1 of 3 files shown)


OpenDreamRuntime/Objects/DreamList.cs
```diff
@@ -83,6 +83,10 @@ protected DreamList(DreamRuntime runtime)
             } else {
                 if (!ContainsValue(key)) _values.Add(key);
 
+                if (_associativeValues is null)
+                {
+                    _associativeValues = new Dictionary<DreamValue, DreamValue>(1);
+                }
```
PJB3005, line 89:
```suggestion
                _associativeValues ??= new Dictionary<DreamValue, DreamValue>(1);
```

OpenDreamRuntime/Objects/DreamList.cs
```diff
@@ -160,9 +164,18 @@ protected DreamList(DreamRuntime runtime)
         public DreamList Union(DreamList other) {
             DreamList newList = new DreamList(Runtime);
             newList._values = _values.Union(other.GetValues()).ToList();
-            foreach ((DreamValue key, DreamValue value) in other.GetAssociativeValues()) {
-                newList._associativeValues[key] = value;
+            var otherVals = other.GetAssociativeValues();
+            if (newList._associativeValues is null)
+            {
+                newList._associativeValues = new Dictionary<DreamValue, DreamValue>(otherVals?.Count ?? 0);
             }
```
PJB3005, line 171:
```suggestion
            newList._associativeValues ??= new Dictionary<DreamValue, DreamValue>(otherVals?.Count ?? 0);
```

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -354,18 +354,21 @@ static class DMOpcodeHandlers {
                 Dictionary<DreamValue, DreamValue> argListNamedValues = argList.GetAssociativeValues();
 
                 foreach (DreamValue value in argListValues) {
-                    if (!argListNamedValues.ContainsKey(value)) {
+                    if (argListNamedValues != null && !argListNamedValues.ContainsKey(value)) {
```
PJB3005, line 357: I feel like it may be better to make `DreamList` have `ContainsKey` helpers and similar so that usage code doesn't have to keep in mind that the dict can be null. This sounds like it'd be very error prone otherwise.

Also applies to other changes below.

wixoaGit: Agreed

wixoaGit (approved):

PR #353, opened 2021-08-26
Implemented length_char

### LINKED ISSUE #268: `length_char()` not implemented

DamianX:
Closes #268

~~Also changes the existing `length` proc to return the UTF-8 byte count, to match BYOND.~~ `length` stays the same


ZeWaka:
> I'm of the opinion that any text procs working in bytes will just lead to unexpected behavior and errors, and that if we *really* need it we can add procs like `length_bytes()`

Yeah I think this is probably the best way forward - even if breaking compat. with BYOND

PJB3005:
> Yeah I think this is probably the best way forward - even if breaking compat. with BYOND

It's absolutely not. Working with *just* code points is not sufficient for any complex text operations, so trying to hide this fact is just counter productive.

ike709:
After talking with wixoa and PJB, everyone seems in agreement that strings can all be left as UTF-16. It reduces the complexity of everything and nobody is leveraging the UTF-8 support in BYOND because it's still quite new.

DamianX:
Reverted the change to `length`

wixoaGit (approved):

PR #355, opened 2021-08-27
Implement cal()() FFI.

PJB3005:
Man modern C# makes this so damn easy.


wixoaGit:
The other functionality is already implemented. Although, someone brought up that you can use a string instead of a path to call a global proc (`call("proc/globalproc")()`). This implementation would either try to load a "proc/globalproc" DLL or runtime because there is no second argument.

PR #375, opened 2021-09-06
[Robust] Removes System.Console sandbox violations

ike709:


ike709:
See https://github.com/wixoaGit/OpenDream/issues/374

Content.Client/Interface/DreamInterfaceManager.cs
```diff
@@ -51,13 +52,13 @@ class DreamInterfaceManager : IDreamInterfaceManager {
 
             if (dmfParser.Warnings.Count > 0) {
                 foreach (CompilerWarning warning in dmfParser.Warnings) {
-                    Console.WriteLine(warning);
+                    Logger.Warning(warning.ToString());
```
ike709, line 55: I am unsure if these should be `.ToString()` or just `.Message`

wixoaGit: `CompilerWarning` overrides `ToString()` to return `"Warning at " + location + ": " + Message;`. `ToString()` should be used here.

wixoaGit (approved):

PR #376, opened 2021-09-07
Adds unit testing to the RT branch

ike709:


ike709:
The errors are in CefGlue. I do not know what to do about that except we might have to fork it.

wixoaGit (approved):

PR #379, opened 2021-09-08
Adds italic, b, and bold as valid escape sequences

ike709:
Move NTstation from 220 to 186 errors with one easy trick.


wixoaGit (approved):

wixoaGit:
Why are these nowhere to be seen in the reference?

PR #380, opened 2021-09-08
Improves raw string handling

ike709:
Raw strings containing newlines no longer error. This fixes 21 errors on modern TG.

### DIFF
--- a/OpenDreamShared/Compiler/DMPreprocessor/DMPreprocessorLexer.cs
+++ b/OpenDreamShared/Compiler/DMPreprocessor/DMPreprocessorLexer.cs
@@ -197,13 +197,31 @@ namespace OpenDreamShared.Compiler.DMPreprocessor {
                         textBuilder.Append('@');
                         textBuilder.Append(delimiter);
 
-                        if (delimiter == '{') delimiter = '}';
+                        if (delimiter == '{')
+                        {
+                            delimiter = '}';
+                            char prev;
+                            do
+                            {
+                                prev = c;
+                                c = Advance();
 
-                        do {
-                            c = Advance();
+                                textBuilder.Append(c);
+                                if (c == '}' && prev == '\"')
+                                {
+                                    break;
+                                }
+                            } while (!AtEndOfSource);
+                        }
+                        else
+                        {
+                            do
+                            {
+                                c = Advance();
+                                textBuilder.Append(c);
+                            } while (c != delimiter);
+                        }
 
-                            textBuilder.Append(c);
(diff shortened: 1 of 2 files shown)


ike709:
Slight issue: This doesn't handle raw strings that contain `{}` inside of them

OpenDreamShared/Compiler/DMPreprocessor/DMPreprocessorLexer.cs
```diff
@@ -197,15 +197,56 @@ class DMPreprocessorLexer : TextLexer {
...
+                            {
+                                startIdx++;
+                                stopLength = startIdx + 1;
+                            }
+
+                            if (textBuilder[textBuilder.Length-2] == '"')
+                            {
+                                stopLength = startIdx + 2;
+                            }
+                        }
```
wixoaGit, line 246: It'd be better to instead have two separate `CreateToken` calls after building the text for each type of raw string.
`token = CreateToken(TokenType.DM_Preproc_ConstantString, text, text.Substring(2, text.Length - 3));` after line 227, and
`token = CreateToken(TokenType.DM_Preproc_ConstantString, text, text.Substring(3, text.Length - 5));` after line 219.

OpenDreamShared/Compiler/DMPreprocessor/DMPreprocessorLexer.cs
```diff
@@ -197,15 +197,56 @@ class DMPreprocessorLexer : TextLexer {
                         textBuilder.Append('@');
                         textBuilder.Append(delimiter);
 
-                        do {
-                            c = Advance();
+                        if (delimiter == '{')
```
wixoaGit, line 200: A long raw string needs to start with `@{"`, not just `@{`

OpenDreamShared/Compiler/DMPreprocessor/DMPreprocessorLexer.cs
```diff
@@ -197,15 +197,56 @@ class DMPreprocessorLexer : TextLexer {
...
+                        {
+                            delimiter = '}';
+                            char prev;
+                            do
+                            {
+                                prev = c;
+                                c = Advance();
+
+                                textBuilder.Append(c);
+                                if (c == '}' && prev == '\"')
```
wixoaGit, line 210: This would break with `@{"}`

wixoaGit (changes requested):

ike709:
I cba to finish this but someone should.

Issue #121, opened 2021-05-26
`~=` and `~!` operators not implemented


ZeWaka:
http://www.byond.com/docs/ref/#/operator/~=
`http://www.byond.com/docs/ref/#/operator/~!`

ike709:
Implemented for lists in #384

PR #384, opened 2021-09-08
Implements list equivalence

ike709:
See #121

### DIFF
--- a/OpenDreamShared/Compiler/DM/DMParser.cs
+++ b/OpenDreamShared/Compiler/DM/DMParser.cs
@@ -1343,6 +1343,25 @@ namespace OpenDreamShared.Compiler.DM {
             return expression;
         }
 
+        public DMASTExpression ExpressionEquivalence()
+        {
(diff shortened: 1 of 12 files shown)


OpenDreamShared/Compiler/DM/DMParser.cs
```diff
@@ -1343,6 +1343,25 @@ public partial class DMParser : Parser<Token> {
             return expression;
         }
 
+        public DMASTExpression ExpressionEquivalence()
```
wixoaGit, line 1346: Equivalence should have the same operator precedence as `==` and `!=`. You can do this by instead adding them to the `ExpressionComparison()` method.

wixoaGit (changes requested):

OpenDreamShared/Compiler/DM/DMParser.cs
ike709: Fixed

ike709:
Good to go

wixoaGit (approved):

PR #386, opened 2021-09-09
Adds support for `var/thing[]` lists in object definitions

ike709:
Fixes some number of errors in all codebases

### DIFF
--- a/OpenDreamShared/Compiler/DM/DMParser.cs
+++ b/OpenDreamShared/Compiler/DM/DMParser.cs
@@ -109,7 +109,29 @@ namespace OpenDreamShared.Compiler.DM {
                             Whitespace();
 
                             DMASTExpression value;
-                            if (Check(TokenType.DM_Equals)) {
+                            if (Check(TokenType.DM_LeftBracket))
+                            {
+                                //Type information
+                                if (varPath.FindElement("list") != 0)
+                                {
+                                    var idx = varPath.PathString.IndexOf("var/", StringComparison.Ordinal) + 4;
+                                    varPath.PathString = varPath.PathString.Insert(idx, "list/");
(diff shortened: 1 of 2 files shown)


OpenDreamShared/Compiler/DM/DMParser.cs
```diff
@@ -109,7 +109,29 @@ public partial class DMParser : Parser<Token> {
                             Whitespace();
 
                             DMASTExpression value;
-                            if (Check(TokenType.DM_Equals)) {
+                            if (Check(TokenType.DM_LeftBracket)) //TODO: Multidimensional lists
+                            {
+                                //Type information
+                                if (varPath.FindElement("list") != 0)
```
ZeWaka, line 115: `var/list[]` - not typed by your pathstring stuff?

ike709: lists named `list` already don't work for other reasons

OpenDreamShared/Compiler/DM/DMParser.cs
```diff
@@ -496,7 +514,7 @@ public partial class DMParser : Parser<Token> {
                     //TODO: Multidimensional lists
                     if (Check(TokenType.DM_LeftBracket)) {
                         //Type information
-                        if (varPath.Path.FindElement("list") != 0) {
+                        if (varPath is not null && !varPath.Path.IsList()) {
```
wixoaGit, line 517:
```suggestion
                        if (varPath is not null && !varPath.Path.IsDescendantOf(DreamPath.List)) {
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #387, opened 2021-09-09
Implements implicit `in world` loops

ike709:
Stuff like `for(var/mob/M)`

### DIFF
--- a/DMCompiler/DM/Visitors/DMVisitorProcBuilder.cs
+++ b/DMCompiler/DM/Visitors/DMVisitorProcBuilder.cs
@@ -218,8 +218,7 @@ namespace DMCompiler.DM.Visitors {
                         //This is terrible but temporary
                         //TODO: See https://github.com/wixoaGit/OpenDream/issues/50
                         var obj = DMObjectTree.GetDMObject(varDeclaration.Type.Value);
-                        var list = statementForList.List as DMASTIdentifier;
-                        if (list is not null && list.Identifier == "world" && !obj.IsSubtypeOf(new DreamPath("atom/")))
+                        if (statementForList.List is DMASTIdentifier list && list.Identifier == "world" && !obj.IsSubtypeOf(DreamPath.Atom))
                         {
                             var warn = new CompilerWarning(null, "Cannot currently loop 'in world' for non-ATOM types");
                             Program.Warning(warn);
--- a/DMCompiler/DM/Visitors/DMVisitorProcBuilder.cs
+++ b/DMCompiler/DM/Visitors/DMVisitorProcBuilder.cs
@@ -217,7 +217,7 @@ namespace DMCompiler.DM.Visitors {
                     {
                         //This is terrible but temporary
                         //TODO: See https://github.com/wixoaGit/OpenDream/issues/50
-                        var obj = DMObjectTree.GetDMObject(new DreamPath(varDeclaration.Type.ToString()));
+                        var obj = DMObjectTree.GetDMObject(varDeclaration.Type.Value);
                         var list = statementForList.List as DMASTIdentifier;
                         if (list is not null && list.Identifier == "world" && !obj.IsSubtypeOf(new DreamPath("atom/")))
                         {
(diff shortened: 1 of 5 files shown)


wixoaGit:
These loops are not implicit `in world` loops. They loop through every existing instance of a type, wether or not it's in the world. For example, `for (var/client/C)` loops through every client, despite them not being in `world.contents`.

https://github.com/wixoaGit/OpenDream/issues/50#issuecomment-865178884

ike709:
According to the docs, a for loop will by default loop through the world. Clients and datums are a special case and a separate issue.

image

wixoaGit:
I suspect the docs aren't being entirely correct here. It works for _any_ type, there is no special case for just clients and datums.

I'd rather this not silently create a non-working loop for types that aren't atoms.

OpenDreamShared/Compiler/DM/DMParser.cs
```diff
@@ -781,50 +781,66 @@ public partial class DMParser : Parser<Token> {
...
-                    DMASTProcBlockInner body = ProcBlock();
-                    if (body == null) {
-                        DMASTProcStatement statement = ProcStatement();
-
-                        if (statement == null) Error("Expected body or statement");
-                        body = new DMASTProcBlockInner(new DMASTProcStatement[] { statement });
+                    //Implicit "in world"
+                    if (rangeBegin is DMASTConstantNull && rangeEnd is null && defaultStep)
+                    {
+                        var path = new DreamPath(variableDeclaration.Type.ToString());
```
ike709, line 815: Not sure why but the compiler hated it when I used `variableDeclaration.Type` directly.

DMCompiler/DM/Visitors/DMVisitorProcBuilder.cs
```diff
@@ -212,7 +213,18 @@ class DMVisitorProcBuilder : DMASTVisitor {
                     _proc.BreakIfFalse();
 
                     DMASTProcStatementVarDeclaration varDeclaration = statementForList.Initializer as DMASTProcStatementVarDeclaration;
-                    if (varDeclaration != null && varDeclaration.Type != null) {
+                    if (varDeclaration != null && varDeclaration.Type != null)
+                    {
+
```
ike709, line 218:
```suggestion
```

OpenDreamShared/Compiler/DM/DMParser.cs
```diff
@@ -781,50 +781,60 @@ public partial class DMParser : Parser<Token> {
...
-                        body = new DMASTProcBlockInner(new DMASTProcStatement[] { statement });
-                    }
-
-                    return new DMASTProcStatementForStandard(initializer, comparator, incrementor, body);
+                    return new DMASTProcStatementForStandard(initializer, comparator, incrementor, GetForBody());
                 } else if (variableDeclaration != null) {
                     DMASTExpression rangeBegin = variableDeclaration.Value;
                     Whitespace();
-                    Consume(TokenType.DM_To, "Expected 'to'");
+                    if (rangeBegin is not DMASTConstantNull)
```
wixoaGit, line 788: This allows you to do an implicit world loop if you do `for (var/atom/A = null)`. This does not work in byond.

OpenDreamShared/Compiler/DM/DMParser.cs
```diff
@@ -781,50 +781,60 @@ public partial class DMParser : Parser<Token> {
...
+                    {
+                        Consume(TokenType.DM_To, "Expected 'to'");
+                    }
+
                     Whitespace();
                     DMASTExpression rangeEnd = Expression();
-                    if (rangeEnd == null) Error("Expected an expression");
+                    if (rangeBegin is not DMASTConstantNull && rangeEnd == null) Error("Expected an expression");
```
wixoaGit, line 795: This would allow `for (var/i = null to)`

wixoaGit (changes requested):

wixoaGit (approved):

PR #402, opened 2021-09-12
Remove Check() parser array allocations.

PJB3005:
This pattern (`ReadOnlySpan<T> x = new T[] { ... }`) is optimized by Roslyn to create a Span from assembly image memory instead of allocating, if T : byte. I tested and it works for enums too.

This saves ~700 MB of allocations parsing /vg/.

### DIFF
--- a/OpenDreamClient/Interface/DreamInterface.cs
+++ b/OpenDreamClient/Interface/DreamInterface.cs
@@ -258,7 +258,7 @@ namespace OpenDreamClient.Interface {
                     string value = query.GetValues(attribute)[^1];
 
                     Token attributeValue = new DMFLexer(null, value).GetNextToken();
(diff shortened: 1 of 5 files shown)


OpenDreamClient/Interface/DreamInterface.cs
```diff
@@ -258,7 +258,7 @@ class DreamInterface {
                     string value = query.GetValues(attribute)[^1];
 
                     Token attributeValue = new DMFLexer(null, value).GetNextToken();
-                    if (DMFParser.ValidAttributeValueTypes.Contains(attributeValue.Type)) {
+                    if (Array.IndexOf(DMFParser.ValidAttributeValueTypes, attributeValue.Type) >= 0) {
```
ZeWaka, line 261: actually based wow, never thought about it this way

wixoaGit (approved):

Issue #294, opened 2021-08-03
`filter()` not implemented


ike709:
See also `atom/list/filters`

http://www.byond.com/docs/ref/#/proc/filter

PR #403, opened 2021-09-12
Implented filters

ShiftyRail:
... except not really.

Currently, none of the filters are implemented on the renderer's side because I have absolutely no idea what I am doing. 
All this does is make it so the compiler can understand what you mean by "filter" and replicate BYOND's behaviour.

Smarter people than me will probably be able to do that.

If you do not specify a filter type, or an invalid filter type (like "asf"), it silently fails and returns null.

If you do specify a valid filter type, but gives it an invalid argument name, it will runtime and tell you that this argument name is invalid. 

The way they work is pretty opaque. A filter is not a datum. You cannot explicitly type it either. But once it is created, you *can* access some of its variables with the `:` operator, and, in the event of a filter attached to an atom (although not one attached to an image, or not one existing in vacuum), you can alter them and change them, even animate them.

To replicate this, I made filters their own object with a typepath in `/__filter` to allow users to create a `/filter/` path if they want. I gave them vars on a subtype-per-type basis.

As none of the filters are implemented server-side, this will just runtime when trying to create a filter. 🤡 

This is related to #294, but doesn't really close it since it just make it crash at runtime rather than failing to compile.

I made the atom's filters a list, although they are not exactly that in BYOND.

Tell me I did anything wrong.


ShiftyRail:
Filters are now a "byond native proc" which takes named arguments, throw null if the type is invalid, and throws a runtime if the type is valid but the named arguments aren't. This isn't really what happens with the real BYOND, who will not accepted any kind of non-named arguments at compile time.

I am a bit stuck as to what to do next. Ideally you create an object on the C# side. I tried to look at `DreamMetaObjectList` for reference but I am not sure if the right place to look for inspiration.

ike709:
Dead PR?

PR #407, opened 2021-09-13
DM Error Handling

ike709:
This is `try`/`catch` support, basically.

TODO:

- [ ] Try/catch blocks compile
- [ ] Only the try block executes by default
- [ ] Add the `/datum/exception` type
- [ ] Properly handle `catch(var/exception/varname)` rather than using placeholder copypasted proc definition code
- [ ] If an exception is thrown in the try block, properly move to the catch block
- [ ] Implement support for `world/Error()` (future PR?)
- [ ] `EXCEPTION()` macro
- [ ] Implement `throw`

### DIFF
--- a/OpenDreamShared/Compiler/CompilerError.cs
+++ b/OpenDreamShared/Compiler/CompilerError.cs
@@ -13,6 +13,7 @@ namespace OpenDreamShared.Compiler {
             Token = token;
             Message = message;
             StackTrace = new StackTrace(true);
+            throw new Exception(message);
         }
 
         public override string ToString() {
--- a/OpenDreamShared/Compiler/DM/DMParser.cs
+++ b/OpenDreamShared/Compiler/DM/DMParser.cs
@@ -513,6 +513,7 @@ namespace OpenDreamShared.Compiler.DM {
                 if (procStatement == null) procStatement = While();
                 if (procStatement == null) procStatement = DoWhile();
                 if (procStatement == null) procStatement = Switch();
+                if (procStatement == null) procStatement = TryCatch();
 
                 if (procStatement != null) {
                     Whitespace();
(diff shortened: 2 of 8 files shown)


OpenDreamShared/Compiler/CompilerError.cs
```diff
@@ -13,6 +13,7 @@ public struct CompilerError {
             Token = token;
             Message = message;
             StackTrace = new StackTrace(true);
+            throw new Exception(message);
```
ike709, line 16:
```suggestion
```

DMCompiler/DM/Visitors/DMProcBuilder.cs
```diff
@@ -414,5 +415,49 @@ class DMProcBuilder {
...
+                            DMExpression.Emit(_dmObject, _proc, parameter.Value, parameter.ObjectType);
+                        } catch (CompileErrorException e) {
+                            Program.Error(e.Error);
+                        }
+                        _proc.Assign();
+
+                        _proc.AddLabel(afterDefaultValueCheck);
+                    }
+                }
+            }
```
wixoaGit, line 453: I tested this code in byond:
```
try
	throw null
catch(var/exception/E = 3)
	usr << isnull(E)
	usr << E
```
This compiled fine, and output:
```
1

```
It appears that default values don't do anything

OpenDreamShared/Compiler/DM/DMParser.cs
```diff
@@ -1016,6 +1017,47 @@ public partial class DMParser : Parser<Token> {
...
+                Newline();
+                Consume(TokenType.DM_Catch, "Expected catch");
+
+                // catch(var/exception/E)
+                // TODO: handle correctly
+                DMASTDefinitionParameter[] parameters = null;
+                if (Check(TokenType.DM_LeftParenthesis))
+                {
+                    BracketWhitespace();
+                    parameters = DefinitionParameters();
```
wixoaGit, line 1041: You cannot have more than one var definition in a catch. It may be better to instead parse this with `ProcVarDeclaration(allowMultiple: false)`

wixoaGit (changes requested):

wixoaGit (approved):

PR #414, opened 2021-09-19
for loop edge case handling

mumencoder:
this happens occasionally in the unmodified ss13 repos, where the statement expression has side effects as well


wixoaGit:
I'd rather not use any lookahead if possible (`SavePosition()`/`RestorePosition()`). Can this be done another way?

ike709:
This fixes #527

wixoaGit (changes requested):

wixoaGit (approved):

PR #418, opened 2021-09-20
Converts Whitespace() to use a readonlyspan

ike709:
Should reduce memory usage


wixoaGit (approved):

PJB3005 (commented):
This doesn't do anything. The compiler already recognizes that `Check()` takes a span so the optimization already applied.

ike709:
/shrug

Consistency, then.

Issue #374, opened 2021-09-06
[Robust] List of sandbox violations


ike709:
EDIT: Once every linked PR is merged, the only remaining problems are the usage of reflection and the DMI parser needing to be moved to the server.

These all need to be fixed so OpenDream can go on the SS14 hub:

- [x] [ERRO] res.typecheck: Sandbox violation: Access to type not allowed: [System.Drawing.Primitives]System.Drawing.Point
- [x] [ERRO] res.typecheck: Sandbox violation: Access to type not allowed: [System.Runtime]System.Reflection.CustomAttributeExtensions
- [x] [ERRO] res.typecheck: Sandbox violation: Access to type not allowed: [System.Web.HttpUtility]System.Web.HttpUtility
- [x] [ERRO] res.typecheck: Sandbox violation: Access to type not allowed: [System.Runtime]System.Reflection.FieldInfo
- [x] [ERRO] res.typecheck: Sandbox violation: Access to type not allowed: [System.IO.Compression]System.IO.Compression.DeflateStream
- [x] [ERRO] res.typecheck: Sandbox violation: Access to type not allowed: [System.Drawing.Primitives]System.Drawing.Size
- [x] [ERRO] res.typecheck: Sandbox violation: Access to type not allowed: [System.IO.FileSystem]System.IO.File
- [x] [ERRO] res.typecheck: Sandbox violation: Access to type not allowed: [System.Runtime]System.IO.Path
- [x] [ERRO] res.typecheck: Sandbox violation: Access to type not allowed: [System.Console]System.Console
- [x] [ERRO] res.typecheck: Sandbox violation: Access to type not allowed [System.ComponentModel]System.ComponentModel.CancelEventArgs
- [x] [ERRO] res.typecheck: Sandbox violation: Access to type not allowed:[System.Collections.Specialized]System.Collections.Specialized.NameValueCollection
- [x] [ERRO] res.typecheck: Sandbox violation: Access to type not allowed: [System.Runtime]System.IO.BinaryReader
- [x] [ERRO] res.typecheck: Sandbox violation: Access to method not allowed: string [System.Runtime]System.Convert.ToString(char)
- [x] [ERRO] res.typecheck: Sandbox violation: Access to type not allowed: [System.Runtime]System.IO.SeekOrigin
- [x] [ERRO] res.typecheck: Sandbox violation: Access to type not allowed: [System.Text.Json]System.Text.Json.Serialization.JsonIgnoreAttribute
[...]

PR #420, opened 2021-09-24
[Robust] Fixes a lot of sandbox violations

ike709:
See #374 

This fixes basically everything except reflection.

Also needs an engine PR to add a few things to the whitelist.


OpenDreamShared/Dream/IconAppearance.cs
```diff
@@ -144,7 +144,8 @@ public enum AppearanceProperty {
                     color += "ff";
                 }
 
-                Color = Convert.ToUInt32(color, 16);
+                Color = uint.Parse(color, System.Globalization.NumberStyles.HexNumber);
```
ike709, line 147: Someone tell me if this is actually the same result or not.

wixoaGit: `Convert.ToUInt32` accepts null while `uint.Parse` instead throws an exception. That wouldn't cause any issues here.

DMCompiler/Compiler/DM/DMASTHelper.cs
```diff
@@ -5,11 +5,11 @@
 using System.Reflection;
 using System.Text;
 
-namespace OpenDreamShared.Compiler.DM {
+namespace DMCompiler.Compiler.DM {
 
 
     public static partial class DMAST {
-        private static List<Type> printable_field_types = new() { typeof(string), typeof(Dream.DreamPath), typeof(int), typeof(float) };
+        private static List<Type> printable_field_types = new() { typeof(string), typeof(int), typeof(float) };
```
wixoaGit, line 12: You don't need to remove DreamPath from this, it can be accessed from the `OpenDreamShared.Dream` namespace.

DMCompiler/Compiler/DM/DMASTHelper.cs
```diff
@@ -5,11 +5,11 @@
 using System.Reflection;
 using System.Text;
```
ike709, line 7:
```suggestion
using System.Text;
using OpenDreamShared.Dream;

```

DMCompiler/Compiler/DM/DMASTHelper.cs
ike709, line 13:
```suggestion
        private static List<Type> printable_field_types = new() { typeof(string), typeof(Dream.DreamPath), typeof(int), typeof(float) };
```

ike709:
There

wixoaGit (approved):

PR #426, opened 2021-10-02
preprocessor refactor

mumencoder:
has a few things going for it so far:

fully tokenizes most/all of the major SS13 codebases (output not verified yet)
same token output as master for the testing version of tgstation
no token buildup in the preprocessor, uses 20-30MB memory throughout the whole tokenization phase


redmoogle:
[output.txt](https://github.com/wixoaGit/OpenDream/files/7447809/output.txt)
took 10 minutes and came out with this

PR #435, opened 2021-10-17
Use InvariantCulture system culture

pali6:
If your default system culture uses `,` as the decimal separator it breaks both the compiler and the server currently (mostly because of `Single.TryParse`). ~~Client works but why take the chances.~~ Client unaffected by this PR now.

I'm unsure if this is the best place where to put this so please yell at me if it should be moved.

### DIFF
--- a/OpenDreamClient/Program.cs
+++ b/OpenDreamClient/Program.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Globalization;
 
 namespace OpenDreamClient {
     class Program {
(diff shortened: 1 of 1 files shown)


wixoaGit:
> Client works but why take the chances.

The client has a few places where the user can input things. I don't believe it has anything that would be affected by this (unless you're on the RT branch which has a more correct number prompt that parses floats), but this may be a good reason to not set the default.

pali6:
Fair, that makes sense. Removed the client bit.

PR #439, opened 2021-10-27
global procs added

mumencoder:
test case:
```
#include "map.dmm"
#include "interface.dmf"

var/a = "a"
var/static/b = "b"

/proc/fn()
    return "fn"

/world/proc/fn()
    return "worldfn"

/world/New()
    var/a = "aloc"
    var/b = "bloc"

    //world.log << "[a]"
    //world.log << "[global.a]"
    //world.log << "[b]"
    //world.log << "[global.b]"
    //world.log << "[fn()]"
    //world.log << "[global.fn()]"
    //world.log << "[abs(-1)]"

    ASSERT(a == "aloc")
    ASSERT(global.a == "a")
    ASSERT(b == "bloc")
    ASSERT(global.b == "b")
    ASSERT(fn() == "worldfn")
    ASSERT(global.fn() == "fn")
    ASSERT(abs(-1) == 1)
    shutdown()
```


ike709:
Fixes https://github.com/wixoaGit/OpenDream/issues/515 ?

OpenDreamRuntime/Procs/Native/DreamProcNative.cs
```diff
@@ -8,6 +11,50 @@ static class DreamProcNative {
...
 
             DreamObjectDefinition root = objectTree.GetObjectDefinition(DreamPath.Root);
+
+            Assembly asm = Assembly.GetExecutingAssembly();
+            
+            var q = from t in Assembly.GetExecutingAssembly().GetTypes()
+                    where t.Name == "DreamProcNativeRoot" && t.IsClass && t.Namespace == "OpenDreamRuntime.Procs.Native"
+                    select t;
+
+            foreach (var t in q)
```
wixoaGit, line 21: There's a more direct way to get the type of a class, `typeof(DreamProcNativeRoot)`

OpenDreamRuntime/Procs/Native/DreamProcNative.cs
```diff
@@ -8,6 +11,50 @@ static class DreamProcNative {
...
+                        } catch { }
+
+                        try {
+                            var func2 = methodI.CreateDelegate<AsyncNativeProc.HandlerFn>();
+                            proc = new AsyncNativeProc(attr?.Name, null, argumentNames, null, defaultArgumentValues, func2.Invoke);
+                        } catch { }
+
+                        if (proc == null) {
+                            throw new System.Exception($"Cannot resolve signature for proc '{attr?.Name}' in DreamProcNativeRoot");
+                        }
```
wixoaGit, line 51: You can do `methodI.GetCustomAttribute<AsyncStateMachineAttribute>() != null` to check if the method is async
```suggestion
                        if (methodI.GetCustomAttribute<AsyncStateMachineAttribute>() != null) {
                            var func = methodI.CreateDelegate<AsyncNativeProc.HandlerFn>();
                            proc = new AsyncNativeProc(attr?.Name, null, argumentNames, null, defaultArgumentValues, func.Invoke);
                        } else {
                            var func = methodI.CreateDelegate<NativeProc.HandlerFn>();
                            proc = new NativeProc(attr?.Name, null, argumentNames, null, defaultArgumentValues, func);
                        }
```

OpenDreamRuntime/Procs/Native/DreamProcNative.cs
```diff
@@ -8,6 +11,50 @@ static class DreamProcNative {
...
+
+            Assembly asm = Assembly.GetExecutingAssembly();
+            
+            var q = from t in Assembly.GetExecutingAssembly().GetTypes()
+                    where t.Name == "DreamProcNativeRoot" && t.IsClass && t.Namespace == "OpenDreamRuntime.Procs.Native"
+                    select t;
+
+            foreach (var t in q)
+            {
+                foreach (var memberI in t.FindMembers(MemberTypes.Method, BindingFlags.Public | BindingFlags.Static, null, null)) {
```
wixoaGit, line 23: You can use `t.GetMethods(BindingFlags.Public | BindingFlags.Static)` here to avoid having to get `methodI` in a second step later on.

OpenDreamRuntime/Procs/Native/DreamProcNative.cs
```diff
@@ -8,6 +11,50 @@ static class DreamProcNative {
...
+                    if (memberI.Name.StartsWith("NativeProc")) {
```
wixoaGit, line 24: This should check if the method has a `DreamProcAttribute` instead of checking its name.

wixoaGit (changes requested):

DMCompiler/DM/Visitors/DMObjectBuilder.cs
```diff
@@ -112,23 +114,33 @@ class DMObjectBuilder {
...
-                if (procDefinition.Body != null)
-                {
-                    foreach (var stmt in GetStatements(procDefinition.Body))
-                    {
+                if (procDefinition.ObjectPath == null) {
+                    if (DMObjectTree.HasNamedGlobalProc(procDefinition.Name)) {
+                        throw new CompileErrorException(new CompilerError(procDefinition.Location, $"{procDefinition.Name} is already defined in global scope"));
+                    }
+                    int gproc_id = DMObjectTree.CreateNamedGlobalProc(procDefinition.Name, proc);
+                    string gproc_internal = $"USRPROC${procDefinition.ObjectPath}${proc.Name}";
```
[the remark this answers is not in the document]
mumencoder: proc.Name may not be unique if other types of global procs are added later (like ones defined on objects that still store themselves in the global procs list, or ones that users won't normally need access to and the name will never be displayed/relevant but still need to be namespaced somehow wrt global ID resolution)

That said, I probably still should have added the Name field to the JSON so now ProcDefinitionJson stores both names and the runtime will pass proc.Name when creating a global DMProc.

OpenDreamRuntime/Procs/Native/DreamProcNative.cs
mumencoder: My intent was to match a signature with a constructor in a uniform way, if other signatures/constructors are added later or if more control over the exception that is thrown for failure was desirable (this code might be eventually moved out into a more reusable function). Going to commit this suggestion since these features are not too important right now and the original does throw out exception information for potential debugging.

ike709:
This PR has some bugs, see discord.

ike709:
Most issues fixed, but this does currently break `rgb()` in obj defs:
`Error at code\game\objects\structures\stool_bed_chair_nest\chairs.dm:128:14: Invalid initial value for "color"`

OpenDreamShared/Json/DreamObjectJson.cs
```diff
@@ -23,8 +23,10 @@ public class GlobalListJson
         public int GlobalCount { get; set; }
         public Dictionary<int, object> Globals { get; set; }
     }
-
+  
     public class ProcDefinitionJson {
+        public string Name { get; set; }
```
wixoaGit, line 28: This greatly bloats the output json file

OpenDreamRuntime/Objects/DreamObjectTree.cs
```diff
@@ -232,24 +232,29 @@ public class TreeEntry {
...
-                    if (procDefinition.Arguments != null) {
-                        foreach (ProcArgumentJson argument in procDefinition.Arguments) {
-                            argumentNames.Add(argument.Name);
-                            argumentTypes.Add(argument.Type);
-                        }
-                    }
-
-                    var proc = new DMProc($"{objectDefinition.Type}/{jsonProc.Key}", null, argumentNames, argumentTypes, bytecode, procDefinition.MaxStackSize, procDefinition.WaitFor ?? true);
-                    objectDefinition.SetProcDefinition(jsonProc.Key, proc);
+                    var manager = IoCManager.Resolve<IDreamManager>();
```
wixoaGit, line 256: This isn't used
```suggestion
```

OpenDreamShared/Json/DreamCompiledJson.cs
```diff
@@ -3,6 +3,8 @@
 namespace OpenDreamShared.Json {
     public class DreamCompiledJson {
         public List<string> Strings { get; set; }
+        public List<ProcDefinitionJson> GlobalProcs { get; set; }
+        public List<KeyValuePair<string,int>> InternalNameToGlobalProcId { get; set; }
```
wixoaGit, line 7: Why a list of KeyValuePairs and not a Dictionary?

wixoaGit (changes requested):

wixoaGit (approved):

Issue #145, opened 2021-06-15
`:` doesn't support var access


ike709:
Example code:
```
/obj/item/weapon/implantcase/attackby(obj/item/weapon/I, mob/user)

	if (istype(I, /obj/item/weapon/pen))
		var/t = input(user, "What would you like the label to be?", text("[]", src.name), null)  as text
		if (user.equipped() != I)
			return
		if ((get_dist(src, usr) > 1 && src.loc != user))
			return
		t = html_encode(t)
		if (t)
			src.name = text("Glass Case- '[]'", t)
		else
			src.name = "Glass Case"
	else
		if (!( istype(I, /obj/item/weapon/implanter) ))
			return
	if (I:imp)
		if ((src.imp || I:imp.implanted))
			return
		I:imp.loc = src
		src.imp = I:imp
		I:imp = null
		src.update()
		I:update()
	else
		if (src.imp)
			if (I:imp)
				return
			src.imp.loc = I
			I:imp = src.imp
			src.imp = null
			update()
			I:update()
	return
```
Exception:
```
Unhandled exception. System.Exception: Cannot dereference property "implanted" because a type specifier is missing
   at DMCompiler.DM.Expressions.Dereference..ctor(DMExpression expr, DMASTDereference astNode, Boolean includingLast) in C:\Users\main\Documents\GitHub\OpenDream\OpenDream\DMCompiler\DM\Expressions\Dereference.cs:line 29
   at DMCompiler.DM.Visitors.DMVisitorExpression.VisitDereference(DMASTDereference dereference) in C:\Users\main\Documents\GitHub\OpenDream\OpenDream\DMCompiler\DM\Visitors\DMVisitorExpression.cs:line 335
   at OpenDreamShared.Compiler.DM.DMASTDereference.Visit(DMASTVisitor visitor) in C:\Users\main\Documents\GitHub\OpenDream\OpenDream\OpenDreamShared\Compiler\DM\DMAST.cs:line 1450
   at DMCompiler.DM.DMExpression.Create(DMObject dmObject, DMProc proc, DMASTExpression expression, Nullable`1 inferredPath) in C:\Users\main\Documents\GitHub\OpenDream\OpenDream\DMCompiler\DM\DMExpression.cs:line 30
[...]

PR #440, opened 2021-10-29
change direct to lax deref for certain expressions

mumencoder:
test case:
```

#include "map.dmm"
#include "interface.dmf"

/obj
    inner
        var/ele2 = 4

    var/ele = 2
    var/innerobj

    var/obj/inner/innerobj_ty

    proc/inner_proc()
        return new /obj
    proc/elefn()
        return 3


/proc/new_obj()
    return new /obj

/world/New()
    ASSERT(new_obj().ele == 2)
    ASSERT(new_obj()?.ele == 2)
    ASSERT(new_obj().inner_proc().ele == 2)
    ASSERT(new_obj()?.inner_proc()?.ele == 2)

    var/obj/o2 = new /obj
    ASSERT(o2.inner_proc().elefn() == 3)
    ASSERT(o2?.inner_proc()?.elefn() == 3)

    o2.innerobj_ty = new /obj/inner
    ASSERT(o2:innerobj_ty.ele2 == 4)

    var/list/l = newlist(/obj)
    ASSERT(l[1].ele == 2)
    ASSERT(l[1].elefn() == 3)

    l[1].innerobj = new /obj/inner
    ASSERT(l[1].innerobj.ele2 == 4)

    var/o
    // This is still not allowed
    //world.log << "[o.ele]"
    //world.log << "[o.elefn()]"

    shutdown()
```

### DIFF
--- a/DMCompiler/DM/Expressions/Dereference.cs
+++ b/DMCompiler/DM/Expressions/Dereference.cs
@@ -15,6 +15,24 @@ namespace DMCompiler.DM.Expressions {
         public override DreamPath? Path => _path;
         DreamPath? _path;
 
+        public static bool DirectConvertable(DMExpression expr, DMASTDereference astNode) {
+            if (expr.Path == null && astNode.Expression is DMASTProcCall) {
+                return true;
+            }
+            else if (astNode.Expression is DMASTListIndex) {
+                return true;
+            }
+            else if (astNode.Expression is DMASTDereference deref) {
+                if (expr is Dereference _deref) {
+                    return DirectConvertable(_deref._expr, deref);
+                }
+            }
+            else if (astNode.Expression is DMASTDereferenceProc) {
+                return true;
+            }
+            return false;
+        }
+
         public Dereference(DMExpression expr, DMASTDereference astNode)
             : base(null) // This gets filled in later
         {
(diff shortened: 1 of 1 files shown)


mumencoder:
fixes #145

DMCompiler/DM/Expressions/Dereference.cs
```diff
@@ -15,6 +15,27 @@ class Dereference : LValue {
...
+                    return true;
+                }
+                if (expr is Dereference _deref) {
+                    return DirectConvertable(_deref._expr, deref);
+                }
+            }
+            else if (astNode.Expression is DMASTDereferenceProc) {
+                return true;
+            }
+            return false;
```
wixoaGit, line 36: A switch would be more readable, imo
```suggestion
            switch (astNode.Expression) {
                case DMASTDereference deref when deref.Type == DMASTDereference.DereferenceType.Search:
                case DMASTProcCall when expr.Path == null:
                case DMASTDereferenceProc:
                case DMASTListIndex:
                    return true;
                case DMASTDereference deref when expr is Dereference _deref:
                    return DirectConvertable(_deref._expr, deref);
                default: return false;
            }
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #454, opened 2021-11-04
[RT] Updates the readme for when we finally make RT the main branch

ike709:


README.md
```diff
@@ -4,14 +4,18 @@
 
 This project is currently in early stages, and is not feature-complete.
 
-The compiler and server should work fine for Linux-based machines, but we currently rely on WPF and WebView2 for the client, so it's Windows-only.
+The compiler and server should work fine for Linux-based machines. The client should also work fine on Linux but none of the developers use it. More details and potentially useful troubleshooting for running on Linux can be found in the haphazardly written FAQ [here](https://github.com/wixoaGit/OpenDream/blob/RobustToolbox/LINUX_FAQ.md).
```
ZeWaka, line 7: lolololol

ZeWaka (approved):

README.md
wixoaGit, line 7: The LINUX_FAQ.md only contains outdated information about CEF and setup. I'm removing it.

This line in the readme was meant to point out that one piece of the software was incompatible with Linux, which is no longer the case. Removing it entirely seems best.

wixoaGit (changes requested):

README.md
ike709, line 7:
```suggestion
All parts of OpenDream should work fine on Windows and Linux, though the latter is not used as much and therefore isn't as thoroughly tested.
```

ike709:
Tweaked the linux mention to just say we don't test it as much, which is accurate.

wixoaGit (approved):

PR #460, opened 2021-11-10
feat: NativeProc - nonspantext

### LINKED ISSUE #357: `nonspantext()` not implemented
http://www.byond.com/docs/ref/#/proc/nonspantext

Also see https://github.com/wixoaGit/OpenDream/issues/304

SantaGitHub:
Closes #357


PJB3005 (changes requested):
This doesn't seem to implement the following behavior from BYOND correctly:
1. If you pass a multi-char string it should return the first index of *any* of the characters in the string, not what `IndexOf()` does which is the first occurence of the entire string. You can use `IndexOfAny` for this if you need (use the span variant). https://docs.microsoft.com/en-us/dotnet/api/system.memoryextensions.indexofany?view=net-5.0
2. This doesn't handle the reverse seeking behavior if `Start` is negative (it will throw).

SantaGitHub:
@PJB3005 i fix it, review pls

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -866,6 +866,35 @@ public static DreamValue NativeProc_list2params(DreamObject instance, DreamObjec
...
+            {
+                index = text.IndexOfAny(needles.ToCharArray(), (int)start);
+            }
+
+            if (index > 0)
+            {
+                index++; // Increase, because in DM the enumaration starts from 1
+            }
+
+            return new DreamValue(index);
```
PJB3005, line 895: The BYOND proc counts the amount of characters passed before hitting either the end of the string or one of the needle characters. You're returning an absolute index right now.

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -866,6 +866,35 @@ public static DreamValue NativeProc_list2params(DreamObject instance, DreamObjec
...
+        {
+            string text = arguments.GetArgument(0, "Haystack").GetValueAsString();
+            string needles = arguments.GetArgument(1, "Needles").GetValueAsString();
+            float start = arguments.GetArgument(2, "Start").GetValueAsFloat() - 1; // Decrease, because in DM the enumeration starts from 1
+
+            int index = 0;
+            if (start < 0)
+            {
+                start = text.Length + start;
+                index = text.IndexOfAny(needles.ToCharArray(), (int)start);
```
PJB3005, line 883:
```suggestion
                index = text.AsSpan().IndexOfAny(needles, (int)start);
```
You can do this to avoid turning the needles string into an array, which is more optimal (also do this below).

PJB3005: Also, you could move this `index = text.IndexOfAny(needles.ToCharArray(), (int)start);` out of the if/else because it is identical between the two blocks. Just move it after the rest so the only thing the `if` will contain is `start = text.Length + start;`

PJB3005 (commented):
I would encourage you to try to play around with the actual proc in DM to make sure you get the behavior correct. Ideally we would have a unit testing harness or something for this but alas.

You can use the MoMMI bot on Discord for testing if you want.

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
SantaGitHub: AsSpan().IndexOfAny() not support "start index"
image

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
SantaGitHub: fix

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
SantaGitHub: fix

SantaGitHub:
I looked at the discord and did not understand how this proc works in DM

wixoaGit:
This proc gives you the amount of characters between the start index and the needle (or end of the string if it doesn't find the needle)

```
nonspantext("abcdef", "d")
```
This returns `3` because there are 3 characters between the first index and `d`

```
nonspantext("abcdef", "a", 3)
```
This returns `4` because there is no `a` after/on character 3, and so it instead counts the amount of characters between index 3 and the end of the string

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
PJB3005: Use `AsSpan((int) start).IndexOfAny(needles)` instead then. That *should* work. But note that your returned index will be relative to the start position.

ike709:
http://www.byond.com/docs/ref/#/proc/nonspantext

OpenDreamRuntime.Tests/UnitTest1.cs
```diff
@@ -444,5 +445,30 @@ public void UnicodeProcsTest()
 
             Assert.Zero(runtime.ExceptionCount);
         }
+
+        [TestCase("Hello, World!", ", ", -1, 6)]
+        [TestCase("Hello, World!", ", ", 3, 3)]
+        [TestCase("Hello, World!", ", ", 7, 11)]
```
wixoaGit, line 451: Adjust for the previous fixes and add an out-of-bounds test case
```suggestion
        [TestCase("Hello, World!", ", ", -1, 1)]
        [TestCase("Hello, World!", ", ", 3, 3)]
        [TestCase("Hello, World!", ", ", 7, 0)]
        [TestCase("Hello, World!", ", ", 14, 0)]
        [TestCase("Hello, World!", ", ", 0, 0)]
```

OpenDreamRuntime.Tests/UnitTest1.cs
```diff
@@ -444,5 +445,30 @@ public void UnicodeProcsTest()
...
+            var valueResultDreamValue = new DreamValue(valueResult);
+
+            var listDreamValue = new List<DreamValue>() { haystackDreamValue, needlesDreamValue, startDreamValue };
+            var dicDreamValue = new Dictionary<string, DreamValue>() { ["Haystack"] = haystackDreamValue, ["Needles"] = needlesDreamValue, ["Start"] = startDreamValue };
+
+            var result = DreamThread.Run(runtime, async state =>
+            {
+                var world = runtime.WorldInstance;
+                var proc = world.GetProc("nonspantext");
+                return await state.Call(proc, world, null, new DreamProcArguments(listDreamValue, dicDreamValue));
```
wixoaGit, line 468: `dicDreamValue` is unnecessary because the arguments are already being passed in with `listDreamValue`
```suggestion

            var result = DreamThread.Run(runtime, async state =>
            {
                var world = runtime.WorldInstance;
                var proc = world.GetProc("nonspantext");
                return await state.Call(proc, world, null, new DreamProcArguments(listDreamValue));
```

wixoaGit (changes requested):

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -866,6 +866,34 @@ public static DreamValue NativeProc_list2params(DreamObject instance, DreamObjec
...
         }
 
+        [DreamProc("nonspantext")]
+        [DreamProcParameter("Haystack", Type = DreamValueType.String)]
+        [DreamProcParameter("Needles", Type = DreamValueType.String)]
+        [DreamProcParameter("Start", Type = DreamValueType.Float, DefaultValue = 1)]
+        public static DreamValue NativeProc_nonspantext(DreamObject instance, DreamObject usr, DreamProcArguments arguments)
+        {
+            string text = arguments.GetArgument(0, "Haystack").GetValueAsString();
+            char[] needles = arguments.GetArgument(1, "Needles").GetValueAsString().ToCharArray();
```
wixoaGit, line 876:
```suggestion
            string needles = arguments.GetArgument(1, "Needles").GetValueAsString();
```

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -866,6 +866,34 @@ public static DreamValue NativeProc_list2params(DreamObject instance, DreamObjec
...
+            int start = (int)arguments.GetArgument(2, "Start").GetValueAsFloat();
+
+            if (start < 0)
+            {
+                start = text.Length + start;
+            }
+            var index = text.IndexOfAny(needles, start);
```
wixoaGit, line 883:
```suggestion
            var index = text.AsSpan(start - 1).IndexOfAny(needles);
```

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -866,6 +866,34 @@ public static DreamValue NativeProc_list2params(DreamObject instance, DreamObjec
...
+            if (index == -1)
+            {
+                index = text.IndexOfAny(needles);
+                index = text.Length + index - start;
+            }
+            else
+            {
+                index = index - start + 1;
+            }
```
wixoaGit, line 892:
```suggestion
```

wixoaGit (changes requested):
The following three changes are to apply PJB's optimization

wixoaGit (approved):

PR #468, opened 2021-11-18
Eat trailing semicolons after for() and while()

ike709:
This is valid DM: `for(A, A && !ismob(A), A=A.loc);`

The semicolon seems to have no effect.

### DIFF
--- a/OpenDreamShared/Compiler/DM/DMParser.cs
+++ b/OpenDreamShared/Compiler/DM/DMParser.cs
@@ -789,6 +789,7 @@ namespace OpenDreamShared.Compiler.DM {
                     }
 
                     ConsumeRightParenthesis();
+                    Check(TokenType.DM_Semicolon);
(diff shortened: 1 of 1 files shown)


OpenDreamShared/Compiler/DM/DMParser.cs
```diff
@@ -919,6 +923,7 @@ public DMASTProcStatement ProcStatement()
                 DMASTExpression conditional = Expression();
                 if (conditional == null) Error("Expected conditional");
                 ConsumeRightParenthesis();
+                Check(TokenType.DM_Semicolon);
```
wixoaGit, line 926: This doesn't seem necessary for a do/while

ike709:
```suggestion
```

wixoaGit (approved):

Issue #471, opened 2021-11-18
Enumerating `dir` causes an exception


ike709:
`for(dir in list(NORTH,EAST,SOUTH,WEST))`

wixoaGit:
Fixed by 5d611b717c3cf6eeee7eb737a7ae53c0e935fe2a.

Issue #470, opened 2021-11-18
Using `newlist` as an identifier causes issues


ike709:
```
/mob/living/carbon/proc/custom_dreams(list/dreamlist, mob/user)
    var/list/newlist = dreamlist.Copy()
    for(var/i in 1 to newlist.len)
        newlist[i] = replacetext(newlist[i], "DREAMER", "[user.name]")
    return newlist
```
`newlist` also happens to be a keyword so this causes a NRE

ike709:
See https://github.com/wixoaGit/OpenDream/commit/566ad3567ce924632095b3259247035d9ad17da6

PR #472, opened 2021-11-18
More robust exception handling

ike709:
Paradise will now compile without throwing a real exception, minus #470 and #471 which need to be fixed anyways.

### DIFF
--- a/DMCompiler/DM/DMObject.cs
+++ b/DMCompiler/DM/DMObject.cs
@@ -47,7 +47,7 @@ namespace DMCompiler.DM {
 
         public DMVariable GetVariable(string name) {
             if (Variables.TryGetValue(name, out DMVariable variable)) {
-                if ((variable.Value.ValType & DMValueType.Unimplemented) == DMValueType.Unimplemented && Program.WarnForUnimplemented)
+                if (variable.Value is not null && (variable.Value.ValType & DMValueType.Unimplemented) == DMValueType.Unimplemented && Program.WarnForUnimplemented)
                 {
                     Program.Warning(new CompilerWarning(null, $"{Path}.{name} is not implemented and will have unexpected behavior"));
                 }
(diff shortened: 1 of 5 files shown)


DMCompiler/DM/DMObject.cs
```diff
@@ -47,7 +47,7 @@ class DMObject {
 
         public DMVariable GetVariable(string name) {
             if (Variables.TryGetValue(name, out DMVariable variable)) {
-                if ((variable.Value.ValType & DMValueType.Unimplemented) == DMValueType.Unimplemented && Program.WarnForUnimplemented)
+                if (variable.Value is not null && (variable.Value.ValType & DMValueType.Unimplemented) == DMValueType.Unimplemented && Program.WarnForUnimplemented)
```
wixoaGit, line 50:
```suggestion
                if ((variable.Value?.ValType & DMValueType.Unimplemented) == DMValueType.Unimplemented && Program.WarnForUnimplemented)
```

DMCompiler/DM/Visitors/DMVisitorExpression.cs
```diff
@@ -84,6 +84,11 @@ class DMVisitorExpression : DMASTVisitor {
             } else if (name == "args") {
                 Result = new Expressions.Args();
             } else {
+                if (_proc is null)
```
wixoaGit, line 87: What code causes this to happen?

wixoaGit (changes requested):

DMCompiler/DM/Visitors/DMVisitorExpression.cs
ike709: Setting an unknown identifier to an unknown identifier in an object definition.

wixoaGit (approved):

PR #486, opened 2021-11-27
Adds `--unique-errors`

ike709:
Adds an arg of dubious use, `--unique-errors`.

It's an inaccurate way of getting a rough idea of what individual things are broken.

I'm not sure if it's even worth merging or not, but it does make it slightly easier to look through the error list for things to work on.


Farseer2:
Wouldn't `HashSet` fit better?

ike709:
Not useful enough to merge

PR #490, opened 2021-11-29
Adds the cefglue fix to the readme

ike709:


ZeWaka:
i mean this should only occur if you previously had the submodule set to non-robust cefglue right?

PJB3005:
This feels like a pretty minor thing since it only happens to the subset of people who checked out the robust repo before we changed the submodule address.

ike709:
fair

PR #493, opened 2021-11-30
Fix sound volume.

Zumorica:
OpenAL volume is in decibels, BYOND's volume goes from 0 to 100.
I fixed this issue by doing a funny logarithm volume scale.
In C#, `MathF.Log10(0)` is negative infinity, therefore resulting in completely muted audio.

DM Snippet for easy testing (works in testgame!)
```DM
	verb/input_num()
		var/v = input("A") as num
		usr << sound("test.ogg", volume=v)
		usr << "you entered [v]"
```

### DIFF
--- a/OpenDreamClient/Audio/DreamSoundEngine.cs
+++ b/OpenDreamClient/Audio/DreamSoundEngine.cs
@@ -1,3 +1,4 @@
+using System;
 using OpenDreamClient.Resources;
 using OpenDreamClient.Resources.ResourceTypes;
 using OpenDreamShared.Network.Messages;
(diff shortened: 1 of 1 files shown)


OpenDreamClient/Audio/DreamSoundEngine.cs
```diff
@@ -44,7 +45,7 @@ public void PlaySound(int channel, ResourceSound sound, float volume)
 
             StopChannel(channel);
 
-            var source = sound.Play(AudioParams.Default.WithVolume(volume));
+            var source = sound.Play(AudioParams.Default.WithVolume(20 * MathF.Log10(volume)));
```
ZeWaka, line 48:
```suggestion
            var source = sound.Play(AudioParams.Default.WithVolume(20 * MathF.Log10(volume))); // convert from DM volume (0-100) to OpenAL volume (db)
```

wixoaGit (approved):

PR #494, opened 2021-11-30
DreamList.Create() no longer calls InitSpawn()

ike709:
Partially addresses the first point in https://github.com/wixoaGit/OpenDream/issues/342


wixoaGit:
You can do the same for the `Create(IEnumerable<object> collection)` below it.

wixoaGit (approved):

PR #506, opened 2021-12-04
Adds support for for()

### LINKED ISSUE #463: for() support
DM supports for() with no arguments as a method to create an infinite loop. It's slightly faster than while(1), for some reason.

```
for()
    world << "Tacos!"
    sleep(5)
````

### PR DESCRIPTION
**Pull request by EdgeLordExe**
For() now compiles down to an infinite loop as specified by byond. 

Also constant while loops will also be compiled down to this

closes #463

### CODE REVIEW

File: DMCompiler/DM/Visitors/DMProcBuilder.cs [hunk 1]
```diff
@@ -318,6 +319,22 @@ class DMProcBuilder {
             _proc.DestroyEnumerator();
         }
 
+        //Generic infinite loop, while loops with static expression as their conditional with positive truthfullness get turned into this as well as empty for() calls
```
ZeWaka, line 322: I don't see where you did this for `while` loops.

EdgeLordExe: aa i forgot to do this for while loops gimme a moment

DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -857,6 +857,12 @@ public DMASTProcStatement ProcStatement()
 
                             initializer = new DMASTProcStatementExpression(variable.Location, new DMASTAssign(variable.Location, variable, value));
                         }
+                    } else {
```
ike709, line 860:
```suggestion
                    } else if(Current().Type != TokenType.DM_Comma && Current().Type != TokenType.DM_Semicolon) {
```
Needed otherwise loops without initializers like `for (,user.bhunger>0,user.bhunger--)` will break

ike709: Not ideal but for loops need a rewrite anyways.

wixoaGit (approved):

PR #509, opened 2021-12-04
Fixes LastElement throwing an exception

ike709:
While compiling NTstation a path string of `"/"` is causing an exception because Elements is 0.

Not sure if this is the proper fix but it does stop the exception.


wixoaGit:
This is likely to silently cause issues in code that depends on paths having a LastElement. It should be the responsibility of the code using LastElement to handle empty paths.

wixoaGit (approved):

PR #510, opened 2021-12-05
Adds location support to CompilerWarnings

ike709:
Not quite as useful until we have location information stored in expressions, but it does point out unsupported `in world` loops.

### DIFF
--- a/OpenDreamShared/Compiler/CompilerError.cs
+++ b/OpenDreamShared/Compiler/CompilerError.cs
@@ -2,18 +2,15 @@
 
 namespace OpenDreamShared.Compiler {
     public struct CompilerError {
-        public Token Token;
         public Location? Location;
         public string Message;
 
         public CompilerError(Token token, string message) {
-            Token = token;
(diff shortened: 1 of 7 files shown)


OpenDreamShared/Compiler/CompilerError.cs
```diff
@@ -24,24 +21,21 @@ public struct CompilerError {
     }
 
     public struct CompilerWarning {
-        public Token Token;
+        public Location? Location;
```
wixoaGit, line 24:
```suggestion
        public Location Location;
```

wixoaGit (changes requested):

OpenDreamShared/Compiler/CompilerError.cs
```diff
@@ -2,18 +2,15 @@
 
 namespace OpenDreamShared.Compiler {
     public struct CompilerError {
-        public Token Token;
         public Location? Location;
```
ike709, line 5:
```suggestion
        public Location Location;
```

wixoaGit (approved):

PR #511, opened 2021-12-05
Tweaks switch() errors

ike709:
1. `Expected dedent` in the switch() parsing has been reworded to match BYOND's error
2. Using `else if()` as a switch case throws that error too since it produces unexpected behavior in both BYOND and OD


DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -1163,6 +1163,10 @@ public DMASTProcStatement ProcStatement()
             } else if (Check(TokenType.DM_Else)) {
                 var loc = Current().Location;
                 Whitespace();
+                if (Check(TokenType.DM_If))
+                {
+                    DMCompiler.Error(new CompilerError(loc, "Expected \"if\" or \"else\""));
```
wixoaGit, line 1168: Use `Error()` to trigger the parser's error recovery
```suggestion
                    Error("Expected \"if\" or \"else\"");
```

wixoaGit (changes requested):

PR #516, opened 2021-12-09
Stub `_dm_db` procs

ike709:


ZeWaka:
We're going to have an internal SQLite instance for these right? Spinup on first use?

ike709:
/shrug

Sorting out the implementation is a problem for the future.

wixoaGit (approved):

PR #520, opened 2021-12-10
Parse global/static/const/tmp in proc bodies

ike709:


wixoaGit:
Could you add an unimplemented warning too?

wixoaGit (approved):

PR #531, opened 2021-12-10
Stub `world/var/map_format`

ike709:


DMCompiler/DMStandard/Defines.dm
```diff
@@ -178,4 +178,10 @@
...
 #define VIS_UNDERLAY 64
-#define VIS_HIDE 128
\ No newline at end of file
+#define VIS_HIDE 128
+
+//world.map_format
+#define TOPDOWN_MAP 1
+#define ISOMETRIC_MAP 2
+#define SIDE_MAP 3
+#define TILED_ICON_MAP 4
```
wixoaGit, line 187: These are not the correct values.
```suggestion
#define TOPDOWN_MAP 0
#define ISOMETRIC_MAP 1
#define SIDE_MAP 2
#define TILED_ICON_MAP 32768
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #532, opened 2021-12-10
Fix exception with negative sleeps

### LINKED ISSUE #489: Negative sleeps cause an exception
Same as https://github.com/wixoaGit/OpenDream/commit/0672db926d2bebe68f180c2b6f8e538201d0622e just need to apply the fix to `sleep()`

mirrorcult:
Code used to test:

```dm
/world/New()
	..()
	world.log << "Running"
	sleep(-1)
	world.log << "No exception"
```

Fixes #489. Does not reproduce byond behavior exactly, just stops the exception


ZeWaka (changes requested):
negative sleeps have special behavior, this is incorrect - see the linked issue

wixoaGit (approved):

PR #533, opened 2021-12-11
Implements default `atom/var/text` behavior

### LINKED ISSUE #523: `atom/var/text` isn't implemented
See http://www.byond.com/docs/ref/#/atom/var/text

ike709:
Fixes #523 

Implements the default behavior for `atom/var/text`, which is setting it to the first letter of name.

According to the docs, `text` is actually quite a complex feature. But this gets the default value working.


ZeWaka:
Yeah we abuse `text` on goon for our custom ascii mode, we do stuff like coloring the text and other funsies.

wixoaGit:
Would it be better to do this in the runtime during object tree initialization?

ike709:
You tell me

wixoaGit:
I'd prefer for this to be done in the runtime. This is engine behavior instead of language behavior.

PR #539, opened 2021-12-17
Two compiler optimizations

ike709:
1. PathString's getter was relatively quite expensive. StringBuilder is actually a worse option when you're only concatenating a few strings. Replaced it with a regular `string.Join()` and shaved off a small amount of both time and memory usage.
2. Reorganized the dozen or so procStatement method calls to be primarily sorted by frequency, then moved `If()` last because although it's by far the most common it's also by far the most expensive of them.

Altogether this shaves about 5 seconds off my compile time, making us officially 4x faster at compiling than byond.

### DIFF
--- a/DMCompiler/Compiler/DM/DMParser.cs
+++ b/DMCompiler/Compiler/DM/DMParser.cs
@@ -531,6 +531,7 @@ namespace DMCompiler.Compiler.DM {
 
                 return new DMASTProcStatementExpression(loc, expression);
             } else {
+                // These are sorted by frequency, except If() is moved to the end because it's really slow (relatively)
                 DMASTProcStatement procStatement = Return();
                 if (procStatement == null) procStatement = ProcVarDeclaration();
                 if (procStatement == null) procStatement = For();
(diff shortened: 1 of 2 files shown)


DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -531,21 +531,21 @@ public DMASTProcStatement ProcStatement()
 
                 return new DMASTProcStatementExpression(loc, expression);
             } else {
-                DMASTProcStatement procStatement = ProcVarDeclaration();
-                if (procStatement == null) procStatement = Return();
-                if (procStatement == null) procStatement = Break();
-                if (procStatement == null) procStatement = Continue();
-                if (procStatement == null) procStatement = Goto();
-                if (procStatement == null) procStatement = Del();
+                DMASTProcStatement procStatement = Return();
```
ike709, line 534:
```suggestion
                // These are sorted by frequency, except If() is moved to the end because it's really slow (relatively)
                DMASTProcStatement procStatement = Return();
```

wixoaGit (approved):

PR #544, opened 2021-12-20
Turns object _stack[] into DreamValue _stack[]

EdgeLordExe:
Makes the stack slightly less boxed:tm: , i still have to like, actually unbox IDreamProcIdentifier and DreamProcArguments ( latter is much easier), for now i just added a new DreamValueType, that stores them directly. once i figure out what to do, im gonna fix it. Either way the game should be slightly faster.

haven't tested on SS13, test game seems to run fine, i didnt notice any runtime errors from the small amount of testing i did.

### DIFF
--- a/OpenDreamRuntime/DreamValue.cs
+++ b/OpenDreamRuntime/DreamValue.cs
@@ -4,6 +4,7 @@ using System.Text.Json.Serialization;
 using OpenDreamRuntime.Objects;
 using OpenDreamRuntime.Resources;
 using OpenDreamShared.Dream;
+using OpenDreamRuntime.Procs;
 
 namespace OpenDreamRuntime {
     [JsonConverter(typeof(DreamValueJsonConverter))]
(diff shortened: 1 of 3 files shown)


OpenDreamRuntime/DreamValue.cs
```diff
@@ -74,7 +75,7 @@ public enum DreamValueType {
                 DreamObject => DreamValueType.DreamObject,
                 DreamPath => DreamValueType.DreamPath,
                 DreamProc => DreamValueType.DreamProc,
-                _ => throw new ArgumentException("Invalid DreamValue value (" + value + ", " + value.GetType() + ")")
+                _ => DreamValueType.Reference
```
wixoaGit, line 78: Could this be changed to accept only `IDreamProcIdentifier` and `DreamProcArguments` to keep the error handling for invalid values?

wixoaGit (changes requested):

OpenDreamRuntime/DreamValue.cs
EdgeLordExe: that's a good idea

wixoaGit (approved):

PR #548, opened 2021-12-23
Enhances the world.log implementation

ike709:
1. Outputting `world.log` to the console is now properly formatted. Also paves the way for implement loglevels in DM at some point if we wanted to.
2. Implements setting `world.log` to a string/file and outputting to that file instead.

I'm of the opinion that there should be an easy, toggleable way to have it continue printing to the console even if you set it to a file, so I'll probably add a config in a followup PR for that.


OpenDreamRuntime/Resources/LogOutputResource.cs
```diff
@@ -0,0 +1,49 @@
...
+                base.Output(value);
+            }
+        }
+
+        public void WriteLog(LogLevel level, string message)
+        {
+            if (ResourcePath != null)
+            {
+                base.Output(new DreamValue(message));
+                Logger.LogS(level, "world.log", message);
```
ike709, line 42:
```suggestion
```

OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectWorld.cs
```diff
@@ -48,6 +49,9 @@ class DreamMetaObjectWorld : DreamMetaObjectRoot {
                     _cfg.SetCVar(CVars.NetTickrate, variableValue.GetValueAsInteger()); break;
                 case "maxz":
                     _dreamMapManager.SetZLevels(variableValue.GetValueAsInteger()); break;
+                case "log":
+                    _dreamManager.WorldLog = new LogOutputResource(GetLogPath(variableValue));
+                    break;
```
wixoaGit, line 54: Running this code in byond shows that setting `world.log` to a string causes it to actually be set to a resource
```
/mob/verb/test()
    world.log = "output.txt"
    usr << istext(world.log)
    usr << isfile(world.log)
```
Output:
```
0
1
```
Instead of having `_dreamManager.WorldLog` always be a special `LogOutputResource` it'd likely be better as a more generic `DreamResource` (keeping something akin to the `ConsoleOutputResource` that currently exists). You could return this in the getter for `world.log` which would remove the need for the special handling in the `BitShiftLeft` opcode.

Doing `world.log = null` also appears to set it back to the console.

wixoaGit (changes requested):

ike709:
Including the loglevel is a breaking change but IMO it's an improvement and also it's useful for development.

OpenDreamClient/GlobalUsings.cs
```diff
@@ -0,0 +1,14 @@
+// Global usings for OpenDreamClient
```
ZeWaka, line 1: muh atomize

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -538,6 +538,8 @@ static class DMOpcodeHandlers {
                 case DreamValue.DreamValueType.Float when second.Type == DreamValue.DreamValueType.Float:
                     state.Push(new DreamValue(first.GetValueAsInteger() << second.GetValueAsInteger()));
                     break;
+                case DreamValue.DreamValueType.String when second.Type == DreamValue.DreamValueType.String:
```
ZeWaka, line 541: waaa

ZeWaka (approved):

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -538,6 +538,8 @@ static class DMOpcodeHandlers {
...
+                    throw new Exception("Invalid bit shift left operation on " + first + " and " + second);
```
ike709, line 542:
```suggestion
```

OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectWorld.cs
```diff
@@ -48,6 +48,20 @@ class DreamMetaObjectWorld : DreamMetaObjectRoot {
...
                     _dreamMapManager.SetZLevels(variableValue.GetValueAsInteger()); break;
+                case "log":
+                    if (variableValue.TryGetValueAsString(out var logStr))
+                    {
+                        dreamObject.SetVariableValue("log", new DreamValue(_dreamRscMan.LoadResource(logStr)));
+                    }
+                    else if (variableValue.TryGetValueAsDreamResource(out var logRsc))
+                    {
+                        dreamObject.SetVariableValue("log", new DreamValue(logRsc));
+                    }
```
wixoaGit, line 59: The variable's value is already set to `variableValue` so this has no effect.
```suggestion
```

OpenDreamRuntime/DreamManager.cs
```diff
@@ -137,5 +137,22 @@ public void Update()
 
             GlobalProcs[name] = proc;
         }
+
+        public void WriteWorldLog(string message, LogLevel level = LogLevel.Info)
+        {
+            if (!WorldInstance.GetVariable("log").TryGetValueAsDreamResource(out var logRsc))
+            {
+                WorldInstance.SetVariableValue("log", new DreamValue(new ConsoleOutputResource()));
```
wixoaGit, line 145: Shouldn't `world.log` always be a resource if `WorldInstance` is loaded?

wixoaGit (changes requested):

ike709:
review addressed

wixoaGit (approved):

PR #555, opened 2021-12-24
Fixes a few exceptions

### LINKED ISSUE #557: Setting the value of `src` causes an exception
`Assign()` can't cope with it, causing an exception in `PeekIdentifier()`

ike709:
Does a few things:
1. Makes sure a turf's loc is set before init to fix issues in New()
2. Adds the ability to modify an actual variable value. Fixes (partially) #556
3. Made Atom's OnVariableSet() a switch
4. `src` can be set to a DreamObject. Fixes #557

### DIFF
--- a/OpenDreamRuntime/DreamMapManager.cs
+++ b/OpenDreamRuntime/DreamMapManager.cs
@@ -139,21 +139,25 @@ namespace OpenDreamRuntime {
                 DreamPath areaType = cellDefinition.Area != null ? _dreamManager.ObjectTree.Types[cellDefinition.Area.Type].Path : _defaultArea;
                 DreamObject area = GetArea(areaType);
 
-                DreamObject turf;
-                if (cellDefinition.Turf != null) {
-                    turf = CreateMapObject(cellDefinition.Turf);
-                } else {
-                    turf = _dreamManager.ObjectTree.CreateObject(_defaultTurf);
-                    turf.InitSpawn(new DreamProcArguments(null));
-                }
-
                 int x = block.X + blockX - 1;
                 int y = block.Y + block.Height - blockY;
 
-                SetTurf(x, y, block.Z, turf);
-                SetArea(x, y, block.Z, area);
+                DreamObject turf;
+                if (cellDefinition.Turf != null) {
+                    turf = CreateMapObject(cellDefinition.Turf);
+                    SetTurf(x, y, block.Z, turf);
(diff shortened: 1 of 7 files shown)


OpenDreamRuntime/DreamMapManager.cs
```diff
@@ -139,21 +139,25 @@ public struct Cell {
...
                     turf = CreateMapObject(cellDefinition.Turf);
+                    SetTurf(x, y, block.Z, turf);
+                    SetArea(x, y, block.Z, area);
+                    turf.InitSpawn(new DreamProcArguments(new() {DreamValue.Null}));
                 } else {
                     turf = _dreamManager.ObjectTree.CreateObject(_defaultTurf);
+                    SetTurf(x, y, block.Z, turf);
+                    SetArea(x, y, block.Z, area);
                     turf.InitSpawn(new DreamProcArguments(null));
                 }
```
wixoaGit, line 156: This repeated code could be moved outside the if/else
```suggestion
                } else {
                    turf = _dreamManager.ObjectTree.CreateObject(_defaultTurf);
                }
                
                SetTurf(x, y, block.Z, turf);
                SetArea(x, y, block.Z, area);
                turf.InitSpawn(new DreamProcArguments(null));
                    
```

OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectAtom.cs
```diff
@@ -36,94 +35,119 @@ class DreamMetaObjectAtom : DreamMetaObjectDatum {
...
+                        appearance.PixelOffset.Y = variableValue.GetValueAsInteger();
+                    });
+                    break;
+                case "layer":
+                    UpdateAppearance(dreamObject, appearance => {
+                        appearance.Layer = variableValue.GetValueAsFloat();
+                    });
+                    break;
+                case "invisibility":
+                    var vis = variableValue.TryGetValueAsFloat(out var val) ? Convert.ToInt32(Math.Floor(val)) : 0;
```
wixoaGit, line 74: `TryGetValueAsInteger()` will floor a float, and sets the output value to 0 on failure
```suggestion
                    variableValue.TryGetValueAsInteger(out int vis);
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #561, opened 2021-12-26
Makes `spawn()` and `sleep()` actually yield

### LINKED ISSUE #560: Proc scheduler seems to be having issues
Specifically, it really seems like `spawn(0)` isn't working correctly and the code is executing way later than it should.

ike709:
This partially addresses https://github.com/wixoaGit/OpenDream/issues/447 but not to the point where I'd call it fixed.

It also fixes #560 

At the moment you need to either get lucky with your connect attempt lining up with sufficient yielding, or pass the *client* each of the cvars in `server_config.toml` like you pass the json to the server. I'm going to ask around in SS14 Discord about the possibility of adding a `client_config.toml` for use when not using the launcher.

### DIFF
--- a/OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
+++ b/OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
@@ -1271,7 +1271,7 @@ namespace OpenDreamRuntime.Procs.Native {
             float delay = state.Arguments.GetArgument(0, "Delay").GetValueAsFloat();
             int delayMilliseconds = (int)(delay * 100);
 
-            // TODO: This is may not be the proper behaviour, see https://www.byond.com/docs/ref/#/proc/sleep
+            // TODO: This may not be the proper behaviour, see https://www.byond.com/docs/ref/#/proc/sleep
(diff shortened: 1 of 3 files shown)


OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -1271,12 +1271,14 @@ public static DreamValue NativeProc_shutdown(DreamObject instance, DreamObject u
             float delay = state.Arguments.GetArgument(0, "Delay").GetValueAsFloat();
             int delayMilliseconds = (int)(delay * 100);
 
-            // TODO: This is obviously not the proper behaviour, see https://www.byond.com/docs/ref/#/proc/sleep
+            // TODO: This is may not be the proper behaviour, see https://www.byond.com/docs/ref/#/proc/sleep
```
wixoaGit, line 1274:
```suggestion
            // TODO: This may not be the proper behaviour, see https://www.byond.com/docs/ref/#/proc/sleep
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #564, opened 2021-12-28
Retain _currentPath proc elements in DMParser

mumencoder:
This was causing the AST to mark procs defined like:

```
proc 
    Fn1()
    Fn2()
```

as overrides


wixoaGit:
This check was there because without it the procs would end up on a type called `/a/b/proc` instead of `/a/b`

PR #569, opened 2021-12-29
escape sequence priority change

mumencoder:
for long escapes read the entire word before making a choice

allows \himself \herself to take priority over \him and \her

### DIFF
--- a/DMCompiler/Compiler/DM/DMLexer.cs
+++ b/DMCompiler/Compiler/DM/DMLexer.cs
@@ -7,30 +7,25 @@ using OpenDreamShared.Compiler;
 namespace DMCompiler.Compiler.DM {
     public partial class DMLexer : TokenLexer {
         public static List<string> ValidEscapeSequences = new() {
-            "t", "n",
-            "[", "]",
-            "\\", " ", "\"", "'",
-            "<", ">",
-
             "icon",
             "Roman", "roman",
             "The", "the",
-            "A", "a", "An", "an",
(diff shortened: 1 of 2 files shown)


DMCompiler/Compiler/DM/DMLexer.cs
```diff
@@ -7,30 +7,25 @@
...
-            "[", "]",
-            "\\", " ", "\"", "'",
-            "<", ">",
-
             "icon",
             "Roman", "roman",
             "The", "the",
-            "A", "a", "An", "an",
+            "A", "a",
+            "An", "an",
```
ZeWaka, line 14: shouldn't these be reversed?

DMCompiler/Compiler/DM/DMLexer.cs
```diff
@@ -7,30 +7,25 @@
...
             "th",
             "s",
             "He", "he",
```
ZeWaka, line 17: Below `He`rs?

ZeWaka (commented):
unless i'm reading your changes wrong and the order in the list doesn't matter

DMCompiler/Compiler/DM/DMLexer.cs
mumencoder: Doesn't matter, the main change is how the escapes are read in to match to the list, and that part only uses Contains

wixoaGit (approved):

PR #579, opened 2022-02-02
Make bad ..() calls a warning instead of an error

ike709:
Bad DM code has too much of this. If we ever make a clone of C# pragmas we could consider having it default as an error.

### DIFF
--- a/DMCompiler/DM/Expressions/Procs.cs
+++ b/DMCompiler/DM/Expressions/Procs.cs
@@ -81,7 +81,7 @@ namespace DMCompiler.DM.Expressions {
         public override (DMReference Reference, bool Conditional) EmitReference(DMObject dmObject, DMProc proc) {
             if (!proc.IsOverride)
             {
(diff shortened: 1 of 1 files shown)


DMCompiler/DM/Expressions/Procs.cs
```diff
@@ -81,7 +81,7 @@ class ProcSuper : DMExpression {
         public override (DMReference Reference, bool Conditional) EmitReference(DMObject dmObject, DMProc proc) {
             if (!proc.IsOverride)
             {
-                throw new CompileErrorException(Location, "Cannot call parents via ..() in a proc definition");
+                DMCompiler.Warning(new CompilerWarning(Location, "Cannot call parents via ..() in a proc definition"));
```
wixoaGit, line 84:
```suggestion
                DMCompiler.Warning(new CompilerWarning(Location, "Calling parents via ..() in a proc definition does nothing"));
```

wixoaGit (changes requested):

ZeWaka:
😭

wixoaGit (approved):

PR #580, opened 2022-02-05
Adds support for enumerating clients and datums

### LINKED ISSUE #50: Special loops like `for (var/client/C)` aren't supported
title

```
Error at code\game\admin\ban.dm:126:20: Expected 'to'
Error at code\game\admin\ban.dm:126:20: Expected an expression
```

ike709:
Didn't want to get back to logging yet so I did this instead.

Fixes #50

~~Still needs support for datums but `for(var/client/C)` is actually used a fair bit.~~


OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -39,15 +39,38 @@ static class DMOpcodeHandlers {
...
 
             if (list == null) {
                 if (listObject == null) {
                     list = null;
-                } else if (listObject.IsSubtypeOf(DreamPath.Atom) || listObject.IsSubtypeOf(DreamPath.World)) {
+                } else if (listObject.IsSubtypeOf(DreamPath.Atom))
+                {
                     list = listObject.GetVariable("contents").GetValueAsDreamList();
+                }
+                else if (listObject.IsSubtypeOf(DreamPath.World)) {
```
wixoaGit, line 55: `for(var/datum/D in world)` is not the same thing as `for(var/datum/D)`. This implementation leads to the two having the same behavior.

A better way to do this would be to have the compiler output a new `CreateTypeEnumerator` opcode that pushes an enumerator for the instances of a type.

OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectDatum.cs
```diff
@@ -1,14 +1,31 @@
...
 
 namespace OpenDreamRuntime.Objects.MetaObjects {
     class DreamMetaObjectDatum : DreamMetaObjectRoot {
         public override bool ShouldCallNew => true;
+        private bool _inDatumList = false;
 
         private IDreamManager _dreamManager = IoCManager.Resolve<IDreamManager>();
 
+        public override void OnObjectCreated(DreamObject dreamObject, DreamProcArguments creationArguments) {
+            if (dreamObject.IsSubtypeOf(DreamPath.Datum) && !dreamObject.IsSubtypeOf(DreamPath.Atom)) // Atoms are in world.contents
```
wixoaGit, line 13: The DreamObject here will always be a datum.

OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectDatum.cs
```diff
@@ -1,14 +1,31 @@
-﻿using Robust.Shared.IoC;
+﻿using OpenDreamRuntime.Procs;
+using OpenDreamShared.Dream;
+using Robust.Shared.IoC;
 
 namespace OpenDreamRuntime.Objects.MetaObjects {
     class DreamMetaObjectDatum : DreamMetaObjectRoot {
         public override bool ShouldCallNew => true;
+        private bool _inDatumList = false;
```
wixoaGit, line 8: There only exists a single meta-object for each type. This isn't correctly keeping track of what datums are in the datum list, which isn't necessary because `OnObjectCreated()` and `OnObjectDeleted()` should only be called once for each DreamObject.

wixoaGit (changes requested):

ike709:
Reviews addressed, survives basic testing.

DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -979,9 +979,9 @@ public DMASTProcStatement ProcStatement()
                     Newline();
 
 
-                    //Implicit "in world"
+                    //Implicit "in world", handled later
```
wixoaGit, line 982: This isn't an implicit "in world"

DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -979,9 +979,9 @@ public DMASTProcStatement ProcStatement()
...
                     if (variableDeclaration.Value is null && rangeEnd is null && defaultStep) {
-                        return new DMASTProcStatementForList(loc, initializer, variable, new DMASTIdentifier(loc, "world"), GetForBody());
+                        return new DMASTProcStatementForList(loc, initializer, variable, null, GetForBody());
```
wixoaGit, line 984: I think it would be better to have a new AST node for this, something like `DMASTProcStatementForType`

DMCompiler/DM/Visitors/DMProcBuilder.cs
```diff
@@ -412,8 +412,33 @@ public void ProcessStatementSet(DMASTProcStatementSet statementSet)
...
+                    _proc.PushPath(varDeclaration.Type.Value);
+                    _proc.CreateTypeEnumerator();
+                }
+                else
+                {
+                    // If List is null it's "for(var/atom/A)" (or an atom subtype) which is implicitly "in world"
+                    DMExpression.Emit(_dmObject, _proc, statementForList.List ?? new DMASTIdentifier(statementForList.Location, "world"));
+                    _proc.PushPath(varDeclaration.Type.Value);
+                    _proc.CreateListEnumerator();
+                }
```
wixoaGit, line 433: It would be better to always use `CreateTypeEnumerator` and have the runtime decide to use the world's contents if you're enumerating an atom type.

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -39,14 +39,17 @@ static class DMOpcodeHandlers {
             return null;
         }
 
-        public static ProcStatus? CreateListEnumerator(DMProcState state) {
+        public static ProcStatus? CreateListEnumerator(DMProcState state)
+        {
+            state.Pop().TryGetValueAsPath(out var type);
```
wixoaGit, line 44: This new pop from the stack never gets used

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -66,6 +69,31 @@ static class DMOpcodeHandlers {
             return null;
         }
 
+        public static ProcStatus? CreateTypeEnumerator(DMProcState state)
+        {
+            state.Pop().TryGetValueAsPath(out var type);
```
wixoaGit, line 74: This needs to error if it's not a path

wixoaGit (changes requested):

DMCompiler/DM/Visitors/DMProcBuilder.cs
```diff
@@ -406,8 +407,25 @@ public void ProcessStatementSet(DMASTProcStatementSet statementSet)
...
+                DMExpression.Emit(_dmObject, _proc, statementForList.List);
+                _proc.CreateListEnumerator();
+            }
+            else
+            {
+                DMExpression.Emit(_dmObject, _proc, statementForList.List);
+                _proc.CreateListEnumerator();
+            }
+
```
wixoaGit, line 428: Code is being repeated in both the `if` and `else` blocks
```suggestion
            DMASTProcStatementVarDeclaration varDeclaration = statementForList.Initializer as DMASTProcStatementVarDeclaration;

            DMExpression.Emit(_dmObject, _proc, statementForList.List);
            _proc.CreateListEnumerator();
```

DMCompiler/DM/Visitors/DMProcBuilder.cs
```diff
@@ -451,6 +460,55 @@ public void ProcessStatementSet(DMASTProcStatementSet statementSet)
...
+                    if (varDeclaration != null && varDeclaration.Type != null)
+                    {
+                        DMExpression.Emit(_dmObject, _proc, statementForType.Variable);
+                        _proc.PushPath(varDeclaration.Type.Value);
+                        _proc.IsType();
+
+                        _proc.ContinueIfFalse();
+                    }
+
```
wixoaGit, line 500: I don't think an `istype()` check is needed here.
```suggestion
```

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -39,14 +39,27 @@ static class DMOpcodeHandlers {
...
+        public static ProcStatus? CreateListEnumerator(DMProcState state)
+        {
+            var popped = state.Pop();
+            DreamList? list = null;
+            if (!popped.TryGetValueAsDreamObject(out var listObject))
+            {
+                if (popped.TryGetValueAsString(out var popStr) && popStr == "world")
+                {
+                    list = state.DreamManager.WorldContentsList;
+                }
```
wixoaGit, line 51: `for (var/A in "world")` doesn't enumerate `world.contents`

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -66,6 +79,38 @@ static class DMOpcodeHandlers {
...
+                list = DreamList.Create(state.DreamManager.Clients);
+            }
+            else if (state.DreamManager.ObjectTree.GetObjectDefinition(type).IsSubtypeOf(DreamPath.Atom))
+            {
+                list = state.DreamManager.WorldContentsList;
+            }
+            else if (state.DreamManager.ObjectTree.GetObjectDefinition(type).IsSubtypeOf(DreamPath.Datum))
+            {
+                list = DreamList.Create(state.DreamManager.Datums);
+            }
```
wixoaGit, line 102: A good optimization here would be to create a new enumerator that can enumerate a list of DreamObjects and return them as DreamValues. Using that here would remove the need to create a new DreamList.

wixoaGit (changes requested):

DMCompiler/DM/Visitors/DMProcBuilder.cs
```diff
@@ -451,6 +460,55 @@ public void ProcessStatementSet(DMASTProcStatementSet statementSet)
...
+                    _proc.BreakIfFalse();
+
+                    if (varDeclaration != null && varDeclaration.Type != null)
+                    {
+                        DMExpression.Emit(_dmObject, _proc, statementForType.Variable);
+                        _proc.PushPath(varDeclaration.Type.Value);
+                        _proc.IsType();
+
+                        _proc.ContinueIfFalse();
+                    }
```
wixoaGit, line 499: Having `CreateTypeEnumerator` only enumerate its given type instead of the more general `world.contents` or `state.DreamManager.Datums` would allow this `istype()` check to be removed

OpenDreamRuntime/Procs/DreamEnumerators.cs
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        DreamValue IEnumerator<DreamValue>.Current => _dreamValueEnumerator.Current;
+
+        public object Current => _dreamValueEnumerator.Current;
+
+        public bool MoveNext() {
+            bool hasNext = _dreamValueEnumerator.MoveNext();
+            if (_filterType != null) {
+                while (hasNext && (_dreamValueEnumerator.Current.Type != DreamValue.DreamValueType.DreamObject || !_dreamValueEnumerator.Current.TryGetValueAsDreamObjectOfType(_filterType.Value, out _)))
```
wixoaGit, line 87:
```suggestion
                while (hasNext && !_dreamValueEnumerator.Current.TryGetValueAsDreamObjectOfType(_filterType.Value, out _))
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #587, opened 2022-02-11
Faster typechecking implementation

ZeWaka:
Since our type tree in DM is single-inheritance, we can optimize as following:
> Types are enumerated in depth-first order so all children of a given parent type all have a contiguous indices; then check whether ID falls in that range.

This practice has been done in other game engines like DOOM 3, Roblox :godmode:, and Unity.

This is better than our existing approach of iterating through the whole path string looking if elements match. It'd also recurse through parents if `parent_type` was used or something also.
We could also use this DFS ordered list potentially for other things.

~11.1% speedup

Statistics & Profiling:
The primary thing you should pay attention to is the ratio (or difference between the trendlines).
New, Old, Ratio (1)

Multiple values were taken for each iteration count, for each category. Ran on Release.
The median was taken for the graph.

Test code:
```dm
/world/New()
    ..()

    var/obj/a/ancestor = new()
    var/obj/a/b/c/d/e/f/g/h/i/j/k/l/m/n/o/longancestor = new()
    var/obj/a/b/c/d/e/f/g/h/i/j/k/l/m/n/o/p/longchild = new()

    var/t1
    var/const/loops = 3300000

    var/time1 = world.timeofday
    for(var/i in 1 to loops)
        istypetest(longchild, ancestor, longancestor)
    t1 = world.timeofday-time1
    world.log << "TOTAL TIME FOR [loops] LOOPS: [t1]"

/proc/istypetest(longchild, ancestor, longancestor)
    var/x = istype(longchild, ancestor)
    var/y = istype(longchild, longancestor)
	var/z = istype(longchild, longchild)
```

### DIFF
--- a/OpenDreamShared/Dream/DreamPath.cs
+++ b/OpenDreamShared/Dream/DreamPath.cs
@@ -131,10 +131,10 @@ namespace OpenDreamShared.Dream {
             Normalize(false);
         }
 
-        public bool IsDerivedFrom(DreamPath ancestor)
-        {
-            return (typeIndex - ancestor.typeIndex) < ancestor.numChildren;
-        }
+        // public bool IsDescendantOf(DreamPath ancestor)
+        // {
+        //     return (typeIndex - ancestor.typeIndex) <= ancestor.numChildren;
+        // }
 
         public bool IsDescendantOf(DreamPath path) {
             if (path.Elements.Length > Elements.Length) return false;
--- a/OpenDreamShared/Dream/DreamPath.cs
+++ b/OpenDreamShared/Dream/DreamPath.cs
@@ -133,6 +133,7 @@ namespace OpenDreamShared.Dream {
 
         public bool IsDescendantOf(DreamPath ancestor)
         {
+            // Unsigned overflow is desired, see: Doom3/unity's impl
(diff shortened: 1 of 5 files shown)


OpenDreamShared/Dream/DreamPath.cs
```diff
@@ -124,14 +131,9 @@ public enum PathType {
...
-            if (path.Elements.Length > Elements.Length) return false;
-
-            for (int i = 0; i < path.Elements.Length; i++) {
-                if (Elements[i] != path.Elements[i]) return false;
-            }
-
-            return true;
+        public bool IsDescendantOf(DreamPath ancestor)
+        {
+            return (typeIndex - ancestor.typeIndex) <= ancestor.numChildren;
```
ZeWaka, line 136:
```suggestion
            // Unsigned overflow is desired, see: Doom3/unity's impl
            return (typeIndex - ancestor.typeIndex) <= ancestor.numChildren;
```

wixoaGit:
DreamPath should have nothing to do with type inheritance, it is supposed to be a raw representation of a path. It would be better to put the `numChildren` and `typeIndex` fields on TreeEntry instead.

ZeWaka:
See: https://github.com/wixoaGit/OpenDream/pull/664

PR #589, opened 2022-02-13
Seals some classes and cleans up some unused imports

ZeWaka:
title

Reasons for not sealing:
NumberPrompt: people might want to make a binary-only or some other weird prompt??
TextPrompt: Could keep this single line, and extend another one to be multiline? unsure what TextPrompt and what MessagePrompt should cover
InputWindow: This can be instantiated, so it's not abstract.

### DIFF
--- a/OpenDreamClient/Audio/DreamSoundChannel.cs
+++ b/OpenDreamClient/Audio/DreamSoundChannel.cs
@@ -1,9 +1,8 @@
-using System;
 using Robust.Client.Graphics;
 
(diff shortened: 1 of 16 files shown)


wixoaGit (approved):

wixoaGit:
> TextPrompt: Could keep this single line, and extend another one to be multiline? unsure what TextPrompt and what MessagePrompt should cover

I don't see any reason to merge TextPrompt and MessagePrompt. MessagePrompt would still have two override the two methods it already does anyways.

PR #593, opened 2022-02-15
Adds a ProcAttributes enum for boolean proc settings

ike709:
Per the ref, every setting that I can find is a bool except `name`, `desc`, `category`, and `invisibility` (I'm not counting `src`). The ones that we've added are also booleans.

So since most of them are bools, I went ahead and refactored WaitFor into an enum of setting flags.

I'll do `name`, `desc`, and `invisibility` in a follow-up.

`hidden`, `instant`, `background`, & `popup_menu` are set by the compiler but not currently used by the runtime.

Also `WhenWritingDefault` prevents serialization of the default enum value. Setting it to null directly while marking it as nullable didn't seem to work.

### DIFF
--- a/DMCompiler/DM/Visitors/DMProcBuilder.cs
+++ b/DMCompiler/DM/Visitors/DMProcBuilder.cs
@@ -161,6 +161,11 @@ namespace DMCompiler.DM.Visitors {
                     {
                         _proc.Attributes &= ~ProcAttributes.Hidden;
                     }
+                    
+                    if (!DMCompiler.Settings.SuppressUnimplementedWarnings) {
+                        DMCompiler.Warning(new CompilerWarning(statementSet.Location, "hidden is not implemented"));
+                    }
+                    
                     break;
                 case "popup_menu":
                     if (constant.IsTruthy()) // The default is to show it so we flag it if it's hidden
@@ -171,6 +176,11 @@ namespace DMCompiler.DM.Visitors {
                     {
                         _proc.Attributes |= ProcAttributes.HidePopupMenu;
                     }
+                    
+                    if (!DMCompiler.Settings.SuppressUnimplementedWarnings) {
+                        DMCompiler.Warning(new CompilerWarning(statementSet.Location, "popup_menu is not implemented"));
+                    }
+                    
                     break;
                 case "instant":
                     if (constant.IsTruthy())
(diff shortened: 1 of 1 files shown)


OpenDreamShared/Dream/Procs/ProcAttributes.cs
```diff
@@ -0,0 +1,18 @@
+using System;
+using System.ComponentModel;
+
+namespace OpenDreamShared.Dream.Procs;
+
+[Flags]
+[DefaultValue(None)]
+public enum ProcAttributes : byte
```
ike709, line 8:
```suggestion
public enum ProcAttributes
```

OpenDreamShared/Dream/Procs/ProcAttributes.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace OpenDreamShared.Dream.Procs;
+
+[Flags]
+[DefaultValue(None)]
+public enum ProcAttributes
+{
+    None = 1 << 0, // Internal
+    IsOverride = 1 << 1, // Internal
+    Unimplemented = 1 << 2,
```
ZeWaka, line 12: Is this not internal too?

ike709: Not really? It's set in DM code by a coder. It just happens to be unique to OD.

DMCompiler/DM/Visitors/DMProcBuilder.cs
```diff
@@ -129,25 +129,77 @@ class DMProcBuilder {
...
-                        throw new CompileErrorException(statementSet.Location, $"waitfor attribute should be a constant");
+                    if (constant.IsTruthy())
+                    {
+                        // "waitfor" defaults to enabled
+                        _proc.Attributes &= ~ProcAttributes.DisableWaitfor;
+                    }
+                    else
+                    {
+                        _proc.Attributes |= ProcAttributes.DisableWaitfor;
                     }
```
wixoaGit, line 149: There is a method on DMProc that does this
```suggestion
                    _proc.WaitFor(constant.IsTruthy());
```

DMCompiler/DM/Visitors/DMProcBuilder.cs
```diff
@@ -129,25 +129,77 @@ class DMProcBuilder {
...
+                    }
+                    else
+                    {
+                        _proc.Attributes &= ~ProcAttributes.Unimplemented;
                     }
-
-                    _proc.Unimplemented = constant.IsTruthy();
                     break;
                 }
+                case "hidden":
```
wixoaGit, line 163: All of these unimplemented attributes should print an unimplemented warning

wixoaGit (changes requested):

ike709:
Review addressed

PR #596, opened 2022-02-18
Adds the remaining verb settings

ike709:
They're not used by the runtime but all of them (minus src) do get set in the runtime on the DMProc.

### DIFF
--- a/DMCompiler/DM/Visitors/DMProcBuilder.cs
+++ b/DMCompiler/DM/Visitors/DMProcBuilder.cs
@@ -253,6 +253,9 @@ namespace DMCompiler.DM.Visitors {
                         _proc.Invisibility = Convert.ToSByte(Math.Clamp(Math.Floor(invisFloat.Value), 0, 100));
                     }
 
+                    if (!DMCompiler.Settings.SuppressUnimplementedWarnings) {
+                        DMCompiler.Warning(new CompilerWarning(statementSet.Location, "set invisibility is not implemented"));
+                    }
                     break;
                 case "src":
                     if (!DMCompiler.Settings.SuppressUnimplementedWarnings) {
--- a/DMCompiler/DM/Visitors/DMProcBuilder.cs
+++ b/DMCompiler/DM/Visitors/DMProcBuilder.cs
@@ -1,4 +1,4 @@
-﻿using OpenDreamShared.Compiler;
+using OpenDreamShared.Compiler;
 using DMCompiler.Compiler.DM;
 using System.Collections.Generic;
 using OpenDreamShared.Dream;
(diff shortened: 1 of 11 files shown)


OpenDreamShared/Dream/Procs/ProcAttributes.cs
```diff
@@ -4,7 +4,6 @@
 namespace OpenDreamShared.Dream.Procs;
 
 [Flags]
-[DefaultValue(None)]
```
ike709, line 7: This doesn't actually seem to be necessary.

DMCompiler/DM/Visitors/DMProcBuilder.cs
```diff
@@ -205,9 +205,58 @@ public void ProcessStatementSet(DMASTProcStatementSet statementSet)
...
+                        DMASTConstantInteger invisInt = statementSet.Value as DMASTConstantInteger;
+                        if(invisInt is null) throw new CompileErrorException(statementSet.Location, "bad num");
+                        _proc.Invisibility = Convert.ToSByte(Math.Clamp(invisInt.Value, 0, 100));
+                    }
+                    else
+                    {
+                        _proc.Invisibility = Convert.ToSByte(Math.Clamp(Math.Floor(invisFloat.Value), 0, 100));
+                    }
+
+                    break;
```
wixoaGit, line 256:
```suggestion
                    if (!DMCompiler.Settings.SuppressUnimplementedWarnings) {
                        DMCompiler.Warning(new CompilerWarning(statementSet.Location, "set invisibility is not implemented"));
                    }
                    break;
```

OpenDreamShared/Dream/Procs/ProcAttributes.cs
wixoaGit: The default value of an enum is whichever has a value of `0`. There is no value here that is zero because `None` is `1 << 0` which is `1`.

wixoaGit (changes requested):

OpenDreamShared/Dream/Procs/ProcAttributes.cs
wixoaGit: This leads the default value of this enum to be `(ProcAttributes)0` because there is no value of `ProcAttributes` equal to zero. `ProcAttributes.None` is `1 << 0` which is `1`.

OpenDreamShared/Dream/Procs/ProcAttributes.cs
```diff
@@ -4,7 +4,6 @@
...
 public enum ProcAttributes
 {
     None = 1 << 0, // Internal
```
ike709, line 9:
```suggestion
    None = 0, // Internal
```

OpenDreamShared/Dream/Procs/ProcAttributes.cs
ike709: Is it fine to just set None to 0? Because I did.

wixoaGit: Yep

wixoaGit (approved):

PR #599, opened 2022-02-28
Implements verb name, category, and hidden & improves the interface

ike709:
Also makes ProcAttributes both work again and not be unnecessarily serialized.

I also improved the interface a bit and fixed the initial tab not being updated until after switching tabs,

Pics of the DMF improvements (most/all of those verbs don't actually have their name set, I promise it does work):
image
image

### DIFF
--- a/OpenDreamShared/Dream/Procs/ProcAttributes.cs
+++ b/OpenDreamShared/Dream/Procs/ProcAttributes.cs
@@ -6,7 +6,7 @@ namespace OpenDreamShared.Dream.Procs;
 [Flags]
 public enum ProcAttributes
 {
-    None = 0, // Internal
+    None = 1 << 0, // Internal
     IsOverride = 1 << 1, // Internal
     Unimplemented = 1 << 2,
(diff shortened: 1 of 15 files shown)


OpenDreamShared/Dream/Procs/ProcAttributes.cs
```diff
@@ -6,7 +6,7 @@ namespace OpenDreamShared.Dream.Procs;
 [Flags]
 public enum ProcAttributes
 {
-    None = 0, // Internal
+    None = 1 << 0, // Internal
```
wixoaGit, line 9: Why was this changed to 1?

wixoaGit (changes requested):

OpenDreamShared/Dream/Procs/ProcAttributes.cs
ike709: Because it made the default value setting serialize it properly.

wixoaGit (approved):

PR #601, opened 2022-03-02
Makes bad compiler args into compiler warnings

ike709:


wixoaGit:
Why is this preferable? It seems to me that if you make a mistake in the arguments you wouldn't want the compiler to go on.

ike709:
Because I'm tired of needing to remember to remove `--experimental-preproc` from all of my Rider configurations.

ike709:
TODO: Add a flag to cope with bad flags.

DMCompiler/Program.cs
```diff
@@ -16,22 +18,29 @@ class Program {
...
                     default: {
                         string extension = Path.GetExtension(arg);
 
                         if (!String.IsNullOrEmpty(extension) && (extension == ".dme" || extension == ".dm")) {
                             settings.Files.Add(arg);
                             Console.WriteLine($"Compiling {Path.GetFileName(arg)}");
                         } else {
-                            Console.WriteLine($"Invalid arg '{arg}'");
-                            return false;
+                            if(skipBad || args.Contains("--skip-bad-args")) {
```
wixoaGit, line 38: The `skipBad` variable can be removed if you're just going to end up doing `args.Contains("--skip-bad-args")` anyways.

wixoaGit (changes requested):

DMCompiler/Program.cs
wixoaGit: That or initialize `skipBad` to be the Contains call

wixoaGit (approved):

PR #602, opened 2022-03-02
Improves the interface presentation a bit

ike709:
Will probably conflict with https://github.com/wixoaGit/OpenDream/pull/599 but it'll be easy to fix

I also fixed the initial tab not being updated until after switching tabs,

Pics:
image
image


ike709:
Ah hell

PR #605, opened 2022-03-03
Implements `background`

### LINKED ISSUE #247: `background` not implemented
http://www.byond.com/docs/ref/#/proc/set/background

ike709:
It... just works?

Closes #247

### DIFF
--- a/DMCompiler/DM/Visitors/DMProcBuilder.cs
+++ b/DMCompiler/DM/Visitors/DMProcBuilder.cs
@@ -57,6 +57,14 @@ namespace DMCompiler.DM.Visitors {
             foreach (DMASTProcStatement statement in block.Statements) {
                 try
                 {
+                    // ProcessStatementSet() needs to be before any loops
+                    var set = statement as DMASTProcStatementSet;
+                    if (set != null)
+                    {
+                        ProcessStatementSet(set);
+                        continue;
+                    }
+
                     ProcessStatement(statement);
                 }
                 catch (CompileAbortException e)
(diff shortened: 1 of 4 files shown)


wixoaGit:
Does the correct behavior for background look like this:
```
for(var/i in 1 to 3)
    world.log << i
    sleep(-1)
```
or this?
```
for(var/i in 1 to 3)
    world.log << i
sleep(-1)
```

The DM reference states `It would only ever be interrupted at the end of the for or while loops.`

ike709:
> You could achieve a similar thing by sprinkling calls to sleep(0) or sleep(-1) in the "working" part of the loop.

This implies the former, IMO.

Also it seems like it wouldn't be that useful if it applied _outside_ of the looping

ike709:
I wonder if I should actually do it in `LoopContinue()`

DMCompiler/DM/Visitors/DMProcBuilder.cs
```diff
@@ -57,6 +57,14 @@ class DMProcBuilder {
             foreach (DMASTProcStatement statement in block.Statements) {
                 try
                 {
+                    // ProcessStatementSet() needs to be before any loops
```
wixoaGit, line 60: This doesn't have any affect. It's just moving the `DMASTProcStatementSet` check outside of `ProcessStatement()`, it's still being processed in the same order.

wixoaGit (changes requested):

wixoaGit (approved):
In the future it'd be better to have a sleep opcode instead of depending on the runtime implementing a `sleep()` proc, but this works for now.

PR #612, opened 2022-03-06
Fixes `||=` not including path info

### LINKED ISSUE #611: Doing `LogicalOrAssign` aka `||=` with an inferred `new` doesn't work
Toplevel var: `var/database/db_handler_props`
Code:
```
var/database/db_handler_props

Prop/proc/LoadSprite()
    db_handler_props ||= new("databases/props.db")
    // other code excluded
```
Error: `An inferred new requires a type!`

ike709:
I'm not entirely sure this is the most correct way to do this, but it fixes #611 

Also changed a random bit of code to a switch and fixed a typo.

### DIFF
--- a/DMCompiler/Compiler/DM/DMParser.cs
+++ b/DMCompiler/Compiler/DM/DMParser.cs
@@ -515,9 +515,8 @@ namespace DMCompiler.Compiler.DM {
                     case DMASTIdentifier identifier:
                         Check(TokenType.DM_Colon);
                         return Label(identifier);
-                    case DMASTLeftShift shift:
+                    case DMASTLeftShift leftShift:
                     {
-                        DMASTLeftShift leftShift = shift;
                         DMASTProcCall procCall = leftShift.B as DMASTProcCall;
 
                         if (procCall != null && procCall.Callable is DMASTCallableProcIdentifier identifier) {
--- a/DMCompiler/Compiler/DM/DMParser.cs
+++ b/DMCompiler/Compiler/DM/DMParser.cs
@@ -306,7 +306,7 @@ namespace DMCompiler.Compiler.DM {
         public bool PathArray(ref DreamPath path, out DMASTExpression implied_value) {
             //TODO: Multidimensional lists
             implied_value = null;
-            if (Check(TokenType.DM_LeftBracket)) 
+            if (Check(TokenType.DM_LeftBracket))
             {
                 var loc = Current().Location;
                 if (!path.IsDescendantOf(DreamPath.List)) {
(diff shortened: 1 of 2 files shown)


DMCompiler/DM/Visitors/DMVisitorExpression.cs
```diff
@@ -256,12 +257,12 @@ internal DMVisitorExpression(DMObject dmObject, DMProc proc, DreamPath? inferred
 
         public void VisitLogicalAndAssign(DMASTLogicalAndAssign land) {
             var lhs = DMExpression.Create(_dmObject, _proc, land.A, _inferredPath);
-            var rhs = DMExpression.Create(_dmObject, _proc, land.B, _inferredPath);
+            var rhs = DMExpression.Create(_dmObject, _proc, land.B, lhs.Path ?? _inferredPath);
```
wixoaGit, line 260: This should just be `lhs.Path`, similar to `VisitAssign()`

`var1 |= var2 |= new()` does not use `var1`'s type if `var2` doesn't have one.
```suggestion
            var rhs = DMExpression.Create(_dmObject, _proc, land.B, lhs.Path);
```

DMCompiler/DM/Visitors/DMVisitorExpression.cs
```diff
@@ -256,12 +257,12 @@ internal DMVisitorExpression(DMObject dmObject, DMProc proc, DreamPath? inferred
...
             Result = new Expressions.LogicalAndAssign(land.Location, lhs, rhs);
         }
         public void VisitLogicalOrAssign(DMASTLogicalOrAssign lor) {
             var lhs = DMExpression.Create(_dmObject, _proc, lor.A, _inferredPath);
-            var rhs = DMExpression.Create(_dmObject, _proc, lor.B, _inferredPath);
+            var rhs = DMExpression.Create(_dmObject, _proc, lor.B, lhs.Path ?? _inferredPath);
```
wixoaGit, line 265:
```suggestion
            var rhs = DMExpression.Create(_dmObject, _proc, lor.B, lhs.Path);
```

DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -508,34 +508,41 @@ public DMASTProcStatement ProcStatement()
...
+            if (expression != null)
+            {
+                switch (expression)
+                {
+                    case DMASTIdentifier identifier:
+                        Check(TokenType.DM_Colon);
+                        return Label(identifier);
+                    case DMASTLeftShift shift:
+                    {
+                        DMASTLeftShift leftShift = shift;
```
wixoaGit, line 520:
```suggestion
                    case DMASTLeftShift leftShift:
                    {
```

wixoaGit (changes requested):

PR #614, opened 2022-03-06
Error tolerance refactor for map parsing/saving

ike709:
Things this does:

- DMMParser has been edited to use normal compiler errors/warnings in several places
- The maps are now parsed & serialized _before_ compilation is considered successful, since it can fail
- The JSON is no longer saved if map parsing fails
- DMMParser now passes type info to DMExpression.Create()
- DMMParser can now cope with duplicate var overrides in a modified type, and will create a warning about it. The last parsed value will be used as the true override, not sure if this matches BYOND behavior.
- TryAsJsonRepresentation() now throws an unimplemented warning when it's unimplemented
- TryAsJsonRepresentation() for ProcCall has been stubbed
- Fixed a null reference exception with printing unimplemented warnings for global procs
- Trying to load multiple maps is now a logger error rather than a not implemented exception

Once the Eternia compiler fixes are merged, this PR will make it also survive map parsing, saving, & loading.

### DIFF
--- a/DMCompiler/Compiler/DMM/DMMParser.cs
+++ b/DMCompiler/Compiler/DMM/DMMParser.cs
@@ -68,11 +68,14 @@ namespace DMCompiler.Compiler.DMM {
                         while (statement != null) {
                             DMASTObjectVarOverride varOverride = statement as DMASTObjectVarOverride;
                             if (varOverride == null) Error("Expected a var override");
-                            if (!varOverride.ObjectPath.Equals(DreamPath.Root)) Error("Invalid var name");
-                            DMExpression value = DMExpression.Create(null, null, varOverride.Value);
-                            if (!value.TryAsJsonRepresentation(out var valueJson)) Error($"Failed to serialize value to json ({value})");
+                            if (!varOverride.ObjectPath.Equals(DreamPath.Root)) DMCompiler.Error(new CompilerError(statement.Location, $"Invalid var name '{varOverride.VarName}' in DMM on type {objectType.Path}"));
+                            DMExpression value = DMExpression.Create(DMObjectTree.GetDMObject(objectType.Path), null, varOverride.Value);
+                            if (!value.TryAsJsonRepresentation(out var valueJson)) DMCompiler.Error(new CompilerError(statement.Location, $"Failed to serialize value to json ({value})"));
 
-                            mapObject.AddVarOverride(varOverride.VarName, valueJson);
+                            if(!mapObject.AddVarOverride(varOverride.VarName, valueJson))
+                            {
+                                DMCompiler.Warning(new CompilerWarning(statement.Location, $"Duplicate var override '{varOverride.VarName}' in DMM on type {objectType.Path}"));
+                            }
 
                             if (Check(TokenType.DM_Semicolon)) {
                                 statement = Statement(requireDelimiter: false);
@@ -93,7 +96,7 @@ namespace DMCompiler.Compiler.DMM {
                             cellDefinition.Objects.Add(mapObject);
                         }
                     }
-                    
+
 
                     if (Check(TokenType.DM_Comma)) {
                         objectType = Path();
--- a/DMCompiler/DM/Expressions/Builtins.cs
+++ b/DMCompiler/DM/Expressions/Builtins.cs
@@ -362,7 +362,7 @@ namespace DMCompiler.DM.Expressions {
             json = null;
             if (!DMCompiler.Settings.SuppressUnimplementedWarnings)
             {
-                DMCompiler.Warning(new CompilerWarning(Location, $"DMM overrides for expression {GetType()} are not implemented"));
+                DMCompiler.Warning(new CompilerWarning(Location, "DMM overrides for newlist() are not implemented"));
             }
             return true; //TODO
         }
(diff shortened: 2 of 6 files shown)


ike709:
Good to go, just need to write the desc.

ike709:
There.

DMCompiler/Compiler/DMM/DMMParser.cs
```diff
@@ -68,11 +68,14 @@ class DMMParser : DMParser {
...
-                            if (!value.TryAsJsonRepresentation(out var valueJson)) Error($"Failed to serialize value to json ({value})");
+                            if (!varOverride.ObjectPath.Equals(DreamPath.Root)) DMCompiler.Error(new CompilerError(statement.Location, $"Invalid var name '{varOverride.VarName}' in DMM on type {objectType.Path}"));
+                            DMExpression value = DMExpression.Create(DMObjectTree.GetDMObject(objectType.Path), null, varOverride.Value);
+                            if (!value.TryAsJsonRepresentation(out var valueJson)) DMCompiler.Error(new CompilerError(statement.Location, $"Failed to serialize value to json ({value})"));
 
-                            mapObject.AddVarOverride(varOverride.VarName, valueJson);
+                            if(!mapObject.AddVarOverride(varOverride.VarName, valueJson))
+                            {
+                                DMCompiler.Warning(new CompilerWarning(statement.Location, $"Duplicate var override '{varOverride.VarName}' in DMM on type {objectType.Path}"));
+                            }
```
wixoaGit, line 78: Might be better to move this warning into `AddVarOverride()` itself. DM code can also override the same var multiple times.

DMCompiler/DM/Expressions/Builtins.cs
```diff
@@ -360,6 +360,10 @@ class NewList : DMExpression {
 
         public override bool TryAsJsonRepresentation(out object json) {
             json = null;
+            if (!DMCompiler.Settings.SuppressUnimplementedWarnings)
+            {
+                DMCompiler.Warning(new CompilerWarning(Location, $"DMM overrides for expression {GetType()} are not implemented"));
```
wixoaGit, line 365:
```suggestion
                DMCompiler.Warning(new CompilerWarning(Location, "DMM overrides for newlist() are not implemented"));
```

DMCompiler/DM/Expressions/Procs.cs
```diff
@@ -133,5 +133,14 @@ class ProcCall : DMExpression {
...
         }
+
+        public override bool TryAsJsonRepresentation(out object json) {
+            json = null;
+            if (!DMCompiler.Settings.SuppressUnimplementedWarnings)
+            {
+                DMCompiler.Warning(new CompilerWarning(Location, $"DMM overrides for expression {GetType()} are not implemented"));
+            }
+            return true; //TODO
+        }
```
wixoaGit, line 144: Proc calls cannot be serialized into json. Was this added because of a `rgb()` call or similar?

DMCompiler/DMCompiler.cs
```diff
@@ -52,15 +52,24 @@ public static class DMCompiler {
...
             bool successfulCompile = preprocessor is not null && Compile(preprocessor.GetResult());
 
-            if (successfulCompile) {
-                Console.WriteLine($"Compilation succeeded with {WarningCount} warnings");
-
+            if (successfulCompile)
+            {
                 //Output file is the first file with the extension changed to .json
                 string outputFile = Path.ChangeExtension(settings.Files[0], "json");
                 List<DreamMapJson> maps = ConvertMaps(preprocessor.IncludedMaps);
```
wixoaGit, line 59: Compiling should exit early if `ErrorCount > 0` after converting the maps, and not go on to call `SaveJson()`

OpenDreamRuntime/DreamMapManager.cs
```diff
@@ -73,7 +73,10 @@ public struct Cell {
 
         public void LoadMaps(List<DreamMapJson> maps) {
             if (maps.Count == 0) throw new ArgumentException("No maps were given");
-            else if (maps.Count > 1) throw new NotImplementedException("Loading more than one map is not implemented");
+            else if (maps.Count > 1)
+            {
+                Logger.Error("Loading more than one map is not implemented, skipping additional maps");
```
wixoaGit, line 78: Seems more like a warning to me
```suggestion
                Logger.Warning("Loading more than one map is not implemented, skipping additional maps");
```

wixoaGit (changes requested):

DMCompiler/DMCompiler.cs
```diff
@@ -181,7 +197,9 @@ public static class DMCompiler {
             return maps;
         }
 
-        private static void SaveJson(List<DreamMapJson> maps, string interfaceFile, string outputFile) {
+        private static string SaveJson(List<DreamMapJson> maps, string interfaceFile, string outputFile)
+        {
+            var beforeErrorCount = DMCompiler.ErrorCount;
```
wixoaGit, line 202: `ErrorCount` should always be zero at this point.
```suggestion
```

DMCompiler/DMCompiler.cs
```diff
@@ -223,8 +241,13 @@ public static class DMCompiler {
                 DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingDefault
             });
 
-            File.WriteAllText(outputFile, json);
-            Console.WriteLine("Saved to " + outputFile);
+            // Successful serialization
+            if (DMCompiler.ErrorCount == beforeErrorCount)
```
wixoaGit, line 245:
```suggestion
            if (ErrorCount > 0)
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #616, opened 2022-03-06
Adds `TryAsConstant()` override for derefs

ike709:
Not fully tested because I still struggle with wrapping my brain around derefs, but it fixes the Eternia error.

### DIFF
--- a/DMCompiler/DM/Expressions/Dereference.cs
+++ b/DMCompiler/DM/Expressions/Dereference.cs
@@ -50,6 +50,19 @@ namespace DMCompiler.DM.Expressions {
             _expr.EmitPushValue(dmObject, proc);
             return (DMReference.CreateField(_propertyName), _conditional);
         }
+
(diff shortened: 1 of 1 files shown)


DMCompiler/DM/Expressions/Dereference.cs
```diff
@@ -50,6 +50,19 @@ public Dereference(Location location, DreamPath? path, DMExpression expr, bool c
             _expr.EmitPushValue(dmObject, proc);
             return (DMReference.CreateField(_propertyName), _conditional);
         }
+
+        public override bool TryAsConstant(out Constant constant)
+        {
+            var obj = DMObjectTree.GetDMObject(_expr.Path.GetValueOrDefault());
```
wixoaGit, line 56: This should fail if `_expr.Path` is null

wixoaGit (changes requested):

wixoaGit (approved):

PR #620, opened 2022-03-07
Makes Eternia init + misc cleanup

ike709:
- Misc Rider-suggested code cleanup
- sealed some things
- Adds coping with null `_current`
- Adds tolerance for `color = ""` which BYOND allows in modified types in DMMs.

### DIFF
--- a/OpenDreamRuntime/DreamMapManager.cs
+++ b/OpenDreamRuntime/DreamMapManager.cs
@@ -73,7 +73,7 @@ namespace OpenDreamRuntime {
 
         public void LoadMaps(List<DreamMapJson> maps) {
             if (maps.Count == 0) throw new ArgumentException("No maps were given");
-            else if (maps.Count > 1) throw new NotImplementedException("Loading more than one map is not implemented");
+            //else if (maps.Count > 1) throw new NotImplementedException("Loading more than one map is not implemented");
             DreamMapJson map = maps[0];
 
             Size = new Vector2i(map.MaxX, map.MaxY);
--- a/OpenDreamRuntime/DreamThread.cs
+++ b/OpenDreamRuntime/DreamThread.cs
@@ -47,10 +47,10 @@ namespace OpenDreamRuntime {
             Invisibility = invisibility;
         }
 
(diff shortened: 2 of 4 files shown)


OpenDreamRuntime/DreamMapManager.cs
```diff
@@ -73,7 +73,7 @@ public struct Cell {
 
         public void LoadMaps(List<DreamMapJson> maps) {
             if (maps.Count == 0) throw new ArgumentException("No maps were given");
-            else if (maps.Count > 1) throw new NotImplementedException("Loading more than one map is not implemented");
+            //else if (maps.Count > 1) throw new NotImplementedException("Loading more than one map is not implemented");
```
ike709, line 76:
```suggestion
            else if (maps.Count > 1) throw new NotImplementedException("Loading more than one map is not implemented");
```

ZeWaka:
Do we really want to be putting `readonly` on fields? I understand it's a micro-opt, but it might be worth profiling times for.
https://codeblog.jonskeet.uk/2014/07/16/micro-optimization-the-surprising-inefficiency-of-readonly-fields/

ike709:
@ZeWaka https://devblogs.microsoft.com/premier-developer/avoiding-struct-and-readonly-reference-performance-pitfalls-with-errorprone-net/

ike709:
and https://devblogs.microsoft.com/premier-developer/the-in-modifier-and-the-readonly-structs-in-c/

OpenDreamRuntime/DreamThread.cs
```diff
@@ -107,10 +107,10 @@ public abstract class ProcState {
         public virtual void ReturnedInto(DreamValue value) {}
     }
 
-    public class DreamThread {
+    public sealed class DreamThread {
         private const int MaxStackDepth = 256;
 
-        private ProcState _current;
+        private ProcState? _current;
```
wixoaGit, line 113: What was causing `_current` to be null?

wixoaGit (changes requested):

OpenDreamRuntime/DreamThread.cs
ike709: Can't reproduce, probably got fixed by one of the other dozen eternia PRs.

PR #628, opened 2022-03-10
Nuke `list()` allocs and fix `del()`

ike709:
- Lists can now allocate the correct size initially
- `EnumerateListing()` now returns a string array and as a bonus I made it append `/` to dirs like the ref says `flist()` does.
- `public static DreamList Create(IEnumerable<object> collection)` is now `public static DreamList Create(string[] collection)` since they're all string arrays anyways
- `del()` no longer throws an exception. `OnObjectDeleted()` tries to access the `loc` variable, so it needs to not be `Deleted` yet.


ike709:
For the test case I used, this reduced peak memory usage by just over ~50% for OpenDream and reduced total allocations of DreamList by ~38%

wixoaGit (approved):

PR #637, opened 2022-03-15
Implements trust levels

### LINKED ISSUE #630: Implement trust levels
Since SS14 server sandboxing isn't really a thing, we need to implement BYOND trustlevels.

ike709:
Closes #630

**This needs extensive testing / a good review**. I want trustlevels implemented before I set up an opendream compiler discord bot, and it'd be nice if they actually work.

I've done some minimal testing with `fexists()` but I can't think of every possible bypass attempt.

Also, when I was unsure of whether or not something should check the trustlevel I erred on the side of caution except when doing so would block TestGame from starting in Ultrasafe.


ZeWaka:
>no docker for the bot

ike709:
The bot will be in a docker container. I'm just paranoid.

ike709:
Need to tweak this to utilize https://github.com/space-wizards/RobustToolbox/commit/68f89c895846c3dcf8e514f238a3837b5486f8cb in ultrasafe.

ike709:
This is something that should be revisited in the future (someone else feel free to take over), but honestly I have no desire to revisit it and I might just press on with making a Discord OD bot without it.

PR #640, opened 2022-03-17
Recompile DMMs without recompiling code

ike709:
Adds the ability to recompile DMMs without recompiling game code.

Right now this is fairly limited, and only supports making edits to existing files. It also recompiles *all* maps instead of just the edited map. Detection also isn't instantaneous and can take a few seconds. But all of that can be improved in the future.

It also lays some groundwork for leaving the program running, which will only happen if the `--incremental-dmm` arg is passed. But this is needed to support features like actual incremental compilation, a REPL, hotloading(?), etc.


ike709:
I'll figure out what I broke when I have time.

ike709:
It's a fun experiment but I can't be bothered to actually maintain this at the moment with my other backlog of PRs.

PR #646, opened 2022-03-19
Adds an OPENDREAM define

ike709:
example usecase:
```
#idef OPENDREAM
var/rustg = "rustg_x64.dll"
#else
var/rustg = "rustg.dll"
#endif
```


AffectedArc07:
Much appreciated

wixoaGit (approved):

PR #647, opened 2022-03-20
Implements `tag`

### LINKED ISSUE #641: Tags need to be implemented
Probably just have a dict of tags & atoms/datums or whatever?

### LINKED ISSUE #642: `locate()` needs to support tags
See #641 first

ike709:
Closes #641 
Closes #642 

Only did basic testing with a single `locate()` followed by `del()` then another `locate()`. Seems fine, though.


OpenDreamRuntime/DreamManager.cs
```diff
@@ -37,6 +37,7 @@ partial class DreamManager : IDreamManager {
         public List<DreamObject> Mobs { get; set; } = new();
         public Random Random { get; set; } = new();
 
+        public Dictionary<string, Queue<DreamObject>> Tags { get; set; } = new();
```
ike709, line 40:
```suggestion
        public Dictionary<string, Queue<DreamObject>> Tags { get; set; } = new();

```

DMCompiler/DMStandard/Types/Datum.dm
```diff
@@ -4,8 +4,12 @@
...
-	var/tag = null as opendream_unimplemented
+	var/tag
 
 	proc/New()
+
 	proc/Del()
-	proc/Topic(href, href_list)
\ No newline at end of file
+		if(tag)
+			tag = null
```
ike709, line 13: when it's not 5am I need to check and make sure the object is actually being removed from the queue when deleted

ike709: Seems fine

ike709:
Everything seems to work.

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -1385,9 +1385,21 @@ static class DMOpcodeHandlers {
...
+                {
+                    state.Push(new DreamValue(DreamObject.GetFromReferenceID(state.DreamManager, refID)));
+                }
+
+                var dreamMan = IoCManager.Resolve<IDreamManager>();
+                if (dreamMan.Tags.ContainsKey(refString))
+                {
+                    state.Push(new DreamValue(dreamMan.Tags[refString].Peek()));
+                }
+                else
```
ike709, line 1398:
```suggestion
                if(int.TryParse(refString, out var refID))
                {
                    state.Push(new DreamValue(DreamObject.GetFromReferenceID(state.DreamManager, refID)));
                }
                else if (state.DreamManager.Tags.ContainsKey(refString))
                {
                    state.Push(new DreamValue(state.DreamManager.Tags[refString].Peek()));
                }
                else
```

DMCompiler/DMStandard/Types/Datum.dm
```diff
@@ -4,8 +4,12 @@
...
 
-	var/tag = null as opendream_unimplemented
+	var/tag
 
 	proc/New()
+
 	proc/Del()
-	proc/Topic(href, href_list)
\ No newline at end of file
+		if(tag)
```
wixoaGit, line 12: This would lead to the object never being deleted if an override doesn't call `..()`. The DM reference says to always call `..()` in `Del()`, is this similar behavior?

OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectDatum.cs
```diff
@@ -28,15 +30,68 @@ class DreamMetaObjectDatum : DreamMetaObjectRoot {
...
+                            queue.Dequeue();
+                        }
+                        else
+                        {
+                            // We're working off of the assumption that the majority of tags are unique most of the time
+                            // But when it isn't we have to do this grossness to remove the correct obj from the queue
+                            var realList = queue.ToList();
+                            realList.Remove(dreamObject);
+                            var replacementQueue = new Queue<DreamObject>(realList);
+                            _dreamManager.Tags.Add(tagStr, replacementQueue);
```
wixoaGit, line 72: It may be better to replace the queues with `List<DreamObject>` to prevent from needing to do this. `Enqueue()` can be replaced with `Add()`, and `locate()` can return the first element in the list without it being any different.

wixoaGit (changes requested):

DMCompiler/DMStandard/Types/Datum.dm
ike709: See discord. This is correct behavior.

wixoaGit: This doesn't actually lead to the object never being deleted. OpenDream will still mark the object as deleted, but not calling `..()` here will lead to a "dangling" reference inside the runtime.

This should be moved to `MetaDreamObjectDatum.OnObjectDeleted()` and the prevention of deletion can be added more correctly outside of this PR.

wixoaGit (changes requested):

wixoaGit (approved):

PR #654, opened 2022-03-31
Misc opcode handler cleanup

ike709:
Had some time to kill so I converted some `GetValueAs` to their `Try` forms, made the Blend() append log to `world.log`, and also made `WriteWorldLog()` accept a sawmill.

### DIFF
--- a/OpenDreamRuntime/DreamManager.cs
+++ b/OpenDreamRuntime/DreamManager.cs
@@ -135,7 +135,7 @@ namespace OpenDreamRuntime {
             GlobalProcs[name] = proc;
         }
 
-        public void WriteWorldLog(string message, LogLevel level = LogLevel.Info)
+        public void WriteWorldLog(string message, LogLevel level = LogLevel.Info, string sawmill = "world.log")
         {
             if (!WorldInstance.GetVariable("log").TryGetValueAsDreamResource(out var logRsc))
             {
@@ -146,11 +146,11 @@ namespace OpenDreamRuntime {
 
             if (logRsc is ConsoleOutputResource) // Output() on ConsoleOutputResource uses LogLevel.Info
             {
-                Logger.LogS(level, "world.log", message);
+                Logger.LogS(level, sawmill, message);
             }
             else
             {
-                logRsc.Output(new DreamValue($"[{LogMessage.LogLevelToName(level)}] world.log: {message}"));
+                logRsc.Output(new DreamValue($"[{LogMessage.LogLevelToName(level)}] {sawmill}: {message}"));
             }
         }
     }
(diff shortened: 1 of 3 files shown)


OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -643,8 +655,8 @@ static class DMOpcodeHandlers {
                     result = second;
                 }
             } else if (second.Value != null) {
-                if (first.Type == DreamValue.DreamValueType.Float && second.Type == DreamValue.DreamValueType.Float) {
-                    result = new DreamValue(first.GetValueAsInteger() | second.GetValueAsInteger());
+                if (first.TryGetValueAsInteger(out var firstInt) && second.TryGetValueAsInteger(out var secondInt)) {
```
ike709, line 658: I'm assuming this is equivalent behavior but I can re-add the typechecks if not.

wixoaGit: It is equivalent behavior.

wixoaGit: It is equivalent behavior.

OpenDreamRuntime/DreamManager.cs
```diff
@@ -146,14 +146,14 @@ public void WriteWorldLog(string message, LogLevel level = LogLevel.Info)
...
+                Logger.LogS(level, sawmill, message);
             }
             else
             {
-                logRsc.Output(new DreamValue($"[{LogMessage.LogLevelToName(level)}] world.log: {message}"));
+                logRsc.Output(new DreamValue($"[{LogMessage.LogLevelToName(level)}] {sawmill}: {message}"));
                 if (_configManager.GetCVar(OpenDreamCVars.AlwaysShowExceptions))
                 {
-                    Logger.LogS(level, "world.log", message);
+                    Logger.LogS(level, $"{sawmill}", message);
```
wixoaGit, line 156:
```suggestion
                    Logger.LogS(level, sawmill, message);
```

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -1381,15 +1402,13 @@ static class DMOpcodeHandlers {
...
+                container.GetVariable("contents").TryGetValueAsDreamList(out containerList);
             } else {
                 containerList = container as DreamList;
             }
 
-            if (value.TryGetValueAsString(out string refString)) {
-                int refID = int.Parse(refString);
-
-                state.Push(new DreamValue(DreamObject.GetFromReferenceID(state.DreamManager, refID)));
+            if (value.TryGetValueAsString(out string refString) && int.TryParse(refString, out var refId)) {
```
wixoaGit, line 1410: BYOND seems to return null if the string is not a valid ref ID, even if the string appears in the list.

wixoaGit (changes requested):

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
wixoaGit: "valid ref ID" includes tags here, not just ref numbers.

wixoaGit (approved):

PR #655, opened 2022-04-02
Improves bad proc arg handling

### LINKED ISSUE #265: Can't pass `null` as a proc arg
`/obj/structure/spacevine/temperature_expose(null, temp, volume)`

### LINKED ISSUE #475: Modified types don't support a space before the {
I'm too tired to fix it rn but I don't want to forget about it.

ike709:
Fixes #265 
Fixes #475 

The error now matches BYOND (with the addition of the token string for the bad arg). Unlike BYOND it also errors for http://www.byond.com/forum/post/2736128

Error recovery should be robust enough to not need to skip.

Also slipped in a fix for #475 and fixed error recovery for `'else if'` switch cases.

Coded this at 4am so it's entirely possible that it could be done more efficiently, but it works and it's an error branch.


DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -1405,22 +1406,64 @@ public DMASTProcStatementLabel Label(DMASTIdentifier expression)
...
+                    parameters.AddRange(DefinitionParameters());
+                }
+
                 while (Check(TokenType.DM_Comma)) {
                     BracketWhitespace();
-
+                    if (Current().Location.Line == 83 && Current().Location.Column == 71)
+                    {
+                        Console.WriteLine("e");
+                    }
```
ike709:
```suggestion
```

DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -1405,22 +1406,60 @@ public DMASTProcStatementLabel Label(DMASTIdentifier expression)
             List<DMASTDefinitionParameter> parameters = new();
             DMASTDefinitionParameter parameter = DefinitionParameter();
 
-            if (parameter != null || Check(TokenType.DM_IndeterminateArgs)) {
+            if (parameter == null && !Check(TokenType.DM_IndeterminateArgs) && Current().Type != TokenType.DM_RightParenthesis && Current().Type != TokenType.EndOfFile)
+            {
+                // Breaking change - BYOND doesn't specify the arg
```
wixoaGit, line 1411: Is this really a "breaking" change?

DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -1405,22 +1406,60 @@ public DMASTProcStatementLabel Label(DMASTIdentifier expression)
...
+                Error($"error: bad argument definition '{Current().PrintableText}'", false);
+                Recover();
+
+            }
+            else {
                 if (parameter != null) parameters.Add(parameter);
 
+                BracketWhitespace();
+                if (Current().Type != TokenType.DM_RightParenthesis && Current().Type != TokenType.DM_Comma && !Check(TokenType.DM_IndeterminateArgs))
```
wixoaGit, line 1420: `DefinitionParameters()` is intended to only parse the vars (`var/a, var/b, var/c`) and not the parentheses. This check (and others) should be moved to `ProcCall()`.

wixoaGit (changes requested):

DMCompiler/Compiler/DM/DMParser.cs
ike709, line 1411:
```suggestion
                // BYOND doesn't specify the arg
```

Hinaichigo:
Will this also fix the error `Duplicate argument "null"` when `null` is passed as an arg multiple times to a proc?

wixoaGit:
> Will this also fix the error `Duplicate argument "null"` when `null` is passed as an arg multiple times to a proc?

Are you talking about procs defined like this?

`/proc/foo(null, null)`

If so, this won't fix that. It was an intentional decision to make 2 arguments with the same name a compile error. These are also not null values being passed to the proc, this is creating a var named `null` that shadows (and replaces) any uses of `null` which is likely unintentional.

Hinaichigo:
That's what I meant, yeah.
So allowing things like:
```
/obj/item/weapon/sword/executioner/afterattack(null, mob/living/user as mob|obj, null, null, null)
	if(complete)
		user.delayNextAttack(30) //thrice the regular attack delay
```
is another silent error with BYOND?

wixoaGit:
> That's what I meant, yeah. So allowing things like:
> 
> ```
> /obj/item/weapon/sword/executioner/afterattack(null, mob/living/user as mob|obj, null, null, null)
> 	if(complete)
> 		user.delayNextAttack(30) //thrice the regular attack delay
> ```
> 
> is another silent error with BYOND?

Yes, multiple parameters with the same name doesn't raise any errors in BYOND. Naming a parameter `null` is actually another issue.

ike709:
Correct

Hinaichigo:
So it's an issue masked by another issue. That makes sense.

DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -166,7 +168,24 @@ public DMASTStatement Statement(bool requireDelimiter = true)
                     if (Check(TokenType.DM_LeftParenthesis)) {
                         DMCompiler.VerbosePrint($"Parsing proc {_currentPath}()");
                         BracketWhitespace();
-                        DMASTDefinitionParameter[] parameters = DefinitionParameters();
+                        var parameters = DefinitionParameters();
+
+                        if (Current().Type != TokenType.DM_RightParenthesis && Current().Type != TokenType.DM_Comma &&
+                            !Check(TokenType.DM_IndeterminateArgs))
+                        {
+                            Error($"error: {parameters.Last().Name}: missing comma ',' or right-paren ')'", false);
```
wixoaGit, line 176: `parameters.Last()` throws an exception if there aren't any parameters.

For example: `proc/t(`

wixoaGit (changes requested):

wixoaGit (approved):

PR #659, opened 2022-04-05
Implements multi-dimensional lists in PathArray()

### LINKED ISSUE #657: `var/list/L[][]` multi-dimensional list syntax broken in object definitions
```
/obj/machinery/computer/arcade/minesweeper
	var/table[31][51]
```

A proc trying to do `table[y1][x1]` gets an unknown identifier error.

ike709:
Closes #657 

I don't see where `list doubly initialized` is printed in BYOND instead of just overriding the value.


DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -304,26 +304,37 @@ public DMASTStatement Statement(bool requireDelimiter = true)
...
 
         public bool PathArray(ref DreamPath path, out DMASTExpression implied_value) {
-            //TODO: Multidimensional lists
             implied_value = null;
-            if (Check(TokenType.DM_LeftBracket))
+            if (Current().Type == TokenType.DM_LeftBracket)
             {
                 var loc = Current().Location;
-                if (!path.IsDescendantOf(DreamPath.List)) {
+                if (!path.Elements[..^1].Contains("list")) {
```
ike709, line 311:
```suggestion
                // Trying to use path.IsDescendantOf(DreamPath.List) here doesn't work
                if (!path.Elements[..^1].Contains("list")) {
```

wixoaGit:
> I don't see where list doubly initialized is printed in BYOND instead of just overriding the value.

This warning gets printed when there is both a list size and an initializer
`var/L[0] = 3`

wixoaGit:
Some testing in Dream Maker shows that empty list sizes (`var/L[]`) does nothing but ensure the var's type is `/list`. `var/L[][][][10]` only creates a 1-dimensional list of size 10, and `var/L[][2][][10]` creates a 2-dimensional list of size 2 and 10. This implementation doesn't have that behavior.

wixoaGit (changes requested):
Some testing in Dream Maker shows that empty list sizes (`var/L[]`) do nothing but ensure the var's type is /list. `var/L[][][][10]` only creates a 1-dimensional list of size 10, and `var/L[][2][][10][]` creates a 2-dimensional list of size 2 and 10. This implementation doesn't have that behavior.

ike709:
Note to myself: Right now opendream doesn't really handle multidimensional lists properly and it needs a new expression to support them.

ike709:
Should match BYOND now

DMCompiler/Compiler/DM/DMAST.cs
```diff
@@ -1053,6 +1054,20 @@ public class DMASTNewPath : DMASTExpression {
         }
     }
 
+    public class DMASTNewMultidimensionalList : DMASTExpression {
+        public DMASTPath Path;
```
wixoaGit, line 1058: This property is unnecessary, it's always going to be creating a `/list`

DMCompiler/Compiler/DM/DMAST.cs
```diff
@@ -1053,6 +1054,20 @@ public class DMASTNewPath : DMASTExpression {
...
+        public DMASTCallParameter[] Dimensions;
```
wixoaGit, line 1059: This could be `DMASTExpression[]`. There's no need to turn them into parameters.

DMCompiler/DM/Expressions/Builtins.cs
```diff
@@ -77,6 +77,21 @@ class NewPath : DMExpression {
         }
     }
 
+    class NewMultidimensionalList : DMExpression {
+        DreamPath TargetPath;
```
wixoaGit, line 81: This property isn't needed or used.

DMCompiler/DM/Expressions/Builtins.cs
```diff
@@ -77,6 +77,21 @@ class NewPath : DMExpression {
...
+        ArgumentList Arguments;
+
+        public NewMultidimensionalList(Location location, DreamPath targetPath, ArgumentList arguments) : base(location) {
+            TargetPath = targetPath;
+            Arguments = arguments;
+        }
+
+        public override void EmitPushValue(DMObject dmObject, DMProc proc) {
+            Arguments.EmitPushArguments(dmObject, proc);
```
wixoaGit, line 90: There's no need to turn the sizes into arguments. You could keep a list of all the expressions, push them onto the stack, and give the `CreateMultidimensionalList` opcode an immediate value describing how many dimensions (a.k.a expressions pushed onto the stack) there are.

wixoaGit (changes requested):

DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -313,26 +315,45 @@ public DMASTStatement Statement(bool requireDelimiter = true)
...
                     var elements = path.Elements.ToList();
                     elements.Insert(elements.IndexOf("var") + 1, "list");
                     path = new DreamPath("/" + String.Join("/", elements));
                 }
 
-                Whitespace();
-                DMASTExpression size = Expression();
-                ConsumeRightBracket();
-                Whitespace();
+                List<DMASTCallParameter> sizes = new List<DMASTCallParameter>(2); // Most common is 1D or 2D lists
```
wixoaGit, line 329: This should be converted to a `List<DMASTExpression>`

DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -313,26 +315,45 @@ public DMASTStatement Statement(bool requireDelimiter = true)
...
+                bool setSize = false;
```
wixoaGit, line 330: This can be removed and replaced with `sizes.Count > 0`

DMCompiler/DM/DMProc.cs
```diff
@@ -571,6 +571,12 @@ public void Pop()
             WriteOpcode(DreamProcOpcode.DeleteObject);
         }
 
+        public void CreateMultidimensionalList(int count) {
+            GrowStack(1);
```
wixoaGit, line 575: This pops `count` values off the stack and pushes 1
```suggestion
            ShrinkStack(count - 1);
```

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -128,6 +128,32 @@ static class DMOpcodeHandlers {
...
+            }
+            List<int> sizes = new List<int>(count);
+            foreach (var dv in values)
+            {
+                if (!dv.TryGetValueAsInteger(out var size))
+                {
+                    throw new Exception("Invalid list size");
+                }
+                sizes.Add(size);
+            }
```
wixoaGit, line 148: There's no need for two loops here. Non-integer sizes are also accepted and are treated as zero.
```suggestion
            List<int> sizes = new List<int>(count);
            for (var i = 0; i < count; i++)
            {
                state.Pop().TryGetValueAsInteger(out var size);
                sizes.Add(size);
            }
```

wixoaGit (changes requested):

DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -313,26 +315,43 @@ public DMASTStatement Statement(bool requireDelimiter = true)
...
-                if (size is not null) {
-                    implied_value = new DMASTNewPath(loc, new DMASTPath(loc, DreamPath.List),
-                        new[] { new DMASTCallParameter(loc, size) });
+                    ConsumeRightBracket();
+                    Whitespace();
                 }
+
+                if (sizes.Count > 0)
+                {
+                    DMASTExpression[] expressions = new DMASTExpression[sizes.Count];
```
ike709, line 347:
```suggestion
                    DMASTExpression[] expressions = sizes.ToArray()
```

DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -313,26 +315,43 @@ public DMASTStatement Statement(bool requireDelimiter = true)
...
+                    Whitespace();
                 }
+
+                if (sizes.Count > 0)
+                {
+                    DMASTExpression[] expressions = sizes.ToArray()
+                    for (var i = 0; i < sizes.Count; i++)
+                    {
+                        expressions[i] = sizes[i];
+                    }
```
ike709, line 351:
```suggestion
```

wixoaGit (approved):

DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -313,26 +315,39 @@ public DMASTStatement Statement(bool requireDelimiter = true)
...
+                    ConsumeRightBracket();
+                    Whitespace();
+                }
 
-                if (size is not null) {
-                    implied_value = new DMASTNewPath(loc, new DMASTPath(loc, DreamPath.List),
-                        new[] { new DMASTCallParameter(loc, size) });
+                if (sizes.Count > 0)
+                {
+                    DMASTExpression[] expressions = sizes.ToArray()
```
ike709, line 347:
```suggestion
                    DMASTExpression[] expressions = sizes.ToArray();
```

PR #660, opened 2022-04-05
Misc cleanup + allow `null` in typepaths

ike709:
Shaves off ~100 MB and 1-2 seconds from Paradise compilation.

### DIFF
--- a/DMCompiler/Compiler/DM/DMParser.cs
+++ b/DMCompiler/Compiler/DM/DMParser.cs
@@ -95,7 +95,8 @@ namespace DMCompiler.Compiler.DM {
             TokenType.DM_Var,
             TokenType.DM_Proc,
             TokenType.DM_Step,
-            TokenType.DM_Throw
+            TokenType.DM_Throw,
+            TokenType.DM_Null
         };
 
         public DMASTFile File() {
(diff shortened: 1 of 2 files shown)


DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -88,6 +88,16 @@ public partial class DMParser : Parser<Token> {
...
         };
 
+        private static readonly TokenType[] IdentifierTypes = {TokenType.DM_Identifier, TokenType.DM_Step};
+
+        private static readonly TokenType[]  ValidPathElementTokens = {
+            TokenType.DM_Identifier,
+            TokenType.DM_Var,
+            TokenType.DM_Proc,
+            TokenType.DM_Step,
+            TokenType.DM_Throw
```
ike709, line 98:
```suggestion
            TokenType.DM_Throw,
            TokenType.DM_Null
```

ike709:
Added `null` to the typepath tokens, since apparently it counts as one in BYOND.

DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -1951,120 +1956,140 @@ public DMASTProcStatementLabel Label(DMASTIdentifier expression)
...
-                        } else if (bracketNesting == 0) {
-                            stringBuilder.Append(c);
                         }
                     }
 
                     if (bracketNesting > 0) Error("Expected ']'");
 
                     string stringValue = stringBuilder.ToString();
-                    if (interpolationValues.Count == 0) {
+                    if (interpolationValues is null || interpolationValues.Count == 0) {
```
wixoaGit, line 2092: As long as `interpolationValues` isn't null it should have at least 1 value

```suggestion
                    if (interpolationValues is null) {
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #662, opened 2022-04-07
Two small fixes

ike709:
1. Fixes whitespace between the first proccall param and first comma.
2. Adds the unicode escape chars.

### DIFF
--- a/DMCompiler/Compiler/DM/DMParser.cs
+++ b/DMCompiler/Compiler/DM/DMParser.cs
@@ -1362,6 +1362,7 @@ namespace DMCompiler.Compiler.DM {
         public DMASTCallParameter[] CallParameters() {
             List<DMASTCallParameter> parameters = new();
             DMASTCallParameter parameter = CallParameter();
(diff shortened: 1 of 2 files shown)


DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -1362,6 +1362,7 @@ public DMASTProcStatementLabel Label(DMASTIdentifier expression)
         public DMASTCallParameter[] CallParameters() {
             List<DMASTCallParameter> parameters = new();
             DMASTCallParameter parameter = CallParameter();
+            BracketWhitespace();
```
wixoaGit, line 1365: This will only check for whitespace before the first comma. There needs to be another `BracketWhitespace()` call at the end of the while loop below for the following code to work

```
proc_call(1
    ,2
    ,3
    ,4)
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #665, opened 2022-04-21
Fix proc definitions not passing current obj properly

ike709:
I don't think there is an issue for this but it does fix a bug with paradise.

### DIFF
--- a/DMCompiler/DM/Visitors/DMObjectBuilder.cs
+++ b/DMCompiler/DM/Visitors/DMObjectBuilder.cs
@@ -147,7 +147,7 @@ namespace DMCompiler.DM.Visitors {
 
                             if (varDeclaration.Value != null) {
                                 DMVisitorExpression._scopeMode = "static";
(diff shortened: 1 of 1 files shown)


DMCompiler/DM/Visitors/DMObjectBuilder.cs
```diff
@@ -147,7 +147,7 @@ class DMObjectBuilder {
 
                             if (varDeclaration.Value != null) {
                                 DMVisitorExpression._scopeMode = "static";
-                                DMExpression expression = DMExpression.Create(_currentObject, DMObjectTree.GlobalInitProc, varDeclaration.Value, varDeclaration.Type);
+                                DMExpression expression = DMExpression.Create(dmObject, DMObjectTree.GlobalInitProc, varDeclaration.Value, varDeclaration.Type);
```
ike709, line 150:
```suggestion
                                DMExpression expression = DMExpression.Create(dmObject, null, varDeclaration.Value, varDeclaration.Type);
```

wixoaGit (approved):

PR #667, opened 2022-04-21
Fixes dereferencing a ternary

### LINKED ISSUE #528: Ternary dereference issue
Code: `if(!(container ? container : src).Adjacent(user))`
Error: `Error at code\modules\hydroponics\seeds.dm:343:16: Invalid property "Adjacent"`

ike709:
Closes #528


wixoaGit:
This doesn't seem correct to me. If a ternary was typed, surely it'd be the b or c expression's type?

Even then, I don't think ternaries are typed, and don't have their dereferences checked.

PR #668, opened 2022-04-21
Adds `Location.Internal`

ike709:
IMO it's a good idea to differentiate between explicit warnings/errors raised internally by OD and actual unknown loc warnings/errors. There's probably more places that should be changed but the unimplemented warning is the most apparent.

Ex:
```
Compiling paradise.dme
Warning at <internal>: Unimplemented proc & var warnings are currently suppressed
```

### DIFF
--- a/OpenDreamShared/Compiler/Location.cs
+++ b/OpenDreamShared/Compiler/Location.cs
@@ -4,7 +4,14 @@ namespace OpenDreamShared.Compiler
 {
     public readonly struct Location
     {
+        /// <summary>
+        /// For when DM location information can't be determined.
+        /// </summary>
         public static readonly Location Unknown = new Location();
+        /// <summary>
+        /// For when internal OpenDream warnings/errors are raised.
+        /// </summary>
+        public static readonly Location Internal = new Location("<internal>", null, null);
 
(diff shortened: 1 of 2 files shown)


OpenDreamShared/Compiler/Location.cs
```diff
@@ -4,7 +4,14 @@ namespace OpenDreamShared.Compiler
 {
     public readonly struct Location
     {
+        /// <summary>
+        /// For when DM location information can't be determined.
+        /// </summary>
         public static readonly Location Unknown = new Location();
+        /// <summary>
+        /// For when internal OpenDream warnings/errors are raised.
```
wixoaGit, line 12: This comment should be extended to include anything internal that needs a location, instead of just warnings/errors. An example of this would be the assignment expressions used in init procs, which currently use `Location.Unknown`.

wixoaGit (changes requested):

OpenDreamShared/Compiler/Location.cs
ike709, line 12:
```suggestion
        /// For when internal OpenDream warnings/errors are raised or something internal needs to be passed a location.
```

wixoaGit (approved):

PR #672, opened 2022-05-01
Moves procs to a single global list/array

ike709:
First step to nuking `DreamPath`. Changes to `call()()` are getting punted to a follow-up PR.

Probably has broken/bad things in here but TestGame seems to work.


Hinaichigo:
Would this fix certain procs not being recognized even though they are declared elsewhere?

wixoaGit:
> Would this fix certain procs not being recognized even though they are declared elsewhere?

No, this just changes how existing procs are referenced internally. Could you open an issue for this?

DMCompiler/DM/DMProc.cs
```diff
@@ -49,6 +49,8 @@ private class DMProcScope {
         public Location Location = Location.Unknown;
         public ProcAttributes Attributes;
         public string Name { get => _astDefinition?.Name; }
+        public int Id;
+        private DMObject _dmObject;
```
wixoaGit, line 53: Move this down with the rest of the private properties.

DMCompiler/DM/Visitors/DMObjectBuilder.cs
```diff
@@ -13,14 +13,14 @@ class DMObjectBuilder {
             DMObjectTree.Reset();
             ProcessFile(astFile);
 
+
+            // TODO Nuke this pass
```
wixoaGit, line 17: Extra line here
```suggestion
            ProcessFile(astFile);

            // TODO Nuke this pass
```

wixoaGit (changes requested):

DMCompiler/DM/DMObject.cs
```diff
@@ -139,24 +128,19 @@ class DMObject {
...
-                    }
-
-                    typeJson.Procs.Add(procs.Key, procJson);
+            if (Procs.Count > 0)
+            {
+                typeJson.Procs = new List<List<int>>(Procs.Count);
+                foreach (var (_, value) in Procs)
+                {
+                    typeJson.Procs.Add(value);
                 }
```
wixoaGit, line 142:
```suggestion
                typeJson.Procs = new List<List<int>>(Procs.Values);
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #673, opened 2022-05-03
Re-adds unit tests

ike709:
I commented out all of the broken ones. Getting the test framework back in is more important that fixing every test, we can address those later.


Content.Tests/SetupCompileDM.cs
```diff
@@ -0,0 +1,35 @@
+using System;
+using System.IO;
+using NUnit.Framework;
+using NUnit.Framework.Internal;
+
+// This it outside of any namespace so it affects the whole assembly.
```
wixoaGit, line 6:
```suggestion
// This is outside of any namespace so it affects the whole assembly.
```

OpenDreamRuntime/EntryPoint.cs
```diff
@@ -40,7 +40,7 @@ public class EntryPoint : GameServer {
 
         public override void PostInit() {
             _commandSystem = EntitySystem.Get<DreamCommandSystem>();
-            _dreamManager.Initialize();
+            _dreamManager.Initialize(null);
```
wixoaGit, line 43: Have this call pass the cvar instead of having multiple locations default to `testingJson ?? _configManager.GetCVar<string>(OpenDreamCVars.JsonPath)`

wixoaGit (changes requested):

OpenDreamRuntime/DreamManager.cs
```diff
@@ -39,16 +39,17 @@ partial class DreamManager : IDreamManager {
         public List<DreamObject> Datums { get; set; } = new();
         public Random Random { get; set; } = new();
 
-        public void Initialize() {
+        //TODO This arg is awful and temporary until RT supports cvar overrides in unit tests
+        public void Initialize(string? testingJson) {
```
wixoaGit, line 43: This is more than just the testing json now, and shouldn't be null.
```suggestion
        public void Initialize(string jsonPath) {
```

OpenDreamRuntime/DreamManager.cs
```diff
@@ -39,16 +39,17 @@ partial class DreamManager : IDreamManager {
...
             InitializeConnectionManager();
 
-            DreamCompiledJson json = LoadJson();
+            DreamCompiledJson json = LoadJson(testingJson);
```
wixoaGit, line 46:
```suggestion
            DreamCompiledJson json = LoadJson(jsonPath);
```

OpenDreamRuntime/DreamManager.cs
```diff
@@ -39,16 +39,17 @@ partial class DreamManager : IDreamManager {
...
             if (json == null)
                 return;
 
             _compiledJson = json;
 
-            _dreamResourceManager.Initialize();
+            _dreamResourceManager.Initialize(testingJson);
```
wixoaGit, line 52:
```suggestion
            _dreamResourceManager.Initialize(jsonPath);
```

OpenDreamRuntime/DreamManager.cs
```diff
@@ -83,7 +84,11 @@ partial class DreamManager : IDreamManager {
                 globalInitProc.Spawn(WorldInstance, new DreamProcArguments(new(), new()));
             }
 
-            _dreamMapManager.LoadMaps(json.Maps);
+            //TODO figure out why this json doesn't work
+            if (testingJson is null)
+            {
+                _dreamMapManager.LoadMaps(json.Maps);
+            }
```
wixoaGit, line 91: Is this still an issue?

wixoaGit (changes requested):

OpenDreamRuntime/DreamManager.cs
ike709: No

wixoaGit (approved):

PR #675, opened 2022-05-04
Hardens the loop stack against exceptions, reduces allocs

ike709:
1. Loop stacks only allocate when needed, and set their capacity to 3 initially (start, continue, end)
2. Loop stacks will now throw a compiler error instead of an exception if you pop them while they're null. I'm not sure what in baycode is causing this but the compiler should *never* really throw exceptions.

### DIFF
--- a/DMCompiler/DM/DMProc.cs
+++ b/DMCompiler/DM/DMProc.cs
@@ -323,7 +323,7 @@ namespace DMCompiler.DM {
             }
             else
             {
-                DMCompiler.Error(new CompilerError(Location.Internal, "Cannot pop empty loop stack"));
+                DMCompiler.Error(new CompilerError(Location, "Cannot pop empty loop stack"));
             }
 
             EndScope();
--- a/DMCompiler/DM/DMProc.cs
+++ b/DMCompiler/DM/DMProc.cs
@@ -373,7 +373,7 @@ namespace DMCompiler.DM {
             }
             else
             {
-                DMCompiler.Error(new CompilerError(Location.Internal, "Cannot peek empty loop stack"));
+                DMCompiler.Error(new CompilerError(Location, "Cannot peek empty loop stack"));
             }
         }
 
--- a/DMCompiler/DM/DMProc.cs
+++ b/DMCompiler/DM/DMProc.cs
@@ -55,7 +55,7 @@ namespace DMCompiler.DM {
         private BinaryWriter _bytecodeWriter = null;
         private Dictionary<string, long> _labels = new();
         private List<(long Position, string LabelName)> _unresolvedLabels = new();
-        private Stack<string> _loopStack = new();
+        [CanBeNull] private Stack<string> _loopStack = null;
         private Stack<DMProcScope> _scopes = new();
         private Dictionary<string, LocalVariable> _parameters = new();
         private int _localVariableIdCounter = 0;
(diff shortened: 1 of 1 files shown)


DMCompiler/DM/DMProc.cs
```diff
@@ -316,7 +317,15 @@ public void LoopJumpToStart(string loopLabel)
...
 
         public void LoopEnd() {
-            AddLabel(_loopStack.Pop() + "_end");
+            if (_loopStack?.TryPop(out var pop) ?? false)
+            {
+                AddLabel(pop + "_end");
+            }
+            else
+            {
+                DMCompiler.Error(new CompilerError(Location.Internal, "Cannot pop empty loop stack"));
```
wixoaGit, line 326: This should use the proc's location, so you can at least tell what proc is causing it.
```suggestion
                DMCompiler.Error(new CompilerError(Location, "Cannot pop empty loop stack"));
```

DMCompiler/DM/DMProc.cs
```diff
@@ -358,14 +367,25 @@ public void LoopJumpToStart(string loopLabel)
...
                 Jump(label.Identifier + "_end");
             }
+            else if (_loopStack?.TryPeek(out var peek) ?? false)
+            {
+                Jump(peek + "_end");
+            }
             else
             {
-                Jump(_loopStack.Peek() + "_end");
+                DMCompiler.Error(new CompilerError(Location.Internal, "Cannot peek empty loop stack"));
```
wixoaGit, line 376:
```suggestion
                DMCompiler.Error(new CompilerError(Location, "Cannot peek empty loop stack"));
```

DMCompiler/DM/DMProc.cs
```diff
@@ -358,14 +367,25 @@ public void LoopJumpToStart(string loopLabel)
...
 
         public void BreakIfFalse() {
-            JumpIfFalse(_loopStack.Peek() + "_end");
+            if (_loopStack?.TryPeek(out var peek) ?? false)
+            {
+                JumpIfFalse(peek + "_end");
+            }
+            else
+            {
+                DMCompiler.Error(new CompilerError(Location.Internal, "Cannot peek empty loop stack"));
```
wixoaGit, line 387:
```suggestion
                DMCompiler.Error(new CompilerError(Location, "Cannot peek empty loop stack"));
```

DMCompiler/DM/DMProc.cs
```diff
@@ -393,12 +413,26 @@ public void LoopJumpToStart(string loopLabel)
...
             {
                 BackgroundSleep();
-                Jump(_loopStack.Peek() + "_continue");
+                if (_loopStack?.TryPeek(out var peek) ?? false)
+                {
+                    Jump(peek + "_continue");
+                }
+                else
+                {
+                    DMCompiler.Error(new CompilerError(Location.Internal, "Cannot peek empty loop stack"));
```
wixoaGit, line 422:
```suggestion
                    DMCompiler.Error(new CompilerError(Location, "Cannot peek empty loop stack"));
```

DMCompiler/DM/DMProc.cs
```diff
@@ -393,12 +413,26 @@ public void LoopJumpToStart(string loopLabel)
...
 
         public void ContinueIfFalse() {
-            JumpIfFalse(_loopStack.Peek() + "_continue");
+            if (_loopStack?.TryPeek(out var peek) ?? false)
+            {
+                Jump(peek + "_continue");
+            }
+            else
+            {
+                DMCompiler.Error(new CompilerError(Location.Internal, "Cannot peek empty loop stack"));
```
wixoaGit, line 434:
```suggestion
                DMCompiler.Error(new CompilerError(Location, "Cannot peek empty loop stack"));
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #676, opened 2022-05-05
Engine update

ike709:
Updates Robust Toolbox.

Seems to work mostly okay except right now the mob in TestGame is invisible. And there's a major sandbox issue.


ike709:
Player now visible.

DMCompiler/DM/DMProc.cs
```diff
@@ -826,15 +826,19 @@ public void Pop()
...
-                case DMReference.Type.Argument: WriteByte(reference.ArgumentId); break;
-                case DMReference.Type.Local: WriteByte(reference.LocalId); break;
-                case DMReference.Type.Global: WriteInt(reference.GlobalId); break;
-                case DMReference.Type.Field: WriteString(reference.FieldName); ShrinkStack(affectStack ? 1 : 0); break;
-                case DMReference.Type.SrcField: WriteString(reference.FieldName); break;
-                case DMReference.Type.Proc: WriteString(reference.ProcName); ShrinkStack(affectStack ? 1 : 0); break;
-                case DMReference.Type.GlobalProc: WriteString(reference.ProcName); break;
-                case DMReference.Type.SrcProc: WriteString(reference.ProcName); break;
+                case DMReference.Type.Argument:
+                case DMReference.Type.Local:
```
wixoaGit, line 830: Arguments and Locals have a max of 255 and can be written as bytes rather than integers. Writing them as integers appears to make the output of stripped /tg/ ~156 KB larger and TestGame ~3 KB larger.

It would be a good idea to rename `DMReference.Int` to `DMReference.Index` to more accurately reflect what it is.
```suggestion
                case DMReference.Type.Argument:
                case DMReference.Type.Local: WriteByte((byte)reference.Int); break;
                
```

OpenDreamRuntime/Procs/DMProc.cs
```diff
@@ -308,8 +308,8 @@ public void ReturnPools()
             DMReference.Type refType = (DMReference.Type)ReadByte();
 
             switch (refType) {
-                case DMReference.Type.Argument: return DMReference.CreateArgument(ReadByte());
-                case DMReference.Type.Local: return DMReference.CreateLocal(ReadByte());
+                case DMReference.Type.Argument: return DMReference.CreateArgument(ReadInt());
+                case DMReference.Type.Local: return DMReference.CreateLocal(ReadInt());
```
wixoaGit, line 312:
```suggestion
                case DMReference.Type.Argument: return DMReference.CreateArgument(ReadByte());
                case DMReference.Type.Local: return DMReference.CreateLocal(ReadByte());
```

wixoaGit (changes requested):

OpenDreamClient/Program.cs
```diff
@@ -4,7 +4,7 @@ namespace OpenDreamClient {
     internal static class Program {
         public static void Main(string[] args) {
             ContentStart.StartLibrary(args, new GameControllerOptions() {
-                Sandboxing = true,
+                Sandboxing = false,
```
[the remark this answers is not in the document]
ike709:
```suggestion
                Sandboxing = true,
```
Shouldn't be.

wixoaGit (approved):

PR #678, opened 2022-05-09
Cap the max compiler errors, cope with exceptions in parameters

ike709:
1. Caps the max compilation errors to 200. Then the compilation will abort similar to BYOND's 100 error limit.
2. Wrap params in a try/catch similar to `ProcBlockInner()`'s


DMCompiler/DMCompiler.cs
```diff
@@ -150,6 +157,12 @@ public static class DMCompiler {
         public static void Error(CompilerError error) {
             Console.WriteLine(error);
             ErrorCount++;
+            if (ErrorCount >= 200)
+            {
+                Console.WriteLine("Maximum compilation errors exceeded, aborting.");
+                EndCompilation(false);
+                Environment.Exit(1);
```
wixoaGit, line 164: `Environment.Exit()` isn't ideal as it leads to control never being returned to the code that called `DMCompiler.Compile()`

wixoaGit (changes requested):

PR #679, opened 2022-05-09
Fixes misc baystation issues

ike709:
1. Adds `switch` and `spawn` to the list of tokens you can use in a typepath
2. Copes with NOP switch cases like
```
if("No")
    ; // do nothing
```
3. Fixes an issue with modified types

### DIFF
--- a/DMCompiler/Compiler/DM/DMParser.cs
+++ b/DMCompiler/Compiler/DM/DMParser.cs
@@ -96,7 +96,8 @@ namespace DMCompiler.Compiler.DM {
             TokenType.DM_Proc,
             TokenType.DM_Step,
             TokenType.DM_Throw,
-            TokenType.DM_Null
+            TokenType.DM_Null,
+            TokenType.DM_Switch
         };
 
         public DMASTFile File() {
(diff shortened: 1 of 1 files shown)


DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -479,7 +481,13 @@ public DMASTStatement Statement(bool requireDelimiter = true)
...
                         procStatements.Add(statement);
                     } else {
-                        if (procStatements.Count == 0) return null;
+                        if (procStatements.Count == 0)
+                        {
+                            // Sometimes ';' gets used in NOP blocks
+                            Delimiter();
+                            Whitespace();
+                            return null;
+                        }
```
wixoaGit, line 490: This isn't the correct way to go about this, since this code should be valid:
```
switch (1)
    if (1)
        ;
        usr << "One";
```
It would be better to parse every empty statement in the block and not return early if the first one is empty. You can do that by removing this entire else block and moving the original `if (procStatements.Count == 0) return null` line above the `return procStatements` one.

wixoaGit (changes requested):

wixoaGit (approved):

PR #680, opened 2022-05-10
Cope with emitting null expressions

ike709:
I'm not entirely sure why this is happening in Baystation, but it's nicer to error with proc location info than just throw an exception.

### DIFF
--- a/DMCompiler/DM/DMExpression.cs
+++ b/DMCompiler/DM/DMExpression.cs
@@ -34,13 +34,21 @@ namespace DMCompiler.DM {
 
         public static DMExpression Create(DMObject dmObject, DMProc proc, DMASTExpression expression, DreamPath? inferredPath = null) {
             var instance = new DMVisitorExpression(dmObject, proc, inferredPath);
-            expression.Visit(instance);
+            if (expression is null)
+            {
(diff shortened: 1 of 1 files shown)


DMCompiler/DM/DMExpression.cs
```diff
@@ -34,13 +34,21 @@ public enum IdentifierPushResult {
...
+            if (expression is null)
+            {
+                DMCompiler.Error(new CompilerError(proc.Location, $"Attempted to visit null expression in proc {dmObject.Path}.{proc.Name}()"));
+            }
+            else
+            {
+                expression.Visit(instance);
+            }
+
             return instance.Result;
```
wixoaGit, line 46: This ends up returning the result of the last successful visit

DMCompiler/DM/DMExpression.cs
```diff
@@ -34,13 +34,21 @@ public enum IdentifierPushResult {
...
         }
 
         public static void Emit(DMObject dmObject, DMProc proc, DMASTExpression expression, DreamPath? inferredPath = null) {
             var expr = Create(dmObject, proc, expression, inferredPath);
-            expr.EmitPushValue(dmObject, proc);
+            expr?.EmitPushValue(dmObject, proc); // This shouldn't be null but it can be
```
wixoaGit, line 51: This should also error if it is null

wixoaGit (changes requested):

PR #688, opened 2022-05-16
Fix `pick(arglist(args))`

ike709:
```
/proc/_pick(...)
	return pick(arglist(args))
```

### DIFF
--- a/DMCompiler/DM/Expressions/Builtins.cs
+++ b/DMCompiler/DM/Expressions/Builtins.cs
@@ -193,8 +193,16 @@ namespace DMCompiler.DM.Expressions {
 
                 proc.PickWeighted(_values.Length);
             } else {
-                foreach (PickValue pickValue in _values) {
-                    pickValue.Value.EmitPushValue(dmObject, proc);
+                foreach (PickValue pickValue in _values)
+                {
+                    if (pickValue.Value is Arglist args)
+                    {
+                        args.EmitPushArglist(dmObject, proc);
+                    }
+                    else
+                    {
+                        pickValue.Value.EmitPushValue(dmObject, proc);
+                    }
                 }
 
                 proc.PickUnweighted(_values.Length);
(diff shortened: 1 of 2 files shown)


wixoaGit:
Odd that it's written that way, it's the same thing as `pick(args)`

wixoaGit:
Odd that it's written that way, it's the same thing as `pick(args)`

DMCompiler/DM/Expressions/Builtins.cs
```diff
@@ -193,8 +193,16 @@ public struct PickValue {
 
                 proc.PickWeighted(_values.Length);
             } else {
-                foreach (PickValue pickValue in _values) {
-                    pickValue.Value.EmitPushValue(dmObject, proc);
+                foreach (PickValue pickValue in _values)
+                {
+                    if (pickValue.Value is Arglist args)
+                    {
+                        args.EmitPushArglist(dmObject, proc);
```
wixoaGit, line 200: The pick opcodes don't expect an arglist

wixoaGit (changes requested):

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -1524,7 +1524,23 @@ static class DMOpcodeHandlers {
...
+                        var valList = args.NamedArguments.Keys.ToArray();
+                        for(var i = 0; i < values.Length; i++)
+                        {
+                            values[i] = new DreamValue(valList[i]);
+                        }
+                    }
+                    else
+                    {
+                        values = args.OrderedArguments.ToArray();
+                    }
```
wixoaGit, line 1541: `DreamProcArguments` has a method for this already
```suggestion
                    values = args.GetAllArguments().ToArray();
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #695, opened 2022-05-16
Cope with whitespace in ExpressionPrimary

ike709:
Paradise:
```
/proc/is_surgery_tool(obj/item/W as obj)
	return (	\
	istype(W, /obj/item/scalpel)			||	\
	istype(W, /obj/item/hemostat)		||	\
	istype(W, /obj/item/retractor)		||	\
	istype(W, /obj/item/cautery)			||	\
	istype(W, /obj/item/bonegel)			||	\
	istype(W, /obj/item/bonesetter)
	)
```

### DIFF
--- a/DMCompiler/Compiler/DM/DMParser.cs
+++ b/DMCompiler/Compiler/DM/DMParser.cs
@@ -1891,10 +1891,9 @@ namespace DMCompiler.Compiler.DM {
         public DMASTExpression ExpressionPrimary(bool allowParentheses = true) {
             if (allowParentheses && Check(TokenType.DM_LeftParenthesis))
             {
-                Whitespace();
+                BracketWhitespace();
                 DMASTExpression inner = Expression();
-                Newline();
-                Whitespace(includeIndentation: true);
+                BracketWhitespace();
                 ConsumeRightParenthesis();
 
                 return inner;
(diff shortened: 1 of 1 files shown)


Hinaichigo:
What does the backslash do here?

ZeWaka:
escapes the newline

DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -1889,9 +1889,12 @@ public DMASTProcStatementLabel Label(DMASTIdentifier expression)
         }
 
         public DMASTExpression ExpressionPrimary(bool allowParentheses = true) {
-            if (allowParentheses && Check(TokenType.DM_LeftParenthesis)) {
+            if (allowParentheses && Check(TokenType.DM_LeftParenthesis))
+            {
                 Whitespace();
                 DMASTExpression inner = Expression();
+                Newline();
+                Whitespace(includeIndentation: true);
```
wixoaGit, line 1897: `BracketWhitespace()` is meant to be used for whitespace inside brackets/parentheses which allows delimiters. The first `Whitespace()` here should be changed too.
```suggestion
                BracketWhitespace();
                DMASTExpression inner = Expression();
                BracketWhitespace();
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #701, opened 2022-05-23
Declare `type` on `/list`

ike709:
image


DMCompiler/DMStandard/Types/List.dm
```diff
@@ -1,5 +1,6 @@
 ﻿/list
 	var/len
+	var/type = /list
```
wixoaGit, line 3: This should be made read-only

wixoaGit (changes requested):

wixoaGit (approved):

PR #703, opened 2022-05-25
Partially automate unit tests

ike709:
How to create a unit test with the new system:

1. Create the DM file for the test somewhere in `Content.Tests/DMProject/Tests`
2. Include the file in the DME. In the future we should just auto-generate the list of files to compile but I couldn't be bothered to set that up.
3. If the test is supposed to runtime, put `// RUNTIME TRUE` on the first line. The only whitespace that matters is having a space between `RUNTIME` and `TRUE`
4. Declare the test's proc to call like so: `/world/proc/[filename]_Proc()` without the file extension.
5. Voila

I didn't move every single existing test over because I didn't want to rewrite all of them, but this makes it easier to add more tests.


Content.Tests/DMTests.cs
```diff
@@ -28,6 +30,40 @@ public void OneTimeSetup()
...
+            string firstLine;
+
+            using(StreamReader reader = new StreamReader(file))
+            {
+                firstLine = reader.ReadLine() ?? "";
+            }
+
+            bool shouldRuntime = firstLine.Contains("RUNTIME TRUE", StringComparison.InvariantCulture);
+
+            var fileName = Path.GetFileName(file)[..^3]; // Ignore the ".dm" extension
```
wixoaGit, line 47:
```suggestion
            var fileName = Path.GetFileNameWithoutExtension(file); // Ignore the ".dm" extension
```

Content.Tests/DMTests.cs
```diff
@@ -28,6 +30,40 @@ public void OneTimeSetup()
...
+
+            var result = DreamThread.Run(async(state) => {
+                var world = _dreamMan.WorldInstance;
+                var proc = world.GetProc($"{fileName}_Proc");
+                return await state.Call(proc, world, null, new DreamProcArguments(null));
```
wixoaGit, line 52: This should check if the proc exists first.
```suggestion
                
                if (world.TryGetProc($"{fileName}_Proc", out DreamProc proc)) {
                    return await state.Call(proc, world, null, new DreamProcArguments(null));
                } else {
                    Assert.Fail($"No proc named {fileName}_Proc");
                    return DreamValue.Null;
                }
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #711, opened 2022-06-01
Render perf optimisations

metalgearsloth:
Not sure how much value it is at this stage.

### DIFF
--- a/OpenDreamClient/Rendering/DMISpriteComponent.cs
+++ b/OpenDreamClient/Rendering/DMISpriteComponent.cs
@@ -35,7 +35,7 @@ namespace OpenDreamClient.Rendering {
             return Icon.GetWorldAABB(transform.Position);
         }
 
-        public bool IsVisible(bool checkWorld = true) {
+        public bool IsVisible(bool checkWorld = true, [CanBeNull] IMapManager mapManager = null) {
             if (Icon?.DMI == null) return false;
             if (Icon.Appearance.Invisibility > 0) return false; //TODO: mob.see_invisibility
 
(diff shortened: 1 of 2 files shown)


OpenDreamClient/Rendering/DMISpriteComponent.cs
```diff
@@ -45,7 +45,8 @@ sealed class DMISpriteComponent : SharedDMISpriteComponent, ILookupWorldBox2Comp
                 if (!_entityManager.TryGetComponent<TransformComponent>(Owner, out var transform))
                     return false;
 
-                EntityUid mapEntity = IoCManager.Resolve<IMapManager>().GetMapEntityId(transform.MapID);
+                IoCManager.Resolve(ref mapManager);
```
metalgearsloth, line 48: IoC dependency collections are backed by ThreadLocals which add up if you have to resolve it heaps, in ss14 we just add nullable arguments as an easy way around it if the class isn't ECS.

OpenDreamClient/Rendering/DreamViewOverlay.cs
```diff
@@ -1,57 +1,58 @@
...
-        private void DrawMap(DrawingHandleWorld handle, EntityUid eye) {
-
+        private void DrawMap(OverlayDrawArgs args, EntityUid eye) {
 
+            _transformSystem ??= _entitySystem.GetEntitySystem<SharedTransformSystem>();
             _lookupSystem ??= _entitySystem.GetEntitySystem<EntityLookupSystem>();
+            var spriteQuery = _entityManager.GetEntityQuery<DMISpriteComponent>();
+            var xformQuery = _entityManager.GetEntityQuery<TransformComponent>();
 
             var entities = _lookupSystem.GetEntitiesInRange(eye, 15);
```
metalgearsloth, line 35: If the eye is attached to the player entity then you could just use the OverlayDrawArgs Mapid + WorldAABB instead?

wixoaGit: That does appear to be a better way to do it, wasn't aware of it

wixoaGit (approved):

OpenDreamClient/Rendering/DreamViewOverlay.cs
metalgearsloth: Either MapID or WorldAABB were just added in the most recent engine update I think so probably not available sooner.

PR #712, opened 2022-06-09
Meta Object Reparenting

### LINKED ISSUE #312: Can't set client's `parent_type` to `/datum`
TG and forks do it.

ike709:
Closes #312 

Right now this doesn't have any safeguards to prevent you from reparenting in a way that'll horrifically break everything, but that should be fine for now and I'll make an issue for it when this is merged.


OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectRoot.cs
```diff
@@ -1,10 +1,17 @@
-﻿using OpenDreamRuntime.Procs;
+using OpenDreamRuntime.Procs;
+using OpenDreamShared.Dream;
 
 namespace OpenDreamRuntime.Objects.MetaObjects {
     [Virtual]
     class DreamMetaObjectRoot : IDreamMetaObject {
```
wixoaGit, line 6: I don't see a point in there being a `DreamMetaObjectRoot` anymore. You could have every meta object implement `IDreamMetaObject` directly instead.

wixoaGit (changes requested):

OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectRoot.cs
ike709: Because it reduces code duplication

Issue #141, opened 2021-06-04
`/icon` procs are not implemented


wixoaGit:
Procs in the `/icon` object are missing implementations.

- [ ] Blend
- [ ] Crop
- [ ] DrawBox
- [ ] Flip
- [ ] GetPixel
- [ ] Height
- [ ] IconStates
- [ ] Insert
- [ ] MapColors
- [ ] Scale
- [ ] SetIntensity
- [ ] Shift
- [ ] SwapColor
- [ ] Turn
- [ ] Width

wixoaGit:
OpenSS13 makes use of `Blend` and `DrawBox`

ike709:
I should also mention that you can't currently set a var to a `new /icon()

ike709:
`Blend()` partially done in https://github.com/OpenDreamProject/OpenDream/pull/948

ike709:
`Scale()` completed in https://github.com/OpenDreamProject/OpenDream/pull/951

itsmeow:
Scale() doesn't properly transform individual frames within the icon, instead just scaling the entire icon arbitrarily, without changing the images inside.

amylizzle:
Suggest editing this issue to just be about `Blend()`'s modes, since proc tracking is now in #2194

tontyGH:
Resolved by #2590?

wixoaGit:
Yep

PR #714, opened 2022-06-09
Implements icon.Insert()

ike709:
Ticks a box on #141 

This is one of the most painful native procs I've ever written. I do not know if it actually works. Yes, it could probably be written better.


OpenDreamRuntime/Procs/Native/DreamProcNativeIcon.cs
```diff
@@ -17,6 +18,41 @@ static class DreamProcNativeIcon {
             return new DreamValue(dreamIconObject.Description.Height);
         }
 
+        [DreamProc("Insert")]
+        [DreamProcParameter("new_icon", Type = DreamValue.DreamValueType.DreamObject)]
+        [DreamProcParameter("icon_state", Type = DreamValue.DreamValueType.String)]
+        [DreamProcParameter("dir", Type = DreamValue.DreamValueType.Float)]
+        [DreamProcParameter("frame", Type = DreamValue.DreamValueType.Float)]
+        [DreamProcParameter("moving", Type = DreamValue.DreamValueType.Float)] // do by hand
```
ike709, line 26:
```suggestion
        [DreamProcParameter("moving", Type = DreamValue.DreamValueType.Float)]
```

Altoids1 (approved):
Looks good to me.

OpenDreamRuntime/Procs/Native/DreamProcNativeIcon.cs
```diff
@@ -17,6 +18,41 @@ static class DreamProcNativeIcon {
...
+            {
+                delay = delayNum;
+            }
+            
+            DreamMetaObjectIcon.DreamIconObject instanceIconObject = DreamMetaObjectIcon.ObjectToDreamIcon[instance];
+            DreamMetaObjectIcon.DreamIconObject newIconObject = DreamMetaObjectIcon.ObjectToDreamIcon[new_icon];
+            instanceIconObject.Description.InsertIcon(newIconObject.Description, icon_state, dir, frame, delay);
+
+            instanceIconObject.Moving = moving ? DreamMetaObjectIcon.DreamIconMovingMode.Movement : DreamMetaObjectIcon.DreamIconMovingMode.NonMovement;
+            return new DreamValue(instanceIconObject);
```
wixoaGit, line 55: This proc doesn't return anything
```suggestion
            return DreamValue.Null;
```

OpenDreamRuntime/Procs/Native/DreamProcNativeIcon.cs
```diff
@@ -17,6 +18,41 @@ static class DreamProcNativeIcon {
...
+        [DreamProcParameter("new_icon", Type = DreamValue.DreamValueType.DreamObject)]
+        [DreamProcParameter("icon_state", Type = DreamValue.DreamValueType.String)]
+        [DreamProcParameter("dir", Type = DreamValue.DreamValueType.Float)]
+        [DreamProcParameter("frame", Type = DreamValue.DreamValueType.Float)]
+        [DreamProcParameter("moving", Type = DreamValue.DreamValueType.Float)]
+        [DreamProcParameter("delay", Type = DreamValue.DreamValueType.Float)]
+        public static DreamValue NativeProc_Insert(DreamObject instance, DreamObject usr, DreamProcArguments arguments) {
+            //TODO Figure out what happens when you pass the wrong types as args
 
+            arguments.GetArgument(0, "new_icon").TryGetValueAsDreamObject(out var new_icon);
```
wixoaGit, line 31: The ref says this could also be an icon resource. This should either handle that, or throw an unimplemented exception.

OpenDreamShared/Resources/DMIParser.cs
```diff
@@ -44,6 +45,80 @@ public sealed class ParsedDMIDescription {
...
+            {
+                List<ParsedDMIState> states;
+                if (icon_state is null)
+                {
+                    states = icon.States.Values.ToList();
+                }
+                else
+                {
+                    states = new List<ParsedDMIState>(1) { icon.States[icon_state] };
+                }
```
wixoaGit, line 58: Avoid an unnecessary `ToList()`
```suggestion
                IEnumerable<ParsedDMIState> states;
                if (icon_state is null) {
                    states = icon.States.Values;
                } else {
                    states = new ParsedDMIState[] { icon.States[icon_state] };
                }
```

OpenDreamShared/Resources/DMIParser.cs
```diff
@@ -44,6 +45,80 @@ public sealed class ParsedDMIDescription {
...
+                                            ? stateDirs.First().Value.Last().Delay
+                                            : stateDirs[dir.Value].Last().Delay;
+                                    }
+                                }
+                            }
+                        }
+                    }
+                }
+
+                // All of that above was just to adjust the inserted icon to match the args. Now we can actually insert it.
```
wixoaGit, line 107: Wouldn't all of the above also modify the icon that was passed to `Insert()`? That wouldn't be desirable behavior.

OpenDreamShared/Resources/DMIParser.cs
```diff
@@ -44,6 +45,80 @@ public sealed class ParsedDMIDescription {
...
+                foreach (var state in states)
+                {
+                    if (States.ContainsKey(state.Name))
+                    {
+                        States[state.Name] = state;
+                    }
+                    else
+                    {
+                        States.Add(state.Name, state);
+                    }
```
wixoaGit, line 118: `States[state.Name] = ...` will create a key if it doesn't already exist
```suggestion
                    States[state.Name] = state;
```

OpenDreamShared/Resources/DMIParser.cs
```diff
@@ -44,6 +45,80 @@ public sealed class ParsedDMIDescription {
...
+                                    else
+                                    {
+                                        // Delay of the nearest frame. Hopefully. Christ
+                                        var stateDirs = States.Last().Value.Directions;
+                                        frames[goodFrame].Delay = dir is null
+                                            ? stateDirs.First().Value.Last().Delay
+                                            : stateDirs[dir.Value].Last().Delay;
+                                    }
+                                }
+                            }
```
wixoaGit, line 102: This is an incorrect implementation of the `delay` arg. This arg does nothing when it is `0` or `null`, and otherwise sets the state to rewind if it is `< 0` and sets the frame's delay to its absolute value.
```suggestion
                            if (delay is not null or 0) {
                                state.Rewind = delay < 0;
                                frames[goodFrame].Delay = Math.Abs(delay.Value);
                            }
```

wixoaGit (changes requested):

wixoaGit (approved):

Issue #700, opened 2022-05-22
Macros, strings, and escapes issue.


Hinaichigo:
Trying to compile `vgstation13-coders/vgstation13`, the following produces the error:
`Error at code\modules\paperwork\pen.dm:23:2: Expected ']'`
```
13	#define REG_NOTBB "\[^\\\[\]+"    // [^\]]+
14	
15	// This WAS a macro, but BYOND a shit.
16	/proc/REG_BBTAG(x)
17		return "\\\[[x]\\\]"
18	
19	// [x]blah[/x]
20	/proc/REG_BETWEEN_BBTAG(x)
21		return "[REG_BBTAG(x)]([REG_NOTBB])[REG_BBTAG("/[x]")]"
22	
23	/datum/speech_filter_action/bbcode/Run(var/text, var/mob/user, var/atom/movable/P)
24		return
```
Substituting everything into line 21, it becomes:
`return "["\\\[[x]\\\]"](["\[^\\\[\]+"])["\\\[["/[x]"]\\\]"]"`

The brackets are not balanced in that string, but it seems that they are brought to parity at runtime.

wixoaGit:
The brackets aren't unbalanced. `\[` in a string escapes that bracket and isn't used in string interpolation.

out-of-phaze:
It looks more like your macro is missing some escapes: `#define REG_NOTBB "\[^\\\[\]+"    // [^\]]+` tries to interpolate `[^\]`.

out-of-phaze:
Ah, looks like I misread. That's a comment. `"\[^\\\[\]+"` is the full macro.
Can you post what the preprocessor output is?

Hinaichigo:
How do I obtain that?

wixoaGit:
> How do I obtain that?

Add the `--dump-preprocessor` argument when running the compiler. This will create a new file next to the DME called `preprocessor_dump.dm`.

Hinaichigo:
Thanks. The file has over 500k lines. Is there a certain section that I should post?

wixoaGit:
The section of code you included in the top of the issue

Hinaichigo:
Oh I just realized what was meant:

```
#define REG_NOTBB "\[^\\\[\]+"    // [^\]]+

// This WAS a macro, but BYOND a shit.
/proc/REG_BBTAG(x)
	return "\\\[[x]\\\]"

// [x]blah[/x]
/proc/REG_BETWEEN_BBTAG(x)
[...]

PR #720, opened 2022-06-12
Add \ref support for strings

ike709:


OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -178,10 +178,30 @@ static class DMOpcodeHandlers {
...
+                            {
+                                var idx = state.DreamManager.ObjectTree.Strings.IndexOf(refStr);
+                                if (idx != -1)
+                                {
+                                    formattedString.Append(idx);
+                                }
+                                else
+                                {
+                                    state.DreamManager.ObjectTree.Strings.Add(refStr);
+                                    formattedString.Append(state.DreamManager.ObjectTree.Strings.Count);
```
ike709, line 198:
```suggestion
                                    formattedString.Append(state.DreamManager.ObjectTree.Strings.Count - 1);
```

Hinaichigo:
Is this related to #700?

ike709:
no

OpenDreamRuntime/Objects/DreamObjectTree.cs
```diff
@@ -25,6 +25,8 @@ public sealed class TreeEntry {
         public TreeEntry[] Types;
         public List<DreamProc> Procs;
         public List<string> Strings; //TODO: Store this somewhere else
+        // Any time `\ref[string]` happens
+        public List<string> RefStrings = new List<string>(); //TODO: Improve this impl.
```
wixoaGit, line 29: This isn't getting used at all

wixoaGit (changes requested):
There's no way for `locate()` to know when the reference is for a string. Currently it will try to find an object with the string's ID.

ike709:
Spurious CI failure again. I think github is having an issue.

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -178,10 +178,30 @@ static class DMOpcodeHandlers {
                             formattedString.Append(value.Stringify());
                             break;
                         }
-                        case StringFormatTypes.Ref: {
-                            DreamObject refObject = state.Pop().GetValueAsDreamObject();
-
-                            formattedString.Append(refObject.CreateReferenceID(state.DreamManager));
+                        case StringFormatTypes.Ref:
```
wixoaGit, line 181: This should be moved out into shared code that both this and the `ref()` proc uses

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -1470,7 +1490,16 @@ static class DMOpcodeHandlers {
             if (value.TryGetValueAsString(out string refString)) {
                 if(int.TryParse(refString, out var refID))
                 {
-                    state.Push(new DreamValue(DreamObject.GetFromReferenceID(state.DreamManager, refID)));
+                    // TODO: This impl. isn't entirely correct, but refs aren't implemented like BYOND's yet
+                    var obj = DreamObject.GetFromReferenceID(state.DreamManager, refID);
+                    if (obj is null && state.DreamManager.ObjectTree.Strings.Count > refID)
```
wixoaGit, line 1495: This leaves the possibility of a `ref()` on a string sharing an id with a DreamObject, and a `locate()` on that ref returning that DreamObject instead.

wixoaGit (changes requested):

PJB3005:
Isn't this going to permanently leak all strings passed through `\ref`? You may want a different string tree that uses weak references instead, and some sort of cleanup routine.

ike709:
True but this is still pretty much a placeholder ref implementation anyways to bandaid round init while we work on other things.

OpenDreamRuntime/IDreamManager.cs
```diff
@@ -33,6 +33,33 @@ public interface IDreamManager {
...
+                }
+
+                ObjectTree.Strings.Add(refStr);
+                var id = ObjectTree.Strings.Count - 1;
+                return Convert.ToInt32(string.Format("{0}{1}", 2, id));
+            }
+
+
+
+            throw new NotImplementedException();
```
wixoaGit, line 60:
```suggestion

            throw new NotImplementedException();
```

OpenDreamRuntime/Objects/DreamObject.cs
```diff
@@ -30,14 +30,18 @@ public class DreamObject {
...
                 if (referenceIDPair.Value == refID) return referenceIDPair.Key;
             }
 
             return null;
         }
 
+
+        /// <summary>
+        /// Do NOT use this directly, instead use IDreamManager.CreateRef()
+        /// </summary>
```
wixoaGit, line 44: This should be moved to inside `DreamManager.CreateRef()` to prevent misuses & confusion. It only operates on a dictionary located in `DreamManager` anyway, there's no longer any reason for it to be here.

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -1470,7 +1470,35 @@ static class DMOpcodeHandlers {
             if (value.TryGetValueAsString(out string refString)) {
                 if(int.TryParse(refString, out var refID))
                 {
-                    state.Push(new DreamValue(DreamObject.GetFromReferenceID(state.DreamManager, refID)));
+                    // The first digit is the type
+                    var typeId = int.Parse(refString.Substring(0, 1));
+                    refID = int.Parse(refString.Substring(1));
```
wixoaGit, line 1475: May be better for code quality to move this to a `LocateRef()` method in DreamManager.

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -1470,7 +1470,35 @@ static class DMOpcodeHandlers {
...
+
+                    switch (typeId)
+                    {
+                        // DreamObject
+                        case 1:
+                        {
+                            var obj = DreamObject.GetFromReferenceID(state.DreamManager, refID);
+                            state.Push(obj is null ? DreamValue.Null : new DreamValue(obj));
```
wixoaGit, line 1483: `new DreamValue(null)` will create a value equivalent to `DreamValue.Null`
```suggestion
                            state.Push(new DreamValue(obj));
```

wixoaGit (changes requested):

ike709:
Review addressed but I haven't thoroughly tested the changes.

wixoaGit:
This breaks `locate()` for DreamObjects currently. `locate("10")` looks for a DreamObject with ID `10` instead of `0`.

OpenDreamRuntime/IDreamManager.cs
```diff
@@ -35,6 +35,75 @@ public interface IDreamManager {
...
+
+            // The first digit is the type
+            var typeId = (RefType)int.Parse(refString.Substring(0, 1));
+            refId = int.Parse(refString.Substring(1));
+
+            switch (typeId)
+            {
+                case RefType.DreamObject:
+                {
+                    var obj = DreamObject.GetFromReferenceID(this, refString);
```
ike709, line 89:
```suggestion
                    var obj = DreamObject.GetFromReferenceID(this, refString.Substring(1));
```

OpenDreamRuntime/IDreamManager.cs
```diff
@@ -35,6 +35,75 @@ public interface IDreamManager {
...
+
+            // The first digit is the type
+            var typeId = (RefType)int.Parse(refString.Substring(0, 1));
+            refId = int.Parse(refString.Substring(1));
+
+            switch (typeId)
+            {
+                case RefType.DreamObject:
+                {
+                    var obj = DreamObject.GetFromReferenceID(this, refString.Substring(1));
```
ike709, line 89:
```suggestion
            var untypedRefString = refString.Substring(1); // The ref minus its ref type prefix
            refId = int.Parse(untypedRefString);

            switch (typeId)
            {
                case RefType.DreamObject:
                {
                    var obj = DreamObject.GetFromReferenceID(this, untypedRefString);
```

ike709:
Ready for review

wixoaGit:
> Isn't this going to permanently leak all strings passed through `\ref`? You may want a different string tree that uses weak references instead, and some sort of cleanup routine.

Yes, this will allow you to get any string existing at compile time or passed through `ref()`. This is actually similar behavior to byond, which can be observed with the following code:

```
for (var/i in 1 to 0x1a0)
	world.log << locate("\[0x6000[num2text(i, 3, 16)]\]")
```

I don't think a cleanup routine would be possible, as there's no way to tell when the code will want to `locate()` it.

wixoaGit (changes requested):

PR #725, opened 2022-06-14
Fix equality between a path and a float

ike709:


ike709:
Test failure is a github actions issue, try re-running.

PR #740, opened 2022-06-17
Implement proc regex replace

### LINKED ISSUE #638: Regex proc replacement is not implemented
Refer to the bottom section of http://www.byond.com/docs/ref/#/regex/proc/Replace

ike709:
Closes #638 

I did basic testing with strings, procs, and just passing null. But I did not come up with every imaginable case, hence the TODOs.

### DIFF
--- a/OpenDreamRuntime/Procs/Native/DreamProcNativeRegex.cs
+++ b/OpenDreamRuntime/Procs/Native/DreamProcNativeRegex.cs
@@ -1,6 +1,7 @@
 ﻿using System.Text.RegularExpressions;
 using OpenDreamRuntime.Objects;
 using OpenDreamRuntime.Objects.MetaObjects;
+using OpenDreamShared.Dream;
 using DreamRegex = OpenDreamRuntime.Objects.MetaObjects.DreamMetaObjectRegex.DreamRegex;
 
 namespace OpenDreamRuntime.Procs.Native {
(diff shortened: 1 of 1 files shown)


OpenDreamRuntime/Procs/Native/DreamProcNativeRegex.cs
```diff
@@ -62,22 +63,70 @@ static class DreamProcNativeRegex {
...
+            DreamValue DoProcReplace(DreamProc proc)
+            {
+                var match = dreamRegex.Regex.Match(haystackSubstring);
+                var captures = match.Captures;
+                List<DreamValue> args = new List<DreamValue>(captures.Count + 1);
+                args.Add(new DreamValue(match.Value));
+                foreach (var capture in captures)
+                {
+                    args.Add(capture is null ? DreamValue.Null : new DreamValue(capture));
+                }
```
wixoaGit, line 108: `captures` is a collection of `Capture` objects, not strings; you need to use their `Value` property. Even then you won't get the correct args because it won't give you all the groups captured by the regex.

`Groups` is the correct thing to use here, the first group will be the entire match and the rest will be what was matched by the capture groups (`( ... )` in the regex).
```suggestion
                var groups = match.Groups;
                List<DreamValue> args = new List<DreamValue>(groups.Count);
                foreach (Group group in groups)
                {
                    args.Add(new DreamValue(group.Value));
                }
```

OpenDreamRuntime/Procs/Native/DreamProcNativeRegex.cs
```diff
@@ -62,22 +63,70 @@ static class DreamProcNativeRegex {
...
+                var result = DreamThread.Run(async(state) => await state.Call(proc, instance, usr, new DreamProcArguments(args)));
+                if (result.TryGetValueAsString(out var replacement))
+                {
+                    return DoTextReplace(replacement);
+                }
+                //TODO Confirm this behavior
+                if (result.Value is null)
```
wixoaGit, line 115:
```suggestion
                if (result == DreamValue.Null)
```

wixoaGit (changes requested):

OpenDreamRuntime/Procs/Native/DreamProcNativeRegex.cs
```diff
@@ -62,22 +63,68 @@ static class DreamProcNativeRegex {
...
+                    var proc = dreamMan.GlobalProcs[procPath.LastElement];
+                    return DoProcReplace(proc);
+                }
+            }
+
+            throw new ArgumentException("Replacement argument must be a string or a proc");
+
+            DreamValue DoProcReplace(DreamProc proc)
+            {
+                var match = dreamRegex.Regex.Match(haystackSubstring);
```
ike709, line 100:
```suggestion
                if(dreamRegex.IsGlobal)
                {
                    throw new NotImplementedException("Proc global regex replacements are not implemented");
                }
                var match = dreamRegex.Regex.Match(haystackSubstring);
```

wixoaGit (approved):

PR #749, opened 2022-07-13
Object can end with curly braket

wrexbe:
Ex:
```
turf
	start {icon = 'forest.dmi'}
```

### DIFF
--- a/DMCompiler/Compiler/DM/DMParser.cs
+++ b/DMCompiler/Compiler/DM/DMParser.cs
@@ -252,7 +252,10 @@ namespace DMCompiler.Compiler.DM {
                         statement = new DMASTObjectDefinition(loc, _currentPath, null);
                     }
 
-                    if (requireDelimiter && !PeekDelimiter() && Current().Type != TokenType.DM_Dedent) {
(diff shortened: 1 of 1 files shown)


DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -252,7 +252,10 @@ public DMASTStatement Statement(bool requireDelimiter = true)
                         statement = new DMASTObjectDefinition(loc, _currentPath, null);
                     }
 
-                    if (requireDelimiter && !PeekDelimiter() && Current().Type != TokenType.DM_Dedent) {
+                    var currentType = Current().Type;
```
wixoaGit, line 255: This isn't a proper fix, handling of the brackets should be left to the `BracedBlock()` method. A better fix may be to have that method end up calling `Statement()` with the `requireDelimiter` argument set to false since delimiters do not appear to be required in braced blocks.

wixoaGit (changes requested):

ike709:
Stale?

Altoids1:
Stale?

wixoaGit:
Stale

PR #757, opened 2022-07-14
Fix  the `||=` and `&&=` operators

ike709:


wixoaGit:
There's a very similar problem in the `&&=` operator, could you apply the fix to that too?

ike709:
@wixoaGit Done and tested.

PR #760, opened 2022-07-18
Cursed turf handling patch

ike709:
This is horrible but I cba to refactor turfs.


wixoaGit:
A more correct fix was done in ae653f0a96ab8c7ad371408edd2f972d7f462cc8

PR #762, opened 2022-07-18
Improve screen_loc parsing

ike709:
- Ignore secondary map controls
- Improve screen_loc exception message usefulness
- Implement doing just `"CENTER"`
- TODO supporting non-15x15


wixoaGit (changes requested):

ike709:
I don't even remember why I was working on this at this point tbh

PR #765, opened 2022-07-23
Fix an issue with ternaries and list indexes

ike709:
See the unit test for the code that this fixes.

Fixes a TG compile error

### DIFF
--- a/DMCompiler/Compiler/DM/DMParser.cs
+++ b/DMCompiler/Compiler/DM/DMParser.cs
@@ -919,7 +919,7 @@ namespace DMCompiler.Compiler.DM {
                 if (Check(TokenType.DM_RightParenthesis)) {
                     return new DMASTProcStatementInfLoop(loc, GetForBody());
                 }
-            
+
                 _allowVarDeclExpression = true;
                 DMASTExpression expr1 = Expression();
                 Whitespace();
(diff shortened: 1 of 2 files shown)


DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -1615,7 +1615,13 @@ public DMASTProcStatementLabel Label(DMASTIdentifier expression)
                         } else {
                             Error("Expected ':'");
                         }
-                    } else {
+                    } else if (b is DMASTListIndex idx && idx.Expression is DMASTDereference idxDeref && idxDeref.Expression is DMASTListIndex)
```
wixoaGit, line 1618: This doesn't work with expressions like `prob(50)?L[1]:L[2]:L[3]`. It should be parsed as `prob(50) ? L[1] : L[2]:L[3]` but is instead parsed as `prob(50) ? L[1]:L[2] : L[3]`.

This is the reason the other case above uses a do/while loop. It'd be nice if this could somehow be handled within that same loop, but I'm not sure what the best way to do that would be.

wixoaGit (changes requested):

ike709:
I'm going to revisit this some time in the far, distant future.

PR #767, opened 2022-07-23
Fix for loops causing exceptions

ike709:
For loops previously would cause a C# exception if the iterator var was invalid, because the expression would be null.

This is a non-ideal bandaid to fix throwing the exception.


wixoaGit:
Invalid iterator vars are handled a little better as of 8df158092f6c9056ccc10480f3c10e7c860159bc, making this unnecessary.

PR #772, opened 2022-08-17
Fixes invalid switch-case ranges not causing warnings/errors

Altoids1:
image

## Summary

While I was thumbing through the code, I found that there wasn't really any typechecking done on switch ranges.
I checked MoMMI and it seems this bug does indeed also exist in DM 514, but wixoa agreed that there should be at least a warning, if not an error, for this behaviour, so I went ahead and added it.

I do not know of any repository which actually makes use of this bug as it exists in DM, so for now this PR marks most cases as being an honest error.

## Changelog
-  Added some exceptions for trying to make a switch-case's range's bounds be anything other than numbers. Nulls are coerced into the value 0 with a compiler warning, and anything else causes a compiletime error.

### DIFF
--- a/DMCompiler/DM/Visitors/DMProcBuilder.cs
+++ b/DMCompiler/DM/Visitors/DMProcBuilder.cs
@@ -668,12 +668,37 @@ namespace DMCompiler.DM.Visitors {
                     string caseLabel = _proc.NewLabelName();
 
                     foreach (DMASTExpression value in switchCaseValues.Values) {
-                        if (value is DMASTSwitchCaseRange range) {
+                        if (value is DMASTSwitchCaseRange range) { // if(1 to 5) or something
                             if (!DMExpression.TryConstant(_dmObject, _proc, range.RangeStart, out var lower))
                                 throw new CompileErrorException(new CompilerError(range.RangeStart.Location, "Expected a constant"));
                             if (!DMExpression.TryConstant(_dmObject, _proc, range.RangeEnd, out var upper))
                                 throw new CompileErrorException(new CompilerError(range.RangeEnd.Location, "Expected a constant"));
-
+                            //DM 514.1580 does NOT care if the constants within a range are strings, and does a strange conversion to 0 or something, without warning or notice.
+                            //We are deviating from parity here and just calling that a CompilerError.
+                            if(lower is not Number)
+                            {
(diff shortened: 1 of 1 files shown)


DMCompiler/DM/Visitors/DMProcBuilder.cs
```diff
@@ -668,12 +668,37 @@ public void ProcessStatementSet(DMASTProcStatementSet statementSet)
...
+                            if(lower is not Number)
+                            {
+                                if (lower is Null) // We do a little null coercion, as a treat
+                                    {
+                                    lower = new Number(lower.Location, 0.0f);
+                                    DMCompiler.Warning(new CompilerWarning(range.RangeStart.Location, "Malformed range, lower bound is coerced from null to 0"));
+                                }
+                                else
+                                {
+                                    throw new CompileErrorException(new CompilerError(range.RangeStart.Location, "Invalid range, lower bound is not a number"));
```
wixoaGit, line 687: It would be better to use `DMCompiler.Error()` here and continue generating the bytecode for the switch case. Doing it this way leads to other possible errors not being found as well as invalid uses of labels (creating errors like `Invalid label "label1"`).

DMCompiler/DM/Visitors/DMProcBuilder.cs
```diff
@@ -668,12 +668,37 @@ public void ProcessStatementSet(DMASTProcStatementSet statementSet)
...
+                            if(upper is not Number)
+                            {
+                                if (upper is Null)
+                                {
+                                    upper = new Number(upper.Location, 0.0f);
+                                    DMCompiler.Warning(new CompilerWarning(range.RangeEnd.Location, "Malformed range, upper bound is coerced from null to 0"));
+                                }
+                                else
+                                {
+                                    throw new CompileErrorException(new CompilerError(range.RangeEnd.Location, "Invalid range, upper bound is not a number"));
```
wixoaGit, line 699: Same here

wixoaGit (changes requested):

DMCompiler/DM/Visitors/DMProcBuilder.cs
Altoids1: You got it

wixoaGit (approved):

PR #773, opened 2022-08-17
Adds a sanity check to ensure opcodes are all unique

### LINKED ISSUE #721: Validate that no opcodes have the same values
Check that every opcode in the enum has a unique value and throw if it doesn't.

Altoids1:
image

## Summary

Closes #721 by implementing an opcode verification step done by DMCompiler, with the code in OpenDreamShared in case anything else wants to double-check itself.

It doesn't really seem like C# takes well to static analysis done at compile-time, so for now this is just something DMCompiler does when doing a compilation.

## Changelog
- The Compiler now warns if some opcodes share the same byte value.


DMCompiler/DMCompiler.cs
```diff
@@ -39,6 +39,10 @@ public static class DMCompiler {
             if (settings.SuppressUnimplementedWarnings) {
                 Warning(new CompilerWarning(Location.Internal, "Unimplemented proc & var warnings are currently suppressed"));
             }
+            if(OpenDreamShared.Dream.Procs.OpcodeVerifier.AreOpcodesInvalid())
+            {
+                Warning(new CompilerWarning(Location.Internal, "Some opcodes have the same byte value! Output assembly may be corrupted."));
```
ike709, line 44: IMO it should error.

Altoids1: Makes no difference to me.

OpenDreamShared/Dream/Procs/DreamProcOpcode.cs
```diff
@@ -248,4 +249,31 @@ public enum Type : byte {
...
+        /// <summary>
+        /// Validates that the opcodes in DreamProcOpcode are all unique, such that none resolve to the same byte.
+        /// </summary>
+        /// <returns>True if there are duplicate opcodes, false if not</returns>
+        public static bool AreOpcodesInvalid() // FIXME: Can this be made into something done during compiletime? Like, *this code's* compiletime? >:/
+        {
+            // I'm not *too* satisfied with this boolean schtick, as opposed to throwing,
+            // but since we're in OpenDreamShared I want each executable to be able to do what they want with this information.
+            
+            ///Key is an int we're already using for an opcode
```
wixoaGit, line 265: Three slashes makes an XML comment which is not valid here
```suggestion
            // Key is an int we're already using for an opcode
```

OpenDreamShared/Dream/Procs/DreamProcOpcode.cs
```diff
@@ -248,4 +249,31 @@ public enum Type : byte {
...
+            HashSet<int> byte_pool = new(); // FIXME: Do not assume that the underlying type is an int. Somehow. Y'all don't have constexpr so
+            foreach (int used_int in Enum.GetValues(typeof(DreamProcOpcode)))
+            {
+                if(byte_pool.Contains(used_int))
+                {
+                    return true;
+                }
+                byte_pool.Add(used_int);
+            }
```
wixoaGit, line 274: You can use `DreamProcOpcode` instead of `int` to do what the `FIXME` suggests.

`byte_pool` and `used_int` do not fit the naming convention for local variables in the rest of the project.
```suggestion
            HashSet<DreamProcOpcode> bytePool = new();
            foreach (DreamProcOpcode usedInt in Enum.GetValues(typeof(DreamProcOpcode)))
            {
                if(bytePool.Contains(usedInt))
                {
                    return true;
                }
                bytePool.Add(usedInt);
            }
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #775, opened 2022-08-21
Fixes writing to constants, adds readonly opendream pseudo-type, and makes var overrides respect their var's type

### LINKED ISSUE #535: Const vars can be reassigned or overridden
Constant vars can be reassigned by code or overridden in object definitions.

```
var/const/C = 1
world.log << C
C = 2
world.log << C
```

Note that in const-only contexts (such as switch cases) the initial value will still be used because the compiler uses the initial value at compile-time.

### LINKED ISSUE #702: Need to be able to flag vars as compiletime-readonly
`list.type` for example is readonly but not a normal const.

Something like `as opendream_readonly` as a new astype probably, similar to unimplementeds.

### LINKED ISSUE #738: Overriden vars don't keep type information
When trying to compile /vg/station, the following error occurs:
`Error at code\modules\cmc\crew.dm:32:15: An inferred new requires a type!`
at the following location:
```
21 /obj/machinery/computer/crew
...
32 	_using = new()
```
`_using` is defined in `code/game/objects/objs.dm` as:
`var/list/mob/_using`
In game, interacting with the crew monitoring computer adds the user's mob to `_using`

Altoids1:
image

Fixes #738, fixes #702, and fixes #535.

## Summary
This bugfix stuff comes with a refactor of `DMObjectBuilder` into the `static class` it probably should be.

As a treat, this PR also has a commit that amends Content.Test so that it actually prints the message of any unexpected runtime into the log file, instead of just declaring that there was a runtime. Should be helpful :)

## Changelog
- `const` now works as an effective keyword, preventing writing of values, within both procs and object definitions.
- Assignments now consistently remember the type of what they're assigning to (for implicit `new` calls)
- Variable overrides in object definitions now work correctly, remembering their initial type and preferring overrides that occur later in the source code, as per DM parity.
- `opendream_compiletimereadonly` is now a valid type marker. Allows for marking something as unwriteable, while also not requiring the value to be literally a `const`.
- Type markers have been slightly reworked so that they are parented to the `DMVariable` they modify, rather than the expression.
- Content.Test now logs the message of any unexpected runtime discovered during testing.
- Fixed statics being overridable in derived classes.
- Fixed vars named `tag` not being overridable in objects that don't inherit from /datum.
- Miscellaneous null-safety improvements!

### DIFF
--- a/Content.Tests/DMTests.cs
+++ b/Content.Tests/DMTests.cs
@@ -89,7 +89,7 @@ namespace Content.Tests
         }
 
         private (bool Success, DreamValue Returned, Exception? except) RunTest() {
-            var prev = _dreamMan.LastException;
+            var prev = _dreamMan.LastDMException;
 
             var result = DreamThread.Run(async (state) => {
                 if (_dreamMan.ObjectTree.TryGetGlobalProc("RunTest", out DreamProc proc)) {
@@ -99,11 +99,11 @@ namespace Content.Tests
                     return DreamValue.Null;
                 }
             });
-            bool retSuccess = _dreamMan.LastException == prev;
+            bool retSuccess = _dreamMan.LastDMException == prev; // Works because "null == null" is true in this language.
             if (retSuccess)
                 return (retSuccess, result, null);
             else
-                return (false, result, _dreamMan.LastException);
+                return (false, result, _dreamMan.LastDMException);
         }
 
         private static IEnumerable<object[]> GetTests()
--- a/Content.Tests/DMTests.cs
+++ b/Content.Tests/DMTests.cs
@@ -70,12 +70,14 @@ namespace Content.Tests
             Assert.IsTrue(compiledFile is not null && File.Exists(compiledFile), $"Failed to compile DM source file");
             Assert.IsTrue(_dreamMan.LoadJson(compiledFile), $"Failed to load {compiledFile}");
 
-            (bool successfulRun, DreamValue returned) = RunTest();
+            (bool successfulRun, DreamValue returned, Exception? exception) = RunTest();
             if (testFlags.HasFlag(DMTestFlags.RuntimeError)) {
                 Assert.IsFalse(successfulRun, "A DM runtime exception was expected");
             } else {
-                //TODO: This should use the runtime exception as the failure message
-                Assert.IsTrue(successfulRun, "A DM runtime exception was thrown");
+                if (exception != null)
+                    Assert.IsTrue(successfulRun, $"A DM runtime exception was thrown: \"{exception.Message}\"");
+                else
+                    Assert.IsTrue(successfulRun, "A DM runtime exception was thrown, and its message could not be recovered!");
             }
 
             if (testFlags.HasFlag(DMTestFlags.ReturnTrue)) {
@@ -86,8 +88,8 @@ namespace Content.Tests
             Cleanup(compiledFile);
         }
 
-        private (bool Success, DreamValue Returned) RunTest() {
-            var prev = _dreamMan.DMExceptionCount;
+        private (bool Success, DreamValue Returned, Exception? except) RunTest() {
+            var prev = _dreamMan.LastException;
 
             var result = DreamThread.Run(async (state) => {
                 if (_dreamMan.ObjectTree.TryGetGlobalProc("RunTest", out DreamProc proc)) {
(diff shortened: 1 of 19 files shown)


DMCompiler/DM/Visitors/DMVisitorExpression.cs
```diff
@@ -192,6 +192,10 @@ public void VisitIdentifier(DMASTIdentifier identifier)
         public void VisitAssign(DMASTAssign assign) {
             var lhs = DMExpression.Create(_dmObject, _proc, assign.Expression, _inferredPath);
             var rhs = DMExpression.Create(_dmObject, _proc, assign.Value, lhs.Path);
+            if(lhs.TryAsConstant(out var _) || ((lhs.ValType & DMValueType.CompiletimeReadonly) == DMValueType.CompiletimeReadonly))
```
wixoaGit, line 195: This does not work. This is because `ValType` is actually a bit of an unfinished property and doesn't always reflect the type of its value.

`opendream_unimplemented` works by attaching the value type to the expression instead of the variable, and snowflaking it with dereference expressions which is less than ideal.

DMCompiler/DM/Expressions/Dereference.cs
```diff
@@ -59,7 +59,7 @@ public override bool TryAsConstant(out Constant constant)
             {
                 var obj = DMObjectTree.GetDMObject(_expr.Path.GetValueOrDefault());
                 var variable = obj.GetVariable(PropertyName);
-                if (variable.IsConst)
+                if (variable != null && variable.IsConst)
```
wixoaGit, line 62:
```suggestion
                if (variable?.IsConst)
```

wixoaGit (changes requested):

DMCompiler/DM/Visitors/DMObjectBuilder.cs
```diff
@@ -124,44 +144,54 @@ class DMObjectBuilder {
...
 
                         if (parentType == null) throw new CompileErrorException(varOverride.Location, "Expected a constant path");
-                        _currentObject.Parent = DMObjectTree.GetDMObject(parentType.Value.Path);
-                        break;
+                            varObject.Parent = DMObjectTree.GetDMObject(parentType.Value.Path);
+                        return;
                     }
                     case "tag":
                         DMCompiler.Error(new CompilerError(varOverride.Location, "tag: may not be set at compile-time"));
-                        break;
```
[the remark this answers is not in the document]
Altoids1: I actually intend on dishing out `opendream_compiletimereadonly` in a separate PR after this one. I'll leave a TODO here though, since I didn't notice that these would need to be changed.

Content.Tests/DMTests.cs
```diff
@@ -97,8 +99,11 @@ public void TestFiles(string sourceFile, DMTestFlags testFlags)
                     return DreamValue.Null;
                 }
             });
-
-            return (_dreamMan.DMExceptionCount == prev, result);
+            bool retSuccess = _dreamMan.LastException == prev;
+            if (retSuccess)
+                return (retSuccess, result, null);
+            else
+                return (false, result, _dreamMan.LastException);
```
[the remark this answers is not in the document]
Altoids1: The main reason I didn't just have it be an infinite list is because this variable is also kept when a real server is running; in very expection-spamming scenarios it would just be a memory leak with no benefit to the runtime as of yet.

We can change this once try-catch is implemented and we have a stronger implementation for exception handling on the runtime.

DMCompiler/DM/DMObject.cs
```diff
@@ -31,8 +36,58 @@ class DMObject {
...
+            {
+                var varOverride = danglingOverrides[i];
+                if (varOverride.VarName == varDefined.Name) // FINALLY we can do this
+                {
+                    if (IsSubtypeOf(maybeAncestor.Path))
+                    {
+                        // Thank god DMObjectBuilder is static, amirite?
+                        DMObjectBuilder.OverrideVariableValue(this, ref varDefined, varOverride.Value); // I'd like to mark DMObjectBuilder as a friend class but that's not a thing in C# so
+                        VariableOverrides[varDefined.Name] = varDefined;
+                        //DMCompiler.Warning(new CompilerWarning(varOverride.Location, $"Resolved ambiguous var override for {varOverride.VarName}"));
```
wixoaGit, line 52:
```suggestion
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #777, opened 2022-08-23
Fixes several bugs with string macros, implements all pronoun/article ones

Altoids1:
image

## Summary

This is a big fat redesign of how we handle string formatting.

Previously, we used to do formatting by sending two UTF16 characters, a 0x00FF and then another one to indicate the use of the interpolated value.

This has been refactored and replaced with `StringFormatEncoder` and its various helper procs and enums. We now (in parity with DM, at least for the first byte) emit 0xFFxx, with the bottom byte denoting the use of the format byte.

Doing this allowed me to implement \the, \The, \A, \an, \himself, and a bunch of other articles and pronouns.

## Changelog
- All of the article and pronoun text macros (\He, \the, \himself, etc. ) are now implemented.
- The Compiler now correctly throws if text macros are missing an interpolated value to depend on.
- The Compiler now gives a descriptive error if the user attempts to use the (ambiguous) \her or \Her text macros.
- The Compiler now is lenient in some cases towards accidental capitalization of lowercase text macros.
- Added explicit unimplemented warning for \th and \s text macros.
- Fixed, probably, several bugs with string formatting.


DMCompiler/Compiler/DM/DMParserHelper.cs
```diff
@@ -243,13 +248,21 @@ private DMASTExpression ExpressionFromString(Token constantToken)
                                         stringBuilder.Append((char)StringFormatTypes.LowerPossessiveAdjective);
                                         break;
 
+                                    case "Him": // BYOND errors here but lets be nice!
+                                        Warning("\"\\Him\" is not an available text macro. Coercing macro into \"\\him\"");
+                                        goto case "him"; // Fallthrough!
```
wixoaGit, line 253: I.. did not know you could do this

Altoids1: The wonders of C# programming 🥴

ZeWaka: neat

OpenDreamShared/Dream/Procs/DreamProcOpcode.cs
```diff
@@ -105,43 +107,119 @@ public enum DreamProcOpcodeParameterType {
...
+    /// <summary>
+    /// Handles how we write format data into our strings.
+    /// </summary>
+    public static class StringFormatEncoder
+    {
+        /// <summary>
+        /// This is the upper byte of the 2-byte markers we use for storing formatting data within our UTF16 strings.<br/>
+        /// </summary>
+        /// <remarks>
+        /// It is not const because (<see langword="TODO"/>) eventually it would be desirable to make it something else<br/>
```
wixoaGit, line 119: Why would a possible future change prevent this from being const?

OpenDreamShared/Dream/Procs/DreamProcOpcode.cs
```diff
@@ -105,43 +107,119 @@ public enum DreamProcOpcodeParameterType {
...
+        /// </summary>
+        /// <remarks>
+        /// <see langword="DO NOT CAST TO CHAR!"/> This requires FormatPrefix to be added to it in order to be a useful formatting character!!
+        /// </remarks>
+        public enum FormatSuffix : UInt16
+        {
+            //States that Interpolated values can have (the [] thingies)
+            StringifyWithArticle = 0x0,    //[] and we include an appropriate article for the resulting value, if necessary            
+            StringifyNoArticle = 0x1,      //[] and we never include an article (because it's elsewhere)
+            ReferenceOfValue = 0x2,        //\ref[]
```
wixoaGit, line 139: If `FormatPrefix` was const you could change these lines to
```suggestion
            StringifyWithArticle = FormatPrefix,    //[] and we include an appropriate article for the resulting value, if necessary
            StringifyNoArticle,      //[] and we never include an article (because it's elsewhere)
            ReferenceOfValue,        //\ref[]
```
This would have every value start from `0xFF00` and count up, which could simplify things a good bit.

wixoaGit (changes requested):

OpenDreamShared/Dream/Procs/DreamProcOpcode.cs
Altoids1: My concept of what a parity-breaking OD would do is, take in an argument during compilation that decides whether to support or not support certain behaviours of DM.

That sort of support, including making this byte not clobber valid unicode, would be a decision at the runtime of the compiler, so it can't be const.

OpenDreamShared/Dream/Procs/DreamProcOpcode.cs
Altoids1: Again, tried to code it for the compiler deciding to do alternative behaviour during its runtime, when reading configuration.

wixoaGit (approved):

PR #779, opened 2022-08-25
Adds primitive implementation for /cilent.computer_id

### LINKED ISSUE #181: `client/computer_id` not implemented
http://www.byond.com/docs/ref/#/client/var/computer_id

NTstation

Altoids1:
Fixes #181

Right now we're really just waiting for RobustToolbox to implement something rigorous.

For now, we're just MD5ing the ckey.

## Note

In the future, I'd like to implement some sort of suite of security warnings to caution against using `computer_id` and other security-insecure things like this. Probably after the warning rework, whenever that happens.

### DIFF
--- a/OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectClient.cs
+++ b/OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectClient.cs
@@ -1,6 +1,8 @@
 using OpenDreamRuntime.Procs;
 using OpenDreamRuntime.Rendering;
 using OpenDreamShared.Dream;
+using System.Security.Cryptography;
+using System.Text;
 
 namespace OpenDreamRuntime.Objects.MetaObjects {
     sealed class DreamMetaObjectClient : IDreamMetaObject {
@@ -98,6 +100,14 @@ namespace OpenDreamRuntime.Objects.MetaObjects {
                 case "key":
                 case "ckey":
                     return new(_dreamManager.GetSessionFromClient(dreamObject).Name);
+                case "computer_id": // FIXME: This is not secure! Whenever RT implements a more robust (heh) method of uniquely identifying computers, replace this impl with that.
+                    MD5 md5 = MD5.Create();
+                    /// <remarks>Check on <see cref="Robust.Shared.Network.NetUserData.HWId"/> if you want to seed from how RT does user identification.
+                    /// We don't use it here because it is probably not enough to ensure security, and (as of time of writing) only works on Windows machines.</remarks>
+                    byte[] brown = Encoding.UTF8.GetBytes(_dreamManager.GetSessionFromClient(dreamObject).Name);
+                    byte[] hash = md5.ComputeHash(brown);
+                    string hashStr = BitConverter.ToString(hash).Replace("-", "").ToLower().Substring(0,15); // Extracting the first 15 digits to ensure it'll fit in a 64-bit number
+                    return new(long.Parse(hashStr, System.Globalization.NumberStyles.HexNumber).ToString()); // Converts from hex to decimal. Output is in analogous format to BYOND's.
                 case "address":
                     return new(_dreamManager.GetSessionFromClient(dreamObject).ConnectedClient.RemoteEndPoint.Address.ToString());
                 case "inactivity":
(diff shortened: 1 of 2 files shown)


OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectClient.cs
```diff
@@ -98,6 +100,12 @@ sealed class DreamMetaObjectClient : IDreamMetaObject {
                 case "key":
                 case "ckey":
                     return new(_dreamManager.GetSessionFromClient(dreamObject).Name);
+                case "computer_id": // FIXME: This is not secure! Whenever RT implements a more robust (heh) method of uniquely identifying computers, replace this impl with that.
+                    MD5 md5 = MD5.Create();
+                    byte[] brown = Encoding.UTF8.GetBytes(_dreamManager.GetSessionFromClient(dreamObject).Name);
```
ZeWaka, line 105: Why not `UserId`?

Altoids1: _"The best way to get the right answer on the internet is not to ask a question; it's to post the wrong answer."_

ZeWaka: not sure if the FIXME still applies

Altoids1: The actual origin of HWId is somewhere deep in the dependencies of RT. I don't happen to know if it is (or even supposed to be) difficult to spoof in any way. In the future, we will still want to append other information onto this value to provide the best identification we reasonably can. Servers rely on this stuff to make their ban system work.

ZeWaka: Wouldn't we append that information on the RT side though? 

Also, SS14 currently uses this for their bans.

wixoaGit: It looks like HWId is [generated using random numbers upon the first run](https://github.com/space-wizards/RobustToolbox/blob/a700750d9e48045fe4bebf31a35318d19ed09329/Robust.Shared/Network/HWId.cs#L20), and is then cached in the registry. It appears to not work for users not on Windows, so this may not be the best placeholder.

ike709: Hence why I said to just use a ckey hash for now

Altoids1: Aaagh!!

OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectClient.cs
```diff
@@ -98,6 +100,14 @@ sealed class DreamMetaObjectClient : IDreamMetaObject {
...
                 case "ckey":
                     return new(_dreamManager.GetSessionFromClient(dreamObject).Name);
+                case "computer_id": // FIXME: This is not secure! Whenever RT implements a more robust (heh) method of uniquely identifying computers, replace this impl with that.
+                    MD5 md5 = MD5.Create();
+                    /// <remarks>Check on <see cref="Robust.Shared.Network.NetUserData.HWId"/> if you want to seed from how RT does user identification.
+                    /// We don't use it here because it is probably not enough to ensure security, and (as of time of writing) only works on Windows machines.</remarks>
+                    byte[] brown = Encoding.UTF8.GetBytes(_dreamManager.GetSessionFromClient(dreamObject).Name);
+                    byte[] hash = md5.ComputeHash(brown);
+                    string hashStr = BitConverter.ToString(hash).Replace("-", "").ToLower();
+                    return new(long.Parse(hashStr, System.Globalization.NumberStyles.HexNumber).ToString()); // Converts from hex to decimal. Output is in analogous format to BYOND's.
```
wixoaGit, line 110: This is throwing an exception on me due to `hashStr` being too large for an Int64 (with the username `JoeGenero2`)

wixoaGit (changes requested):

OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectClient.cs
Altoids1: 😒 forgot C# has overflow exceptions

Altoids1:
image

wixoaGit (approved):

PR #781, opened 2022-08-28
Implements the <> operator

Altoids1:
image

I don't know what ancient TI-BASIC stans are still using this operator in Current Year, but, whatever, it's supported now.

It looks like it's completely equivalent to != so I just implemented it as being read the same way. An easy patch.

### DIFF
--- a/DMCompiler/Compiler/DM/DMLexer.cs
+++ b/DMCompiler/Compiler/DM/DMLexer.cs
@@ -174,6 +174,7 @@ namespace DMCompiler.Compiler.DM {
                                 case "=": token = CreateToken(TokenType.DM_Equals, c); break;
                                 case "==": token = CreateToken(TokenType.DM_EqualsEquals, c); break;
                                 case "!": token = CreateToken(TokenType.DM_Exclamation, c); break;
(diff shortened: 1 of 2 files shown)


DMCompiler/Compiler/DM/DMLexer.cs
```diff
@@ -175,6 +175,7 @@ public partial class DMLexer : TokenLexer {
                                 case "==": token = CreateToken(TokenType.DM_EqualsEquals, c); break;
                                 case "!": token = CreateToken(TokenType.DM_Exclamation, c); break;
                                 case "!=": token = CreateToken(TokenType.DM_ExclamationEquals, c); break;
+                                case "<>": token = CreateToken(TokenType.DM_ExclamationEquals, c); break; // This is syntactically equivalent to the above so why not the same token
```
wixoaGit, line 178: You can use fall-through to avoid writing the same case again.
```suggestion
                                case "<>": // This is syntactically equivalent to the below so why not the same token
                                case "!=": token = CreateToken(TokenType.DM_ExclamationEquals, c); break;
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #783, opened 2022-09-03
Fixes not accepting colon spam with goto labels, adds support for paren-less sleep calls

### LINKED ISSUE #518: `sleep()` without parentheses is unsupported
Byond apparently lets you do stuff like `sleep 5` or `sleep value * other_value` etc.

Altoids1:
image

Fixes #518.

## Summary
In this PR is also a little new tool for the codebase: A defer pattern I found for C#. Here, it is used to consume whitespace regardless of which control flow in `ProcStatementFromExpression` does the returning. I figure it will be useful in many other places in the codebase later on.

## Changelog
- Goto labels with an exaggerated amount of colons, `::::LIKE_THIS::::`, are now parsed correctly.
- Calling sleep without any parentheses (as in `sleep 40`) is now properly implemented.
- Attempting to make a label called `sleep` now compiletimes, as per parity.

### DIFF
--- a/DMCompiler/Compiler/DM/DMAST.cs
+++ b/DMCompiler/Compiler/DM/DMAST.cs
@@ -1713,6 +1713,13 @@ namespace DMCompiler.Compiler.DM {
             Parameters = parameters;
         }
 
+        /// <summary>Helper constructor that converts the expression argument into a positional call parameter.</summary>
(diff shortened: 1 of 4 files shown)


wixoaGit:
BYOND labels no longer allow multiple `:` on either side. This appears to be because of the new v515 `:` operator.

wixoaGit:
Closing in favor of #1461

PR #785, opened 2022-09-06
Unit Test Cleanup Pass

ike709:
I don't feel like revisiting any of my open PRs right now so I did this instead.

- Moves every now-functional test out of Broken Tests
- Cleans up the still-broken Broken Tests to comply with how unit tests are ran now, they can just be tossed in once they work
- Renames some existing tests to be more descriptive


Content.Tests/DMProject/Broken Tests/Arg/ArgProcs.dm
```diff
@@ -0,0 +1,9 @@
+/proc/ArgProcs1(procs, procs2, procs3 = 3)
+	return procs
```
wixoaGit, line 2: All the `ASSERT()`s below seem to test if this returns what the value of `procs3` should be. Is this supposed to be `return procs3`?

Content.Tests/DMProject/Broken Tests/Const/ConstProc.dm
```diff
@@ -0,0 +1,7 @@
+var/const/ConstProc1_a = rgb(0,0,255)
+
+/proc/RunTest()
+	var/const/ConstProc1_b = rgb(0,0,255)
+	ASSERT(ConstProc1_a == "#0000ff")
+	ASSERT(ConstProc1_b == "#0000ff")
+	return TRUE
```
wixoaGit, line 7: The test does not need to return anything
```suggestion
```

Content.Tests/DMProject/Broken Tests/Const/ConstSort.dm
```diff
@@ -1,29 +1,22 @@
...
-
-// Const Proc 1 - No Error
-//TODO Failing test - doesn't compile
-/*var/const/ConstProc1_a = rgb(0,0,255)
-/world/proc/ConstProc1()
-	var/const/ConstProc1_b = rgb(0,0,255)
-	ASSERT(ConstProc1_a == "#0000ff")
-	ASSERT(ConstProc1_b == "#0000ff")
-	return 1*/
+	return 1
```
wixoaGit, line 22:
```suggestion
```

Content.Tests/DMProject/Tests/Const/Const3.dm
```diff
@@ -1,6 +1,8 @@
...
+// RUNTIME ERROR
+
 var/ConstZero2_a
 var/ConstZero2_b = 0
-/world/proc/Const3_Proc()
+
+/proc/RunTest()
 	ConstZero2_a = 1 / ConstZero2_b
-	return 1
+	return TRUE
```
wixoaGit, line 8:
```suggestion
```

Content.Tests/DMProject/Tests/List/ListNullArg.dm
```diff
@@ -0,0 +1,5 @@
+/proc/ListNullArg2(a[5][3])
+	ASSERT(a[1].len == 3)
+
+/proc/RunTest()
+	ListNullArg2()
```
wixoaGit, line 5: This test errors in BYOND due to `ListNullArg2()`'s argument being null. This needs a `// RUNTIME ERROR` and the top and to be moved back to the broken tests folder.

wixoaGit (changes requested):

wixoaGit (approved):

Issue #163, opened 2021-06-28
Can't `spawn()` while enumerating a list


ike709:
Code:
```
for(var/mob/M in world)
    spawn(0) M.UpdateClothing()
```

Exception:
```
=C# StackTrace=
System.NotImplementedException: The method or operation is not implemented.
   at OpenDreamRuntime.Procs.DMProcState..ctor(DMProcState other, DreamThread thread) in C:\Users\main\Documents\GitHub\OpenDream\OpenDream\OpenDreamRuntime\Procs\DMProc.cs:line 160
   at OpenDreamRuntime.Procs.DMProcState.Spawn() in C:\Users\main\Documents\GitHub\OpenDream\OpenDream\OpenDreamRuntime\Procs\DMProc.cs:line 220
   at OpenDreamRuntime.Procs.DMOpcodeHandlers.Spawn(DMProcState state) in C:\Users\main\Documents\GitHub\OpenDream\OpenDream\OpenDreamRuntime\Procs\DMOpcodeHandlers.cs:line 1129
   at OpenDreamRuntime.Procs.DMProcState.InternalResume() in C:\Users\main\Documents\GitHub\OpenDream\OpenDream\OpenDreamRuntime\Procs\DMProc.cs:line 179
   at OpenDreamRuntime.ProcState.Resume() in C:\Users\main\Documents\GitHub\OpenDream\OpenDream\OpenDreamRuntime\DreamThread.cs:line 76
```

wixoaGit:
This was fixed in 436390335973c148c36413b898df7b88976898b6 once it was realized that an enumerator is never going to be enumerated inside a spawned proc state, leaving no need to copy them.

PR #786, opened 2022-09-11
Unit Test Extravaganza - Part 1

ike709:
Ports **many** Dream Test Tools tests and converts them to valid OpenDream unit tests.

We're now up to 214 tests total, of which 142 pass. The failing tests have been added to Broken Tests as a TODO list. In the future I want to add CI to check if open PRs fix broken tests and yell at the author to enable them.

It's not even close to a 1:1 port. Here's a summary:

- Tests now rely on `ASSERT()` rather than comparing outputs to BYOND
- `main()` -> `RunTest()`
- Several DTT tests actually didn't work as intended in BYOND, so I fixed them
- A few DTT tests were duplicates and skipped
- Skipped the operator overload tests because we don't support overloads yet and the tests I looked at seemed sus
- I redid the whitespace on almost every single test as they were grossly inconsistent
- Skipped the preprocessor tests from gcc/clang because we're not gcc/clang
- Skipped [stdlib](https://github.com/mumencoder/dream-test-tools/tree/main/tests/dm/stdlib) tests because this PR is already huge (see: Part 2 soon(tm))
- Skipped [tree](https://github.com/mumencoder/dream-test-tools/tree/main/tests/dm/tree) tests because this PR is already huge (see: Part 2 soon(tm))

Every test was manually validated with BeeBot and/or MoMMI, so everything should(tm) work minus human error.


Content.Tests/DMProject/Broken Tests/Expression/String/multiline_whitespace.dm
```diff
@@ -0,0 +1,8 @@
+//# issue 23
+
+var/str = "A\
+		
+	B"
+
+/proc/RunTest()
+	ASSERT(str == "AB")
```
wixoaGit, line 8: This one's a little odd. This should work, but `str` appears to be null for some reason. If the declaration of `str` is moved to inside the proc then it works correctly.

Not really necessary to fix this now, though.

Content.Tests/DMProject/Broken Tests/Preprocessor/ignore_spaces_comment.dm
```diff
@@ -0,0 +1,7 @@
+
+//# issue 25
+
+/* comment */ /var/v = 5
+
+/proc/RunTest()
+	ASSERT(v == 5)
```
wixoaGit, line 7: Isn't this the same thing as `Broken Tests/Indent/whitespace_before_var1.dm`

Content.Tests/DMProject/Tests/Expression/String/multi_arg.dm
```diff
@@ -0,0 +1,19 @@
+
+//# issue 104
+ASSERT
```
wixoaGit, line 3: What's this `ASSERT` here for?

Content.Tests/DMProject/Tests/Expression/ternary_2.dm
```diff
@@ -0,0 +1,13 @@
...
+/obj/o
+	var/test = 5
+
+/proc/get_obj()
+	return (new /obj/o)
+
+/proc/RunTest()
+	var/obj/o/thing
+	var/c = (thing = get_obj()).test
+	ASSERT(c == 5)
```
wixoaGit, line 13: The name of this test is `ternary_2` but there are no uses of ternaries in here.

Content.Tests/DMProject/Tests/Preprocessor/define_var_clash2.dm
```diff
@@ -0,0 +1,15 @@
...
+
+// the / at the beginning makes the difference
+var/const/A = 5
+
+/proc/nob()
+    world.log << (B)
+
+/proc/RunTest()
+    var/const/B = 8
+    world.log << (A)
```
wixoaGit, line 14: Should be using `ASSERT()` here, even if it is supposed to be a compile error.

Content.Tests/DMProject/Tests/Statements/Control Flow/blocks_1.dm
```diff
@@ -0,0 +1,11 @@
...
+//# issue 466
+
+/proc/RunTest()
+	var/i = 0
+	label_name:
+		if(i < 1)
+			i += 1
+			goto label_name
+
+	ASSERT(i == 1)
```
wixoaGit, line 11: This would pass even if `goto label_name` never did anything. It would be better to check if the code runs at least twice.
```suggestion
		if(i < 2)
			i += 1
			goto label_name

	ASSERT(i == 2)
```

Content.Tests/DMProject/Tests/Statements/Control Flow/spawn_while_enumerating.dm
```diff
@@ -0,0 +1,10 @@
+
+//# issue 163
+
+/proc/update()
+	return
+
+/proc/RunTest()
+	for(var/i = 0; i < 2; i++)
+		spawn(0) update()
```
wixoaGit, line 9: This seems to be a misunderstanding of #163. The issue is about using `spawn()` while inside a `for (i in ...)` loop.

wixoaGit (changes requested):

wixoaGit (approved):

PR #791, opened 2022-09-16
Add DM Runtime Messages to Unit Tests

amylizzle:
image


Altoids1:
This patch already exists (somewhat confusingly) in #775: https://github.com/wixoaGit/OpenDream/pull/775/commits/94bc07ae40a102804205c5efc423477c94943222

amylizzle:
So it does. Pretty similar approaches - which one you wanna keep?

Altoids1:
Yeah, sorry about that, it probably should've been on another PR, I just assumed it'd be complete quicker. 😅 

Whichever gets merged is up to Wixoa. Please do keep this open while we wait on that.

PR #795, opened 2022-09-17
Fold constant AST expressions into strings instead of interpolating them

out-of-phaze:
This is technically not something BYOND is great at doing.
It means things like `set name = "FOO ([STRING_DEFINE])"` will no longer error, which is good; they work fine in BYOND, I'm pretty sure.
However, it also means that `var/const/foo = "FOO ([STRING_DEFINE])"` will work, which technically breaks parity with BYOND, but only in that it's _more_ permissive.

### DIFF
--- a/DMCompiler/Compiler/DM/DMParserHelper.cs
+++ b/DMCompiler/Compiler/DM/DMParserHelper.cs
@@ -143,6 +143,31 @@ namespace DMCompiler.Compiler.DM {
 
                                     if (expressionParser.Errors.Count > 0) Errors.AddRange(expressionParser.Errors);
                                     if (expressionParser.Warnings.Count > 0) Warnings.AddRange(expressionParser.Warnings);
+
+                                    if (expression is DMASTExpressionConstant)
+                                    {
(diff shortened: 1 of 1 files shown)


ike709:
Acceptable parity break IMO

Altoids1:
We're going to "parity break" a lot with having better constant-folding than BYOND. None of that will ever negatively affect the behaviour of any targetted codebase, though. It just means better performance at runtime. Don't worry about it!

Hinaichigo:
Should this fix #700?

out-of-phaze:
> Should this fix #700?

No.

out-of-phaze:
Turns out the name thing doesn't actually work and both are bugged in BYOND. Still, acceptable parity break etc.

wixoaGit:
I'm not a fan of this method of constant-folding string interpolations. The AST should strive to stay as true to the source as it can, and something like this would be better done in the string's DMExpression.

I'm going to close this since it's getting rather stale and I'd prefer this be re-done in another PR.

PR #800, opened 2022-09-21
Fix macro params stringifying incorrectly

### LINKED ISSUE #799: Macro parameters with quotes in them don't stringify properly
Using a macro like `#define OUT(X) world.log << "[#X] => [json_encode(X)]"`, then writing code like `OUT(foo("bar"))`, will output something like the following (assuming foo just returns its argument):
`foo( => "bar"`
The output in DM is `foo("bar") => "bar"`. (It might actually be `whatever => "\"bar\""`, I'm not 100% sure, but that's unrelated to the issue.)
Looks like when the argument gets literal-pasted it screws with strings; should probably escape the string contents?

out-of-phaze:
Properly escapes strings inside stringified macro arguments.
Fixes #799.

### DIFF
--- a/DMCompiler/Compiler/DMPreprocessor/DMMacro.cs
+++ b/DMCompiler/Compiler/DMPreprocessor/DMMacro.cs
@@ -49,7 +49,7 @@ namespace DMCompiler.Compiler.DMPreprocessor {
 
                             tokenTextBuilder.Append('"');
                             foreach (Token parameterToken in parameter) {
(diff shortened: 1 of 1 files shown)


Hinaichigo:
It seems that issue #700 persists even with this change.

out-of-phaze:
> It seems that #700 persists even with this change. The preprocessor output is the same as well.

This isn't a fix for #700.

wixoaGit (approved):

PR #802, opened 2022-09-21
Adds rudementary ref counting to DreamObject

EdgeLordExe:
First pass at ref-counting, i dont have the hardware to test ss13 with this, so if someone could do it and then just send me the list of errors or something it would be great.


EdgeLordExe:
guh

EdgeLordExe:
well it seems like it's working just a little too well since it's destroying some objects prematurely

EdgeLordExe:
A couple of weird things, normal procs are *mostly* working fine, the only big issues i see rn are native procs, some of them work, some of them work only sometimes, some of them break even tough they do not even USE DreamObjects but strings and such, the exceptions are thrown DM side, or rather, when those native procs break they return nothing, which is interesting.

EdgeLordExe:
ok we're down to 1 test failing for whatever reason

EdgeLordExe:
GUH FUCKING FINALLY?

wixoaGit:
This PR is very outdated and would likely need a lot of effort to bring it up to date. And with the advent of #1776, we're probably going to end up depending on C#'s garbage collector instead.

PR #807, opened 2022-09-22
Adds var/vars to /world and /client, fixes several bugs with all var/vars implementations

### LINKED ISSUE #805: "vars[<key>]" not functioning within client/proc
When compiling /vg/station, the following error occurs:
`Error at code\modules\admin\verbs\SDQL_2\SDQL_2.dm:27:7: Unknown identifier "vars"`
at:
```
20	// Assumes the variable actually exists.
21	/datum/proc/SDQL_update(var/const/var_name, var/new_value)
22		vars[var_name] = new_value
23		return 1
24
25	// Because /client isn't a subtype of /datum...
26	/client/proc/SDQL_update(var/const/var_name, var/new_value)
27		vars[var_name] = new_value
28		return 1
```
It doesn't occur on the similar line 22, so it seems that it is specific to `client`.

Altoids1:
image
Fixes #805.
## Summary
At first I thought this would just be an easy thing where I can, like, slap some stuff in DMStandard and call it a bug fixed.

However, while testing that solution, I found some bugginess in how var/vars works for all types, including /datum and everything else.

The classes that inherit from `DreamList` were not overriding some methods that procs like `json_encode` were using to find the associative values of the list, so I also fixed that.

## Changelog
- `json_encode()` is now better at encoding `vars` and `global.vars`.
- Fixed `vars` accesses not including any variables which have not been locally overwritten for the DMObject that is accessed.
- `json_encode()` no longer causes stack overflows when encoding data with cyclical or deep references.
- `/client` and `/world` now have a `vars` var.
- DreamValues of type `DreamResource` now actually attempt to encode themselves when json_encode()'d.

### DIFF
--- a/DMCompiler/DM/Visitors/DMObjectBuilder.cs
+++ b/DMCompiler/DM/Visitors/DMObjectBuilder.cs
@@ -104,7 +104,8 @@ namespace DMCompiler.DM.Visitors {
             }
             else if (varObject.HasLocalVariable(varDefinition.Name))
             {
-                DMCompiler.Error(new CompilerError(varDefinition.Location, $"Duplicate definition of var \"{varDefinition.Name}\""));
+                if(!DoesDefineSnowflakeVars(varDefinition, varObject))
+                    DMCompiler.Error(new CompilerError(varDefinition.Location, $"Duplicate definition of var \"{varDefinition.Name}\""));
                 variable = varObject.GetVariable(varDefinition.Name);
             }
             //TODO: Fix this else-if chaining once _currentObject is refactored out of DMObjectBuilder.
@@ -278,6 +279,20 @@ namespace DMCompiler.DM.Visitors {
         {
             if (varDefinition == null) return false;
             return varDefinition.IsStatic && varDefinition.Name == "vars" && varDefinition.ObjectPath == DreamPath.Root;
+        }
+
+        /// <summary>
+        /// A snowflake helper proc which allows for ignoring variable duplication in the specific case that /world or /client are inheriting from /datum,<br/>
+        /// which would normally throw an error since all of these classes have their own var/vars definition.
+        /// </summary>
+        private static bool DoesDefineSnowflakeVars(DMASTObjectVarDefinition varDefinition, DMObject varObject) {
+            if (DMCompiler.Settings.NoStandard == false)
+                if (varDefinition != null)
+                    if (varDefinition.Name == "vars")
+                        if (varDefinition.ObjectPath == DreamPath.World || varDefinition.ObjectPath == DreamPath.Client)
+                            if (varObject.IsSubtypeOf(DreamPath.Datum))
+                                return true;
+            return false;
         }
 
         /// <summary>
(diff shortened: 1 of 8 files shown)


DMCompiler/DMStandard/Types/Client.dm
```diff
@@ -2,6 +2,7 @@
 	var/list/verbs = list()
 	var/list/screen = list()
 	var/list/images = list() as opendream_unimplemented
+	var/list/vars
```
wixoaGit, line 5: Is setting `/client/parent_type = /datum` going to cause issues with duplicate vars?

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -831,42 +831,61 @@ private static DreamValue CreateValueFromJsonElement(JsonElement jsonElement)
...
-            } else if (value.TryGetValueAsPath(out var path)) {
-                return path.ToString();
-            } else if (value.TryGetValueAsDreamList(out DreamList list)) {
-                if (list.IsAssociative()) {
+            if (value.TryGetValueAsPath(out var path))
+                return HttpUtility.JavaScriptStringEncode(path.ToString());
+            if (value.TryGetValueAsDreamList(out DreamList list)) {
+                List<Object> jsonArray = new();
+                if (!beRecursive) // TODO: handle recursion while preventing infinite recursion / stack overflow
+                    return jsonArray;
```
wixoaGit, line 854: So in it's current state, this won't encode lists inside lists? There are many places in SS13 code that require this to work. TGUI for example calls `json_encode()` on a list of data to be sent to clients, which itself contains lists.

wixoaGit (changes requested):

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
Altoids1: Maybe this is a bit *too* lazy, yeah. I can try to think up a quick functional fix.

DMCompiler/DMStandard/Types/Client.dm
Altoids1: Yuhp. Only solution I can think of would be some snowflake thing in the duplicate var detection.

DMCompiler/DM/Visitors/DMObjectBuilder.cs
```diff
@@ -280,6 +281,19 @@ private static bool DoesOverrideGlobalVars(DMASTObjectVarDefinition varDefinitio
             return varDefinition.IsStatic && varDefinition.Name == "vars" && varDefinition.ObjectPath == DreamPath.Root;
         }
 
+        /// <summary>
+        /// A snowflake helper proc which allows for ignoring variable duplication in the specific case that /world or /client are inheriting from /datum,
+        /// which would normally throw an error since all of these classes have their own var/vars definition.
+        /// </summary>
+        private static bool DoesDefineSnowflakeVars(DMASTObjectVarDefinition varDefinition, DMObject varObject) {
+            if (varDefinition != null)
```
wixoaGit, line 289: This should return false if we aren't even using DMStandard
```suggestion
            if (DMCompiler.Settings.NoStandard)
                return false;
            
            if (varDefinition != null)
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #813, opened 2022-09-24
Fixes `typesof()` for non-paths

ike709:
Replicates BYOND behavior

### DIFF
--- a/OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
+++ b/OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
@@ -1854,7 +1854,9 @@ namespace OpenDreamRuntime.Procs.Native {
             DreamList list = DreamList.Create();
 
             foreach (DreamValue type in arguments.GetAllArguments()) {
-                DreamPath typePath = type.GetValueAsPath();
+                if (!type.TryGetValueAsPath(out var typePath)) {
+                    continue;
+                }
 
                 if (typePath.LastElement == "proc") {
                     DreamPath objectTypePath = typePath.AddToPath("..");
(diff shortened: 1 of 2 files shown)


OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -1854,7 +1854,9 @@ public static DreamValue NativeProc_splittext(DreamObject instance, DreamObject
             DreamList list = DreamList.Create();
 
             foreach (DreamValue type in arguments.GetAllArguments()) {
-                DreamPath typePath = type.GetValueAsPath();
+                if (!type.TryGetValueAsPath(out var typePath)) {
```
wixoaGit, line 1857: This proc also accepts DreamObjects

wixoaGit (changes requested):

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
wixoaGit: This proc apparently also takes strings like `"/obj"`. DreamLists need to be ignored too.

ike709:
It's not as optimal as possible but it should work.

out-of-phaze:
It doesn't work for things like `typesof("/datum/proc")`. Which works in DM, for some absurd reason! Gives a list of all procs defined on datum.

out-of-phaze:
> Separate issue, out of scope :^)

I literally gave you a patch that solves it

ike709:
I commented that _before_ you did :^)

SpaceManiac (approved):

Issue #290, opened 2021-08-03
`world/SetConfig()` not implemented


ike709:
http://www.byond.com/docs/ref/#/world/proc/SetConfig

amylizzle:
Closed in favour of #2194

Issue #289, opened 2021-08-03
`world/GetConfig()` not implemented


ike709:
http://www.byond.com/docs/ref/#/world/proc/GetConfig

amylizzle:
Closed in favour of #2194

PR #817, opened 2022-09-25
Implement world.GetConfig and SetConfig for environment variables

SpaceManiac:
"env" is implemented but "admin" and "ban" are not. Pursuant to #289 and #290.

To test those changes, the following test suite fixes were made:
- Reset `DMObjectTree.Globals` between compilations so references to `world` are correct
- Reset `DreamManager.Globals` between runs to match
- Add runtime when dividing by zero; Const3.dm previously passed because of global confusion
- Use relative paths to `.dm` files to make test explorer easier to read

### DIFF
--- a/OpenDreamRuntime/Procs/Native/DreamProcNativeWorld.cs
+++ b/OpenDreamRuntime/Procs/Native/DreamProcNativeWorld.cs
@@ -9,9 +9,9 @@ namespace OpenDreamRuntime.Procs.Native
     {
         [DreamProc("Export")]
         [DreamProcParameter("Addr", Type = DreamValue.DreamValueType.String)]
-        [DreamProcParameter("File", Type = DreamValue.DreamValueType.DreamObject, DefaultValue = null)]
+        [DreamProcParameter("File", Type = DreamValue.DreamValueType.DreamObject)]
         [DreamProcParameter("Persist", Type = DreamValue.DreamValueType.Float, DefaultValue = 0)]
-        [DreamProcParameter("Clients", Type = DreamValue.DreamValueType.DreamObject, DefaultValue = null)]
+        [DreamProcParameter("Clients", Type = DreamValue.DreamValueType.DreamObject)]
         public static async Task<DreamValue> NativeProc_Export(AsyncNativeProc.State state)
         {
             var addr = state.Arguments.GetArgument(0, "Addr").Stringify();
@@ -41,19 +41,19 @@ namespace OpenDreamRuntime.Procs.Native
 
         [DreamProc("GetConfig")]
         [DreamProcParameter("config_set", Type = DreamValue.DreamValueType.String)]
-        [DreamProcParameter("param", Type = DreamValue.DreamValueType.String, DefaultValue = null)]
+        [DreamProcParameter("param", Type = DreamValue.DreamValueType.String)]
         public static DreamValue NativeProc_GetConfig(DreamObject src, DreamObject usr, DreamProcArguments arguments)
         {
-            var config_set = arguments.GetArgument(0, "config_set").Stringify();
+            arguments.GetArgument(0, "config_set").TryGetValueAsString(out string config_set);
             var param = arguments.GetArgument(1, "param");
 
             switch (config_set) {
                 case "env":
                     if (param == DreamValue.Null) {
                         // DM ref says: "If no parameter is specified, a list of the names of all available parameters is returned."
-                        // but apparently it's actually just null.
+                        // but apparently it's actually just null for "env".
                         return DreamValue.Null;
-                    } else if (Environment.GetEnvironmentVariable(param.Stringify()) is string strValue) {
+                    } else if (param.TryGetValueAsString(out string paramString) && Environment.GetEnvironmentVariable(paramString) is string strValue) {
                         return new DreamValue(strValue);
                     } else {
                         return DreamValue.Null;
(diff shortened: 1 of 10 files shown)


OpenDreamRuntime/Resources/DreamResourceManager.cs
```diff
@@ -21,7 +21,7 @@ public void Initialize(string jsonPath)
 
         public void SetDirectory(string directory) {
             RootPath = directory;
-            Directory.SetCurrentDirectory(RootPath);
+            //Directory.SetCurrentDirectory(RootPath);
```
wixoaGit, line 24: This line is necessary. It was added because external DLL calls depended on this to be set in order to find/open files in the correct directory.

OpenDreamRuntime/Procs/Native/DreamProcNativeWorld.cs
```diff
@@ -38,5 +38,67 @@ public static async Task<DreamValue> NativeProc_Export(AsyncNativeProc.State sta
 
             return new DreamValue(list);
         }
+
+        [DreamProc("GetConfig")]
+        [DreamProcParameter("config_set", Type = DreamValue.DreamValueType.String)]
+        [DreamProcParameter("param", Type = DreamValue.DreamValueType.String, DefaultValue = null)]
```
wixoaGit, line 44: `DefaultValue` is null by default. It's only used when the passed argument is null, so it wouldn't make sense anyways.
```suggestion
        [DreamProcParameter("param", Type = DreamValue.DreamValueType.String)]
```

OpenDreamRuntime/Procs/Native/DreamProcNativeWorld.cs
```diff
@@ -38,5 +38,67 @@ public static async Task<DreamValue> NativeProc_Export(AsyncNativeProc.State sta
...
+                    throw new NotSupportedException("Unsupported GetConfig config_set: " + config_set);
+                default:
+                    throw new ArgumentException("Incorrect GetConfig config_set: " + config_set);
+            }
+        }
+
+        [DreamProc("SetConfig")]
+        [DreamProcParameter("config_set", Type = DreamValue.DreamValueType.String)]
+        [DreamProcParameter("param", Type = DreamValue.DreamValueType.String)]
+        [DreamProcParameter("value", Type = DreamValue.DreamValueType.String, DefaultValue = null)]
```
wixoaGit, line 78:
```suggestion
        [DreamProcParameter("value", Type = DreamValue.DreamValueType.String)]
```

OpenDreamRuntime/Procs/Native/DreamProcNativeWorld.cs
```diff
@@ -38,5 +38,67 @@ public static async Task<DreamValue> NativeProc_Export(AsyncNativeProc.State sta
 
             return new DreamValue(list);
         }
+
+        [DreamProc("GetConfig")]
+        [DreamProcParameter("config_set", Type = DreamValue.DreamValueType.String)]
+        [DreamProcParameter("param", Type = DreamValue.DreamValueType.String, DefaultValue = null)]
+        public static DreamValue NativeProc_GetConfig(DreamObject src, DreamObject usr, DreamProcArguments arguments)
+        {
+            var config_set = arguments.GetArgument(0, "config_set").Stringify();
```
wixoaGit, line 47: `Stringify()` is used to convert any type of DreamValue into a string, often to be displayed to a user (like in `world.log << ...`). Doing it this way would let me do something like this (an error in byond):
```
var/obj/O = new()
O.name = "env"
world.log << world.GetConfig(O, "ENV_VAR")
```

What you want instead is `TryGetValueAsString()` and to handle cases where you weren't passed a string. In this case it should error, as that's what byond does here.

OpenDreamRuntime/Procs/Native/DreamProcNativeWorld.cs
```diff
@@ -38,5 +38,67 @@ public static async Task<DreamValue> NativeProc_Export(AsyncNativeProc.State sta
...
+
+            switch (config_set) {
+                case "env":
+                    if (param == DreamValue.Null) {
+                        var keys = Environment.GetEnvironmentVariables().Keys;
+                        var list = DreamList.Create(keys.Count);
+                        foreach (var key in keys) {
+                            list.AddValue(new DreamValue((string)key));
+                        }
+                        return new DreamValue(list);
```
wixoaGit, line 58: `world.GetConfig("env")` is giving me `null` on byond.

OpenDreamRuntime/Procs/Native/DreamProcNativeWorld.cs
```diff
@@ -38,5 +38,67 @@ public static async Task<DreamValue> NativeProc_Export(AsyncNativeProc.State sta
...
+                    } else if (Environment.GetEnvironmentVariable(param.Stringify()) is string strValue) {
```
wixoaGit, line 59: Shouldn't use `Stringify()` here

OpenDreamRuntime/Procs/Native/DreamProcNativeWorld.cs
```diff
@@ -38,5 +38,67 @@ public static async Task<DreamValue> NativeProc_Export(AsyncNativeProc.State sta
...
+        }
+
+        [DreamProc("SetConfig")]
+        [DreamProcParameter("config_set", Type = DreamValue.DreamValueType.String)]
+        [DreamProcParameter("param", Type = DreamValue.DreamValueType.String)]
+        [DreamProcParameter("value", Type = DreamValue.DreamValueType.String, DefaultValue = null)]
+        public static DreamValue NativeProc_SetConfig(DreamObject src, DreamObject usr, DreamProcArguments arguments)
+        {
+            var config_set = arguments.GetArgument(0, "config_set").Stringify();
+            var param = arguments.GetArgument(1, "param").Stringify();
```
wixoaGit, line 82: Shouldn't use `Stringify()` here

OpenDreamRuntime/Procs/Native/DreamProcNativeWorld.cs
```diff
@@ -38,5 +38,67 @@ public static async Task<DreamValue> NativeProc_Export(AsyncNativeProc.State sta
...
+            var value = arguments.GetArgument(2, "value");
+
+            switch (config_set) {
+                case "env":
+                    if (value == DreamValue.Null) {
+                        Environment.SetEnvironmentVariable(param, null);
+                    } else {
+                        Environment.SetEnvironmentVariable(param, value.Stringify());
```
wixoaGit, line 90: Shouldn't use `Stringify()` here. Byond seems to just not set the value when `value` isn't a string.

wixoaGit (changes requested):

wixoaGit (approved):

PR #820, opened 2022-09-25
Implement regex `\l`

ike709:
It could stand to be more robust (mainly parsing an arbitrary number of preceding escapes) but it's good enough for most usecases


Altoids1:
Weird but alright

wixoaGit (approved):

PR #822, opened 2022-09-25
Fix `isloc()` not coping with null args

ike709:


Altoids1:
It's kinda weird that `TryGetValueAsDreamObject` doesn't return `false` when the object value is just `null`.

wixoaGit (approved):

PR #824, opened 2022-09-25
Fix procs on deleted objects continuing to run

### LINKED ISSUE #823: Procs on deleted objects continue running
This also causes further `System.Exception: Cannot try to get variable on a deleted object` runtimes and the like because whenever the proc on the deleted object tries to access its own object's types/procs it'll trigger that exception.

ike709:
Fixes #823

### DIFF
--- a/OpenDreamRuntime/Procs/DMProc.cs
+++ b/OpenDreamRuntime/Procs/DMProc.cs
@@ -201,13 +201,18 @@ namespace OpenDreamRuntime.Procs {
 
         protected override ProcStatus InternalResume()
         {
+            if (Instance is not null && Instance.Deleted) {
+                ReturnPools();
+                return ProcStatus.Cancelled;
+            }
+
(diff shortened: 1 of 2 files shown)


OpenDreamRuntime/Procs/DMProc.cs
```diff
@@ -201,13 +201,18 @@ public DMProcState(DMProcState other, DreamThread thread)
...
+            }
+
             while (_pc < _proc.Bytecode.Length) {
                 int opcode = _proc.Bytecode[_pc++];
 
                 var status = _opcodeHandlers[opcode].Invoke(this);
 
-                if (status != null) {
-                    if (status == ProcStatus.Returned) {
+                if (status != null ) {
```
wixoaGit, line 214:
```suggestion
                if (status != null) {
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #827, opened 2022-09-27
Implements #if, #elif, #else, and defined() for the preprocessor, fixes macro whitespace issues

### LINKED ISSUE #8: Missing preprocessor directives
Here's some missing preprocessor directives:

- [ ] #if (partially implemented)
- [ ] #elif
- [x] #warn
- [x] #error

### LINKED ISSUE #698: Whitespace prevents a macro from being recognized.
This applies to `de0b016484a68f8f752da82755522bfe07b58261`
When trying to compile `vgstation13-coders/vgstation13`, I encountered the following error:

`Error at code\modules\mob\living\say.dm:474:17: Type /mob/living does not have a proc named "isborer"`

The lines at that location are: 
```
473	if(isborer(M) || ((M in dead_mob_list) && !istype(M, /mob/new_player)))
474		if(isborer (M)) //for borers that are IN CONTROL
475			B = M //why it no typecast
476			if(B.controlling)
477				M = B.host
```
`isborer` is defined as:
`#define isborer(A) istype(A, /mob/living/simple_animal/borer)`

`isborer`in line 473 works fine but it causes an error in line 474, seemingly because of the whitespace.

Altoids1:
image
Finally, fixes #8.
(and fixes #698.)

## Summary
This is achieved primarily by adding some more state to `DMPreprocessor` as its parsing preproc tokens, as well as the implementation of `DMPreprocessorParser`.

I elected to give the preproc its own parser, instead of trying to inherit in a direct way from DMParser, because:
1. The grammar for preproc expressions is very simple compared to DM expressions.
2. It is basically stateless (this could be made static if we want)
3. Some symbols allowed in DM expressions may not be allowed at the same grammatical context for preproc ones (think strings, bitwise operations, initializing new lists, etc)
4. Not being a DMParser allows for a different system of evaluation, where the float values of expressions are determined as parsing is done, rather than evaluated at the end via a TryAsConstant() call. This reduces the complexity of the preproc parser and avoids some missing const-evaluation within the `Constant` expression type.

## Changelog
- `#elif` is now a recognized preprocessor directive; it was not before.
- #if, #elif, and #else are now all functional.
- Fixes macro functors not working if a whitespace exists between the macro's name and its parentheses.
- `defined()` now exists as a functor in preprocessor expressions.
- Attempting to create a macro named `defined` now causes an error.
- All now-passing preproc tests have been moved from Broken Tests into the main Test suite.
- The preprocessor's lexer now emits descriptive warnings when the user tries to use an improperly-capitalized macro directive.


DMCompiler/Compiler/DMPreprocessor/DMPreprocessor.cs
```diff
@@ -22,6 +30,14 @@ public class DMPreprocessor : IEnumerable<Token> {
...
             { "__FILE__", new DMMacroFile() }
         };
+        /// <summary>
+        /// This stores previous evaluations of if-directives that have yet to find their #endif.<br/>
+        /// We do this so that we can A.) Detect whether an #else or #endif is valid and B.) Remember what to do when we find that #else.
+        /// A null value indicates the last directive found was an #else that's waiting for an #endif.
+        /// </summary>
+        private Stack<bool?> _lastIfEvaluations = new(); 
+        private Location _lastSeenIf = Location.Unknown; // used by the errors emitted for when the above two vars are non-zero at exit
+        private Token _lastToken; // The last token emitted via GetNextToken().
```
wixoaGit, line 40: Why does this need to exist? It doesn't appear to actually be used anywhere.

DMCompiler/Compiler/DMPreprocessor/DMPreprocessorParser.cs
```diff
@@ -0,0 +1,294 @@
...
+    PowerExp ::= UnaryExp [** PowerExp]
+    UnaryExp ::= {!}SignExp
+    SignExp ::= [-]Primary
+    Primary ::= '(' Expression ')' | DefinedExp | Constant
+    DefinedExp ::= 'defined(' MacroIdentifier ')'
+    Constant ::= Integer | Float | Null
+    */
+
+    /// <summary>
+    /// An extremely simple parser that acts on Preproc tokens, <br/>
```
wixoaGit, line 31: This doesn't act only on preproc tokens

DMCompiler/Compiler/DMPreprocessor/DMPreprocessorParser.cs
```diff
@@ -0,0 +1,294 @@
...
+    /// held separate from DMParser because of slightly different behaviour and far simpler implementation.
+    /// </summary>
+    internal static class DMPreprocessorParser {
```
wixoaGit, line 34: Is there a reason this can't extend the `Parser<T>` class our other parsers use?

DMCompiler/Compiler/DMPreprocessor/DMPreprocessorParser.cs
```diff
@@ -0,0 +1,294 @@
...
+            foreach (TokenType type in types) {
+                if (Current()?.Type == type) {
+                    Advance();
+                    return true;
+                }
+            }
+            return false;
+        }
+        private static void Error(string msg) {
+            DMCompiler.Error(new CompilerError(Current()?.Location ?? Location.Unknown, msg));
```
wixoaGit, line 76: Use the `Error(Location loc, string message)` version instead.
```suggestion
            DMCompiler.Error(Current()?.Location ?? Location.Unknown, msg);
```

wixoaGit (changes requested):

DMCompiler/Compiler/DMPreprocessor/DMPreprocessorParser.cs
Altoids1: Most practically, I want to make DMPreprocessorParser static eventually and a static object cannot inherit from a non-static (this would be the only stateless parser in the codebase)

Second mostly practically, DMPP doesn't really make use of any of the code available in DMParser or Parser<>, methods of the same name would need to be completely overridden, defeating the purpose of inheritance as a way of reducing code size & complexity.

If it's annoying that it's the odd one out, it could make sense to call it the DMPreprocessorEvaluator instead or something, since it is doing parsing and evaluation in one stroke.

wixoaGit (approved):

PR #756, opened 2022-07-14
Build the global proc list before processing statements

ike709:

Issue #803, opened 2022-09-21
'Unknown identifier' with /var/const


Hinaichigo:
Compiling /vg/station, an error occurs:
`Error at code\datums\supplypacks.dm:337:40: Unknown identifier "access_engine_minor"`
here:
```
330	/datum/supply_packs/sacidtank
331		name = "Sulphuric acid tank"
332		contains = list(/obj/structure/reagent_dispensers/sacid)
333		cost = 15
334		containertype = /obj/structure/closet/crate/secure/large/reinforced/shard/empty
335		containername = "sulphuric acid tank crate"
336		group = "Supplies"
337		one_access = list(access_engine_minor, access_science)
```
`access_engine_minor` is defined in `/code/game/jobs/access.dm` as:
`/var/const/access_engine_minor = 11	// Engineering Foyer`
This error only started occurring fairly recently.

ike709:
@Altoids1 Seems like something your PR should have fixed?

Altoids1:
god

Altoids1:
I got the same error on a version of OD with the const PR rolled back; might be a novel error caused by changes in VGstation codebase itself. This is not a regression, just something I didn't fix.

Stuff like this probably could've been in the const PR, but was not, since late evaluation of expressions is completely unimplemented. The concept of late-evaluation did not even meaningfully exist in the codebase prior.

For the record, the const PR fixed late evaluation of the *left* side of an override assignment. The *right* side remains a TODO.

Hinaichigo:
I don't think this is due to an underlying change in the /vg/station codebase; the format of the access-related stuff involving `/code/game/jobs/access.dm` has been the same for years. Rather I think it's likely due to a change that occurred in how `/var/const` are processed some time over the last few months (not as recent as @Altoids1's recent changes), because I remember trying to compile after pulling the latest version of OpenDream and there were more errors than before the pull. At the time I thought maybe it was just being more verbose about errors that previously had been silently occurring, or something. I'm not sure exactly when this started happening but I may be able to check a rough date in a little bit.

Hinaichigo:
I'm not exactly sure when it happened but I think it may have been around the time #706 was fixed, as I remember that error stopped appearing but then this error appeared. Could be wrong though.

PR #829, opened 2022-09-27
Register global procs and const or global vars before processing statements

out-of-phaze:
Adds a pass to register global procs (procs defined on root) so that they can be used in global var definitions made prior to the proc's definition. Adapted from #756.
Adds a pass to register global and const vars, both on root and all other types. In this pass, global vars are only defined, while const vars also have their values assigned.

This fixes over 131 errors on a modern Bay fork I'm working on.

### DIFF
--- a/DMCompiler/DM/Visitors/DMObjectBuilder.cs
+++ b/DMCompiler/DM/Visitors/DMObjectBuilder.cs
@@ -52,27 +52,27 @@ namespace DMCompiler.DM.Visitors {
         }
 
         private static void ProcessFile(DMASTFile file) {
-            RegisterGlobals(file.BlockInner, DMObjectTree.Root);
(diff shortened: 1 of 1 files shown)


Hinaichigo:
Nice. This should fix about 20 or 1/6 of the remaining errors for /vg/station as well.

Hinaichigo:
> Adds a pass to register global and const vars, both on root and all other types. In this pass, global vars are only defined, while const vars also have their values assigned.

Would this also affect #803?

out-of-phaze:
> > Adds a pass to register global and const vars, both on root and all other types. In this pass, global vars are only defined, while const vars also have their values assigned.
> 
> Would this also affect #803?

Yes, but this fix breaks a bunch of other stuff. Going to pursue other potential solutions for this.

PR #834, opened 2022-10-01
Support const-eval for `==` and `!=`

ike709:
I haven't tested every single comparison but basic testing shows that it works. This PR is required for one of the preproc tests to pass in https://github.com/wixoaGit/OpenDream/pull/827

### DIFF
--- a/DMCompiler/DM/Expressions/Binary.cs
+++ b/DMCompiler/DM/Expressions/Binary.cs
@@ -1,6 +1,7 @@
 using OpenDreamShared.Compiler;
 using OpenDreamShared.Dream;
 using OpenDreamShared.Dream.Procs;
+using Robust.Shared.Maths;
 
 namespace DMCompiler.DM.Expressions {
     abstract class BinaryOp : DMExpression {
@@ -295,7 +296,9 @@ namespace DMCompiler.DM.Expressions {
                 return false;
             }
 
-            constant = lhs.NotEquals(rhs);
+            constant = lhs.Equals(rhs);
+            // We're always going 0 -> 1 or 1 -> 0 so we can do this to avoid branching
+            constant.Value = Math.Abs(constant.Value - 1);
             return true;
         }
     }
--- a/DMCompiler/DM/Expressions/Binary.cs
+++ b/DMCompiler/DM/Expressions/Binary.cs
@@ -261,6 +261,16 @@ namespace DMCompiler.DM.Expressions {
         public Equal(Location location, DMExpression lhs, DMExpression rhs)
             : base(location, lhs, rhs) { }
 
+        public override bool TryAsConstant(out Constant constant) {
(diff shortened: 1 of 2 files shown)


DMCompiler/DM/Expressions/Constant.cs
```diff
@@ -3,6 +3,7 @@
 using OpenDreamShared.Json;
 using System;
 using System.Collections.Generic;
+using Robust.Shared.Localization;
```
ike709, line 6:
```suggestion
```

Altoids1: what

DMCompiler/DM/Expressions/Constant.cs
```diff
@@ -223,6 +248,22 @@ class Number : Constant {
...
+            return new Number(Location, 1); // true
+        }
+
+        public override Constant NotEquals(Constant rhs) {
+            if (rhs is not Number rhsNum || !MathHelper.CloseTo(rhsNum.Value, Value)) {
+                return new Number(Location, 1); // true
+            }
+
+            return new Number(Location, 0); // false
+        }
```
ike709, line 266:
```suggestion
        public override Constant Equals(Constant rhs) {
            if (rhs is not Number rhsNum || !MathHelper.CloseTo(rhsNum.Value, Value)) {
                return new Number(Location, 0); // false
            }

            return new Number(Location, 1); // true
        }

        public override Constant NotEquals(Constant rhs) {
            if (rhs is not Number rhsNum || MathHelper.CloseTo(rhsNum.Value, Value)) {
                return new Number(Location, 1); // true
            }

            return new Number(Location, 0); // false
        }
```

Altoids1:
No longer a dependency of the preproc PR (but you should merge it anyways since Ike was very nice to do this for me)

DMCompiler/DM/Expressions/Binary.cs
```diff
@@ -278,6 +288,16 @@ public NotEqual(Location location, DMExpression lhs, DMExpression rhs)
...
             proc.NotEqual();
         }
+
+        public override bool TryAsConstant(out Constant constant) {
+            if (!LHS.TryAsConstant(out var lhs) || !RHS.TryAsConstant(out var rhs)) {
+                constant = null;
+                return false;
+            }
+
+            constant = lhs.NotEquals(rhs);
```
wixoaGit, line 298: Would there ever a case where `NotEquals()` isn't just `!Equals()`?

wixoaGit (changes requested):

DMCompiler/DM/Expressions/Binary.cs
wixoaGit: Might just be able to use `IsTruthy()` for this too.

ike709:
That's what I get for trying to webedit.

Altoids1:
Eclipsed by #885.

PR #835, opened 2022-10-02
StdLib Unit Tests

ike709:
This ports all of the StdLib unit tests from DTT.

Same deal as https://github.com/wixoaGit/OpenDream/pull/786


Content.Tests/DMProject/Tests/Stdlib/Array/assign_blank.dm
```diff
@@ -0,0 +1,9 @@
+// RUNTIME ERROR
+
+/obj
+	var/thing[]
+
+/proc/RunTest()
+	var/obj/o = new
+	ASSERT(isnull(o.thing))
```
wixoaGit, line 8: This would lead to the test passing if `o.thing` isn't null, despite it _should_ being null.

wixoaGit (changes requested):

wixoaGit (approved):

PR #841, opened 2022-10-08
Implement accessing proc properties

ike709:
Not sure if this is the best way to do it but it seems to work.


OpenDreamRuntime/DreamThread.cs
```diff
@@ -59,6 +59,21 @@ public abstract class DreamProc {
             context.PushProcState(state);
             return context.Resume();
         }
+
+        public DreamValue GetField(string field) {
+            switch (field) {
```
ike709, line 64:
```suggestion
            // TODO: Figure out what byond does when these are null
            switch (field) {
```

OpenDreamRuntime/DreamThread.cs
```diff
@@ -61,6 +61,20 @@ public abstract class DreamProc {
...
+                    return new DreamValue(VerbName);
+                case "category":
+                    return new DreamValue(VerbCategory);
+                case "description":
+                    return new DreamValue(VerbDesc);
+                case "invisibility":
+                    return new DreamValue(Invisibility);
+                default:
+                    throw new Exception($"Cannot get field \"{field}\" from {OwningType.ToString()}.{Name}()");
+
```
SpaceManiac, line 77: You are missing two closing `}` here

ike709:
I forgot this PR was failing CI.

SpaceManiac (approved):

SpaceManiac:
possible future notes:
- /tg/ also uses "hidden" since https://github.com/tgstation/tgstation/pull/53112, but I haven't checked whether this genuinely works or not, because
- BYOND returns `null` instead of erroring at runtime if a bad setting is accessed, although it's probably best for OD to error

wixoaGit (approved):

Issue #833, opened 2022-10-01
Comparison Operators Don't Support Constant Folding


ike709:
- [ ] `>=`
- [ ] `<=`
- [ ] `<`
- [ ] `>`
- [ ] `==`
- [ ] `!=`

See `Binary.cs` and `Constant.cs`

wixoaGit:
Duplicate of #828

ike709:
This one has a checklist

ZeWaka:
This one also needs `~=` added to the checklist

ike709:
not anymore

Hinaichigo:
What about `<>`? I thought I remember someone mentioning that recently, but I'm not sure where.

ZeWaka:
> What about `<>`? I thought I remember someone mentioning that recently, but I'm not sure where.
`<>` is just an alias for `!=` so that should be done in the parser

Issue #828, opened 2022-09-27
Comparison operators do not yet have any constant folding


Altoids1:
This is preventing #827's preprocessor from handling these operators correctly.

This is caused by them having missing `TryAsConstant()` overrides.

ike709:
Specifically just `==` and `!=`?

Altoids1:
I think all inequality operations. ==, <=, <, >, !=, and ~=.

wixoaGit:
Closing in favor of #833

PR #842, opened 2022-10-10
Const-eval for comparisons (`<`,`<=`,`>`,`>=`)

ZeWaka:
Title, related to #833 and related to #828

also left shift expressions weren't getting const-evaled

image


ike709:
Fixes https://github.com/wixoaGit/OpenDream/issues/833 and fixes https://github.com/wixoaGit/OpenDream/issues/828

DMCompiler/DM/Expressions/Binary.cs
```diff
@@ -163,7 +163,13 @@ public LeftShift(Location location, DMExpression lhs, DMExpression rhs)
 
         public override void EmitPushValue(DMObject dmObject, DMProc proc) {
             LHS.EmitPushValue(dmObject, proc);
-            RHS.EmitPushValue(dmObject, proc);
+
+            if (RHS.TryAsConstant(out Constant constValue)) { // RHS could be a const-evaluable expr
+                constValue.EmitPushValue(dmObject, proc);
+            } else {
+                RHS.EmitPushValue(dmObject, proc);
+            }
```
ZeWaka, line 171: This is somewhat of a weird thing. right now, our constant system isn't set up where we can const-eval just the rhs of an expression (see: const << const, foo << const, foo << foo) it seems in the TryAsConstant tree - so we do it here in the EmitPushValue even though i think the proper behavior would be to TryConstant in `ProcessStatementExpression()`

ZeWaka: for the logical conclusion, see #885

ZeWaka: to be clear, this bit should still stay in the for the case of where the rhs can be constant but the lhs can't be - the TryAsConstant needs both to be true

see the case of `world << (5 + 5)`

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -2038,6 +2050,9 @@ private static void HandleSuffixPronoun(ref StringBuilder formattedString, ReadO
                         if (second.TryGetValueAsString(out var s)) return s != "";
                         if (second.Value == null) return false;
                     }
+                    if (second.Value == null) {
```
ZeWaka, line 2053: I forget why I put this instead of DreamValue.null

ZeWaka: Wait the above line means someone else added this as well and it was a bad merge methinks

PaulRitter (approved):

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -1989,6 +1988,15 @@ private static void HandleSuffixPronoun(ref StringBuilder formattedString, ReadO
...
 
+            if (first == DreamValue.Null) {
+                // nulls are equal to each other
+                if (second == DreamValue.Null)
+                    return true;
+                // null != 0 or ""
+                if (second.Type is DreamValue.DreamValueType.Float or DreamValue.DreamValueType.String) {
+                    return false;
+                }
+            }
```
wixoaGit, line 1999: This is already done in the above switch inside `case DreamValue.DreamValueType.DreamObject:`

DMCompiler/DM/Expressions/Constant.cs
```diff
@@ -230,6 +268,29 @@ class Number : Constant {
...
+
+            return new Number(Location, 0); // false
+        }
+
+        public override Constant LessThanOrEqual(Constant rhs) {
+            if (rhs.IsTruthy()) {
+                return new Number(Location, 1); // true
+            }
+            return new Number(Location, 0); // false
+        }
```
wixoaGit, line 293: These aren't actually checking if the Number's value is greater/less than rhs?

DMCompiler/DM/Expressions/Constant.cs
```diff
@@ -96,6 +112,28 @@ class Null : Constant {
             json = null;
             return true;
         }
+
+        public override Constant GreaterThan(Constant rhs) {
+            return new Number(Location, 0); // false
```
wixoaGit, line 117: `null > -1` is true

DMCompiler/DM/Expressions/Constant.cs
```diff
@@ -96,6 +112,28 @@ class Null : Constant {
...
+        public override Constant LessThan(Constant rhs) {
+            if (rhs.IsTruthy()) {
+                return new Number(Location, 1); // true
+            }
+            return new Number(Location, 0); // false
+        }
+
+        public override Constant LessThanOrEqual(Constant rhs) {
+            return new Number(Location, 1); // true
+        }
```
wixoaGit, line 136: Pretty sure these should all be comparing against `0`

DMCompiler/DM/Expressions/Binary.cs
wixoaGit: I don't think it should, #885 would lead both lhs and rhs to go through their own constant folding pass.

wixoaGit (changes requested):

DMCompiler/DM/Expressions/Binary.cs
wixoaGit: I don't think it should, #885 would lead both lhs and rhs to go through their own constant folding pass.

DMCompiler/DM/Expressions/Constant.cs
ZeWaka: Nice catch, didn't have negative numbers for the unit test. Nor a larger number.

wixoaGit (approved):

PR #844, opened 2022-10-12
Removes some unnecessary/off-parity double-to-float conversions in the code

Altoids1:
## Summary
Previously, our exponent implementation would use the double-precision version of `Pow()`, despite this being a single-precision language. That has been fixed.

NOTE that our trig functions (plus `/proc/log()` and so on) still make use of the double-precision variants. **This is intentional,** as BYOND has the same bug and changing this literally breaks a few unit tests. I put a warning near where the trig functions are defined to stop anyone else who may think of the same thing.

## Changelog
- Fixed some unnecessarily-precise floating point operations throughout the codebase.

### DIFF
--- a/DMCompiler/DM/Visitors/DMASTSimplifier.cs
+++ b/DMCompiler/DM/Visitors/DMASTSimplifier.cs
@@ -461,7 +461,7 @@ namespace DMCompiler.DM.Visitors {
                 DMASTConstantInteger aInteger = power.A as DMASTConstantInteger;
                 DMASTConstantInteger bInteger = power.B as DMASTConstantInteger;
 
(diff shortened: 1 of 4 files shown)


OpenDreamClient/Input/ClickMapManager.cs
```diff
@@ -64,7 +64,7 @@ internal sealed class ClickMap {
                 var width = (int)atlas.SubRegion.Width;
                 var height = (int)atlas.SubRegion.Height;
 
-                var dataSize = (int)Math.Ceiling(width * height / 8f);
+                var dataSize = (int)MathF.Ceiling(width * height / 8f);
```
wixoaGit, line 67: This line has nothing to with the DM runtime

wixoaGit (changes requested):

wixoaGit (approved):

PR #845, opened 2022-10-12
Adds operator~= and proper json_encode()ing for /matrix

Altoids1:
## Summary

Here, `operator~=` was implemented by rewriting things a little so that equivalence is handled within MetaObjects rather than relying on snowflake behaviour hard-wired into the operator itself. This should allow for easier support for operator overriding whenever we get that implemented.

I also went ahead and added correct `json_encode()`ing for /matrix.

Adding this new operator means I could also toss out a weird helper proc lying in our test suite, since this operator acts equivalently (heh) to it.

## Changelog 
- Added support for the ~= operator for `/matrix`s.
- Using ~= instead of == on values that lack any special behaviour for that operator no longer causes an unimplemented exception.
- Fixed json_encode() not encoding /matrix values correctly.
- Added more unit testing for ~= and ~! operators.
- Modified the unit tests for /matrix so it uses ~= instead of a weird helper function hidden in the test suite.

### DIFF
--- a/Content.Tests/DMProject/Tests/Operators/equivalence.dm
+++ b/Content.Tests/DMProject/Tests/Operators/equivalence.dm
@@ -29,7 +29,14 @@
 	ASSERT((m3 ~= m2) == FALSE)
 	ASSERT((m3 ~= m3) == TRUE)
 
+	var/matrix/m4 = matrix(1,2,3,4,5,6)
+	ASSERT((m1 ~= m4) == TRUE)
+	ASSERT((m2 ~= m4) == FALSE)
+	ASSERT((m3 ~= m4) == FALSE)
+	ASSERT((m4 ~= m4) == TRUE)
+
 	ASSERT(("apple" ~= m1) == FALSE)
 	ASSERT(("apple" ~! m1))
+	ASSERT(m1 ~! "apple")
 	ASSERT(m1 ~! m2)
 	ASSERT(1 ~= 1)
--- a/Content.Tests/DMProject/Tests/Operators/equivalence.dm
+++ b/Content.Tests/DMProject/Tests/Operators/equivalence.dm
@@ -14,3 +14,22 @@
 	ASSERT((l1 ~! l2) == TRUE)
 	ASSERT((l1 ~! l3) == TRUE)
 	ASSERT((l1 ~! l4) == TRUE)
+
+	var/matrix/m1 = matrix(1,2,3,4,5,6)
+	var/matrix/m2 = matrix(-1,-2,-3,-4,-5,6)
+	var/matrix/m3 = matrix(1,2,3,4,5,6)
+	m3.a = "poop"
+	ASSERT((m1 ~= m1) == TRUE)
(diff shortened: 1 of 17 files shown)


OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectMatrix.cs
```diff
@@ -68,5 +95,19 @@ sealed class DreamMetaObjectMatrix : IDreamMetaObject {
...
+                string elements = "abcdef";
+                for (int i = 0; i < elements.Length; i++) {
+                    left.GetVariable(elements[i].ToString()).TryGetValueAsFloat(out var leftValue); // sets leftValue to 0 if this isn't a float
+                    right.GetVariable(elements[i].ToString()).TryGetValueAsFloat(out var rightValue); // ditto
+                    if (leftValue != rightValue)
+                        return DreamValue.False;
+                }
+                return DreamValue.True;
+            }
+            return a.Equals(b) ? DreamValue.True : DreamValue.False;
```
wixoaGit, line 110: This will never be true because reaching this line means `b` is not a matrix, while `a` will always be.

OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectList.cs
```diff
@@ -145,6 +145,20 @@ public DreamValue OnVariableGet(DreamObject dreamObject, string varName, DreamVa
...
+                if (firstList.GetLength() != secondList.GetLength()) return DreamValue.False;
+                var firstValues = firstList.GetValues();
+                var secondValues = secondList.GetValues();
+                for (var i = 0; i < firstValues.Count; i++) {
+                    if (!firstValues[i].Equals(secondValues[i])) return DreamValue.False;
+                }
+
+                return DreamValue.True;
+            }
+            return a.Equals(b) ? DreamValue.True : DreamValue.False;
```
wixoaGit, line 159: Same here

Content.Tests/DMProject/Tests/Operators/equivalence.dm
```diff
@@ -14,3 +14,22 @@
...
+	m3.a = "poop"
+	ASSERT((m1 ~= m1) == TRUE)
+	ASSERT((m1 ~= m2) == FALSE)
+	ASSERT((m1 ~= m3) == FALSE)
+	ASSERT((m2 ~= m1) == FALSE)
+	ASSERT((m2 ~= m2) == TRUE)
+	ASSERT((m2 ~= m3) == FALSE)
+	ASSERT((m3 ~= m1) == FALSE)
+	ASSERT((m3 ~= m2) == FALSE)
+	ASSERT((m3 ~= m3) == TRUE)
```
wixoaGit, line 30: None of these are testing that two different matrices with the same values are considered equivalent. That should be added.

wixoaGit (changes requested):

wixoaGit (approved):

PR #848, opened 2022-10-19
Fixes macro calls always consuming an extra token whenever parsed

### LINKED ISSUE #846: Whitespace after invalid parameter-macro usage is ignored
A change in #827 intended to fix #698 ended up adding a new bug in the preprocessing of whitespace.

```
#define TEST(x)
/world/New()
	..()
	var/TEST = new /obj()
	var/t = TEST .type
```

`TEST .type` should not be valid here, but the preprocessor is throwing away the space during its search for `TEST(x)`'s arguments.

[Relevant line](https://github.com/wixoaGit/OpenDream/blob/ad1578ed2699ac1cd42ce4812278c303fac9929f/DMCompiler/Compiler/DMPreprocessor/DMPreprocessor.cs#L575)

Altoids1:
Fixes #846.

## Summary
This was a huge and confusing rabbit hole for me that took up an unhealthy portion of my evening, but here you go.

Added a unit test based on 846's snippet so we don't have to deal with this again.

## Changelog

- Macro calls no longer consume an extra token whenever parsed.

### DIFF
--- a/DMCompiler/Compiler/DMPreprocessor/DMPreprocessor.cs
+++ b/DMCompiler/Compiler/DMPreprocessor/DMPreprocessor.cs
@@ -523,6 +523,12 @@ namespace DMCompiler.Compiler.DMPreprocessor {
             _unprocessedTokens.Push(token);
         }
 
+        /// <remarks>
+        /// WARNING: Do not call this with the <see langword="true"/> argument <br/>
+        /// unless you are completely sure that the clobbered whitespace will NEVER have any grammatical significance <br/>
+        /// neither here in the preprocessor, nor in any other parsing pass! <br/><br/>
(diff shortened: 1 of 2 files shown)


DMCompiler/Compiler/DMPreprocessor/DMPreprocessor.cs
```diff
@@ -523,6 +523,11 @@ private List<Token> GetLineOfTokens()
             _unprocessedTokens.Push(token);
         }
 
+        /// <remarks>
+        /// WARNING: Do not call this with the <see langword="true"/> argument <br/>
+        /// unless you are completely sure that the clobbered whitespace will NEVER have any grammatical significance <br/>
+        /// neither here in the preprocessor, nor any other parsing pass!
+        /// </remarks>
```
ike709, line 530: You should ref `CheckForTokenIgnoringWhitespace()` here somewhere

wixoaGit (approved):

PR #850, opened 2022-10-22
Fixes float stringification

### LINKED ISSUE #843: Floats behave slightly funny on OD
Byond seems to not have rounding issues unlike OD
<br>
test code
```dm
var/a = 0
for(var/i in 1 to 21)
    world.log << a
    a += 0.05
```
<br>

```dm
var/a = 0
for(var/i in 1 to 10001)
    a += 0.0001
for(var/i in 1 to 10001)
    a -= 0.0001
world.log << a * 1000000000
```
Results in -50.3933 on byond & -50.393284 on OD
So it seems like the string conversion truncates the floating point numbers

<br>

OD
image

Byond
image

redmoogle:
Fixes #843

should fully replicate byond


Altoids1 (changes requested):
You should add some unit tests for this. :)

redmoogle:
> This pull request has conflicts, please resolve those before we can evaluate the pull request.

You mother f u c k e r

wixoaGit:
Running the unit test fails in both MoMMI and my local version of Dream Maker. I think the behavior here is different depending on your OS?

redmoogle:
> Running the unit test fails in both MoMMI and my local version of Dream Maker. I think the behavior here is different depending on your OS?

CultureInfo?

wixoaGit (approved):

PR #871, opened 2022-10-26
Implement \s and \th macros

itsmeow:
Implements the \s and \th macros, with the only difference being that \th does not confirm it is directly after an expression and not replacing non-floats with 0.

### DIFF
--- a/DMCompiler/Compiler/DM/DMParserHelper.cs
+++ b/DMCompiler/Compiler/DM/DMParserHelper.cs
@@ -176,7 +176,6 @@ namespace DMCompiler.Compiler.DM {
                                 }
                                 i--;
 
-                                bool unimplemented = false;
                                 bool skipSpaces = false;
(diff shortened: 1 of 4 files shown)


DMCompiler/Compiler/DM/DMParserHelper.cs
```diff
@@ -283,13 +283,12 @@ private DMASTExpression ExpressionFromString(Token constantToken)
                                     //Plurals, ordinals, etc
                                     //(things that hug, as a suffix, the [] that they reference)
                                     case "s":
-                                        unimplemented = true;
```
wixoaGit, line 286: These are the last two uses of `unimplemented`. This variable and check below should be removed entirely.

wixoaGit (changes requested):

wixoaGit (approved):

PR #873, opened 2022-10-26
Replace some GetValueAs with TryGetValueAs

itsmeow:
This replaces some calls that could be optional. There needs to be a `MustGetValueAs` for cases where it is guaranteed and should error otherwise, but this handles some cases where it could fail. The problem is figuring out what is supposed to happen in these cases is not very easy.

### DIFF
--- a/OpenDreamRuntime/DreamConnection.cs
+++ b/OpenDreamRuntime/DreamConnection.cs
@@ -83,13 +83,16 @@ namespace OpenDreamRuntime
             _availableVerbs.Clear();
             List<(string, string, string)>? verbs = null;
 
-            if (MobDreamObject != null)
+            if (MobDreamObject != null && MobDreamObject.GetVariable("verbs").TryGetValueAsDreamList(out var mobVerbPathsD))
             {
-                List<DreamValue> mobVerbPaths = MobDreamObject.GetVariable("verbs").GetValueAsDreamList().GetValues();
+                List<DreamValue> mobVerbPaths = mobVerbPathsD.GetValues();
                 verbs = new List<(string, string, string)>(mobVerbPaths.Count);
                 foreach (DreamValue mobVerbPath in mobVerbPaths)
                 {
-                    DreamPath path = mobVerbPath.GetValueAsPath();
+                    if (!mobVerbPath.TryGetValueAsPath(out var path))
+                    {
+                        continue;
+                    }
                     if (path.LastElement is null) continue;
                     var proc = MobDreamObject.GetProc(path.LastElement);
                     _availableVerbs.Add(path.LastElement, proc);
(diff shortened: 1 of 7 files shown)


OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectWorld.cs
```diff
@@ -38,7 +47,7 @@ sealed class DreamMetaObjectWorld : IDreamMetaObject {
             }
 
             DreamValue view = dreamObject.ObjectDefinition.Variables["view"];
-            if (view.TryGetValueAsString(out string viewString)) {
+            if (view.TryGetValueAsString(out var viewString) && viewString != null) {
```
wixoaGit, line 50: `TryGetValueAsString()` should never give you null if it returns true.
```suggestion
            if (view.TryGetValueAsString(out var viewString)) {
```

OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectClient.cs
```diff
@@ -34,22 +33,22 @@ sealed class DreamMetaObjectClient : IDreamMetaObject {
...
+                        Runtime.StateManager.AddClientPerspectiveDelta(ckey, (ClientPerspective) perspective);
+                    }*/
                     break;
                 }
                 case "mob": {
-                    DreamConnection connection = _dreamManager.GetConnectionFromClient(dreamObject);
-
-                    connection.MobDreamObject = value.GetValueAsDreamObject();
+                    DreamConnection? connection = _dreamManager.GetConnectionFromClient(dreamObject);
+                    if (connection != null && value.TryGetValueAsDreamObject(out var valueDreamObject) && valueDreamObject != null) {
```
wixoaGit, line 49: `client.mob = null` is a valid operation.
```suggestion
                    if (connection != null && value.TryGetValueAsDreamObject(out var valueDreamObject)) {
```

OpenDreamRuntime/DreamConnection.cs
```diff
@@ -210,30 +213,29 @@ public void HandleMsgPromptResponse(MsgPromptResponse message)
 
 
         public void OutputDreamValue(DreamValue value) {
-            if (value.Type == DreamValue.DreamValueType.DreamObject) {
-                DreamObject outputObject = value.GetValueAsDreamObject();
-
+            if (value.Type == DreamValue.DreamValueType.DreamObject && value.TryGetValueAsDreamObject(out var outputObject)) {
```
wixoaGit, line 216: `TryGetValueAsDreamObject()` already does the first check for you.
```suggestion
            if (value.TryGetValueAsDreamObject(out var outputObject)) {
```

OpenDreamRuntime/DreamConnection.cs
```diff
@@ -210,30 +213,29 @@ public void HandleMsgPromptResponse(MsgPromptResponse message)
...
-
-                    if (file.Type == DreamValue.DreamValueType.String || file == DreamValue.Null) {
-                        msg.File = (string)file.Value;
-                    } else if (file.TryGetValueAsDreamResource(out DreamResource resource)) {
-                        msg.File = resource.ResourcePath;
-                    } else {
-                        throw new ArgumentException("Cannot output " + value, nameof(value));
-                    }
+                    if (outputObject.GetVariable("channel").TryGetValueAsInteger(out var channelD)
+                        && outputObject.GetVariable("volume").TryGetValueAsInteger(out var volumeD)) {
```
wixoaGit, line 219: This would lead to a sound not playing if their `channel` or `volume` var was invalid. Some testing in byond shows that they still play, and both have a default of `0`. You can just remove the if and leave it as these two lines:
```
outputObject.GetVariable("channel").TryGetValueAsInteger(out var channelD);
outputObject.GetVariable("volume").TryGetValueAsInteger(out var volumeD);
```
`TryGetValueAsInteger()` defaults to setting the var to 0 if it wasn't an integer.

wixoaGit (changes requested):

wixoaGit:
I'm closing this due to inactivity.

PR #874, opened 2022-10-27
Fixes \th interpolation

redmoogle:
For some mystical byond comprehension reason byond will convert 4.52 to 4. should fix that without creating any extra bugs.

### DIFF
--- a/OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
+++ b/OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
@@ -349,7 +349,7 @@ namespace OpenDreamRuntime.Procs {
                             /// For some mystical reason byond converts \th to integers
                             interps[prevInterpIndex].TryGetValueAsFloat(out var temp);
                             /// This is slightly hacky but the only reliable way I know how to replace the number
-                            formattedString.Remove(formattedString.Length - temp.ToString().Length, temp.ToString().Length);
+                            formattedString.Length -= temp.ToString().Length;
                             formattedString.Append(ordinalNumber);
                             switch (ordinalNumber) {
                                 case 1:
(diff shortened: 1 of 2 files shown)


Altoids1:
Weird!

itsmeow:
cursed

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -344,7 +344,13 @@ private static void HandleSuffixPronoun(ref StringBuilder formattedString, ReadO
                         continue;
                     case StringFormatEncoder.FormatSuffix.OrdinalIndicator:
                         // TODO: if the preceding expression value is not a float, it should be replaced with 0 (0th)
-                        if (interps[prevInterpIndex].TryGetValueAsFloat(out var ordinalNumber)) {
+                        if (interps[prevInterpIndex].TryGetValueAsInteger(out var ordinalNumber)) {
+
+                            /// For some mystical reason byond converts \th to integers
```
redmoogle, line 349: would it be better to do `if (interps[prevInterpIndex].TryGetValueAsFloat(out var ordinalNumber))` and cast it to an int?

wixoaGit:
Removing a piece of the string that was already constructed is a little strange, are there any other examples of string macros that come after their interpolated value?

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -344,7 +344,13 @@ private static void HandleSuffixPronoun(ref StringBuilder formattedString, ReadO
                         continue;
                     case StringFormatEncoder.FormatSuffix.OrdinalIndicator:
                         // TODO: if the preceding expression value is not a float, it should be replaced with 0 (0th)
-                        if (interps[prevInterpIndex].TryGetValueAsFloat(out var ordinalNumber)) {
+                        if (interps[prevInterpIndex].TryGetValueAsInteger(out var ordinalNumber)) {
+
+                            // For some mystical reason byond converts \th to integers
+                            // This is slightly hacky but the only reliable way I know how to replace the number
+                            // Need to call stringy to make sure its the right length to cut
+                            formattedString.Length -= interps[prevInterpIndex].Stringify().Length;
```
redmoogle, line 352: pretty sure stringify is the correct thing to do here but im not sure. we'll probably find out soon enough

wixoaGit (approved):

PR #880, opened 2022-10-28
Adds `MustGetValueAs[Type]()`

ike709:
Yes these are essentially the same thing as `GetValueAs[Type]()`, but converting them guarantees that they're being audited and converted to the `Try` variant as-needed. Plus IMO the name makes it clearer that it'll throw if it fails.

In this PR I also convert various `GetValueAs[Type]()` usages until I lost interest.

May conflict with https://github.com/OpenDreamProject/OpenDream/pull/873 but I'm fine with that getting merged first and me fixing any conflicts in this PR.


itsmeow:
Merge this first please

OpenDreamRuntime/DreamConnection.cs
```diff
@@ -211,7 +211,7 @@ public void HandleMsgPromptResponse(MsgPromptResponse message)
 
         public void OutputDreamValue(DreamValue value) {
             if (value.Type == DreamValue.DreamValueType.DreamObject) {
-                DreamObject outputObject = value.GetValueAsDreamObject();
```
SpaceManiac, line 214: Seems to me like replacing a (typecheck, Get) with a (typecheck, Must) instead of a single TryGet is defeating the whole point of this migration? Similar issue with the type checks in DMOpcodeHandlers

OpenDreamRuntime/DreamValue.cs
```diff
@@ -138,9 +137,17 @@ public enum DreamValueType {
             }
         }
 
-        [Obsolete("Deprecated. Use TryGetValueAsFloat() instead.")]
+        public int MustGetValueAsInteger() {
+            if (Type != DreamValueType.Float) {
+                throw new Exception("Value " + this + " was not the expected type of integer");
```
Altoids1, line 142: You can just get rid of these typecheck-and-throws in these Must methods. C# will emit its own runtime whenever it tries to unbox to the type and fails with a `InvalidCastException`.

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -1207,7 +1207,7 @@ private static void HandleSuffixPronoun(ref StringBuilder formattedString, ReadO
 
                 type = typeObject.ObjectDefinition.Type;
             } else {
-                type = typeValue.GetValueAsPath();
+                type = typeValue.MustGetValueAsPath();
```
wixoaGit, line 1210: Could also give a friendlier error message here.

wixoaGit (changes requested):

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -1268,14 +1268,14 @@ public static DreamValue NativeProc_oview(DreamObject instance, DreamObject usr,
...
-                    center = firstArgument.GetValueAsDreamObject();
+                if (firstArgument.TryGetValueAsDreamObject(out var firstObj)) {
+                    center = firstObj;
 
                     if (arguments.ArgumentCount > 1) {
                         distance = arguments.GetArgument(1, "Center").GetValueAsInteger();
                     }
                 } else {
-                    distance = firstArgument.GetValueAsInteger();
+                    distance = firstArgument.MustGetValueAsInteger();
```
ike709, line 1278:
```suggestion
                    if(!firstArgument.TryGetValueAsInteger(out distance)) {
                         throw new Exception("distance is not a number");
                     }
```

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
ike709, line 1210:
```suggestion
                if(!typeValue.TryGetValueAsPath(out type)) {
                    throw new Exception($"istype() attempted to check non-path {typeValue}");
                }
```

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -1322,8 +1324,8 @@ private static void HandleSuffixPronoun(ref StringBuilder formattedString, ReadO
                 case DreamValue.DreamValueType.String:
                     unsafe
                     {
-                        var dllName = source.GetValueAsString();
-                        var procName = state.Pop().GetValueAsString();
+                        var dllName = source.MustGetValueAsString();
```
ike709, line 1327:
```suggestion
                        if(!source.TryGetValueAsString(out var dllName)) {
                            throw new Exception($"{source} is not a valid DLL");
                        }
```

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -1322,8 +1324,10 @@ private static void HandleSuffixPronoun(ref StringBuilder formattedString, ReadO
                 case DreamValue.DreamValueType.String:
                     unsafe
                     {
-                        var dllName = source.GetValueAsString();
-                        var procName = state.Pop().GetValueAsString();
+                        if(!source.TryGetValueAsString(out var dllName)) {
+                            throw new Exception($"{source} is not a valid DLL");
+                        }
+                        var procName = state.Pop().MustGetValueAsString();
```
ike709, line 1330:
```suggestion
                        var popProc = state.Pop();
                        if(!popProc.TryGetValueAsString(out var procName)) {
                            throw new Exception($"{popProc} is not a valid proc name");
                        }
```

wixoaGit (approved):

PR #884, opened 2022-10-29
"Fixes" missing oddball behaviour with bitshift operations

Altoids1:
image

Thanks to @mumencoder for fuzzing this one out for us.

## Summary

DM actually allows the user to do bitshifting on things that are definitely not numbers. When doing so, the values are coerced to 0 for the purposes of the calculation. We now, regrettably, replicate this behaviour.

In an effort to provide good warning, I've added a new warning in Constant evaluation that informs the user when we are doing this stupid move.

You may ask, "But, Altoids, you didn't add this to the Runtime!" That's true, I didn't. Because it is a runtime error in BYOND, when the values given are not being const-rolled.

I hope this nonsense is first on the chopping block when we add pragmas.

## Changelog
- OD now correctly replicates some nonsense with the bitshift operators.

### DIFF
--- a/Content.Tests/DMProject/Tests/Operators/bitshift_more_like_bullshit.dm
+++ b/Content.Tests/DMProject/Tests/Operators/bitshift_more_like_bullshit.dm
@@ -1,0 +1,16 @@
+
+/proc/RunTest()
+	//Normal bitshifting
+	ASSERT((1 << 2) == 4)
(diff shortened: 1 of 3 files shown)


Altoids1:
Converting to a draft for now.
This is done, but in order to work we need to refactor our constant-folding system actually attempt to run on all expressions.

Issue #879, opened 2022-10-27
Weird NaN Comparison behavior


redmoogle:
`ASSERT(1.#IND == 1.#IND)` will return true on byond and false on opendream (what the fuck?)

but doing `ASSERT(trunc(1.#IND) == 1.#IND)` both return false

Altoids1:
I think in the real IEEE spec, NaN is not supposed to be equal to itself. This is BYOND contradicting the floating-point standard, I think.

ike709:
The *actual* NaN value does not equal itself in BYOND, this issue seems to be specific to `1.#IND`

redmoogle:
515 seemingly fixed this

PR #885, opened 2022-10-29
Deletes DMASTSimplifier and replaces it with TryAsConstant() calls

Altoids1:
## Summary

Right now, we aren't really consistent on how const-folding is handled. Things are only consistently folded whenever totally necessary, as in constant contexts like switch cases or whatever.

This fixes that by making ALL expressions go through a TryAsConstant pass during creation. Accordingly, this means that we don't really need DMASTSimplifier anymore, so I've removed that whole thing.

Doing this seems to shave off a kilobyte for the stripped TGstation copy I happen to have.

Works on #833 by adding == and != as const-foldable.

## Changelog
- Crunching down of lvalues and rvalues into constants is now just handled by TryAsConstant.
- Calls to `initial()` and the == and != operators are now also const-folded.
- The compiler now more consistently emits warnings when nulls are coerced to 0 in constant contexts.


DMCompiler/DM/Expressions/Constant.cs
```diff
@@ -83,6 +83,44 @@ abstract class Constant : DMExpression {
         public virtual Constant BinaryOr(Constant rhs) {
             throw new CompileErrorException(Location, $"const operation \"{this} | {rhs}\" is invalid");
         }
+
+        // Not a virtual because it'd be very annoying to write this for every Constant subtype
+        public Constant Equal(Constant rhs) {
+            switch (this) {
+                case Expressions.Null nil:
```
redmoogle, line 90: #879?

Altoids1: That line specifically is for `null`, not NaN.

DMCompiler/DM/Expressions/Builtins.cs
```diff
@@ -451,15 +451,27 @@ class Initial : DMExpression {
...
         public override void EmitPushValue(DMObject dmObject, DMProc proc) {
             if (_expr is LValue lValue) {
                 lValue.EmitPushInitial(dmObject, proc);
                 return;
             }
+            _expr.EmitPushValue(dmObject, proc);
+        }
 
-            throw new CompileErrorException(Location, $"can't get initial value of {_expr}");
+        public override bool TryAsConstant(out Constant constant) {
```
wixoaGit, line 469: `initial()` doesn't appear to constant-fold in byond? This gives me `=: constant initializer required`

```
/var/const/a = 9
/var/const/b = initial(a)
```

Even if it did, I don't think it should use `_expr.TryAsConstant()`, as this should be able to constant-fold even on non-constant values because they always begin with the same value.

wixoaGit (changes requested):

DMCompiler/DM/Expressions/Builtins.cs
Altoids1: I think it's a reasonable change to make `initial()` be const-folded. I think it better adheres to the user's expectation, since it's a proc that is *supposed* to be for querying compiletime values of things. I was quite surprised myself to hear that it wasn't const-folded). I also don't think it's a change that would ever negatively affect behaviour for any target, just make them run that little bit faster.

Rest of the review is on-point though, I'll look into that.

SpaceManiac:
The constant folding is incorrect:
```dm
/obj/var/derp = "/obj"
/datum/foldme/var/derp = "/datum/foldme"
/proc/Test()
    var/datum/foldme/D = new /obj
    src << "initial = [initial(D.derp)]"
```
Correct output is "/obj", this outputs "/datum/foldme".

Altoids1:
I pulled back a bit and reserved the initial() folding for static and const variable access, as those do rely on the type of the variable and not the value it necessarily contains.

Altoids1:
Marking this as having #896 as a dependency, since some things I want to do with `initial()` would break parity and should be deferred to user configuration.

wixoaGit:
This PR is fairly out of date, and I don't see it being picked back up anytime soon. While the removal of `DMASTSimplifier` and better constant folding is desirable, I think it may be better served as a bytecode-level optimization. Anyone feel free to bring it up with me if they disagree.

PR #886, opened 2022-10-30
arithmetic operation edge cases

mumencoder:


Content.Tests/DMProject/Tests/Expression/edge_arithmetic.dm
```diff
@@ -0,0 +1,22 @@
...
+	ASSERT( (-"str") == 0 )
+
+	ASSERT( (1.0 ** "abc") == 1 )
+	ASSERT( (1.0 ** null) == 1 )
+	ASSERT( (null ** "abc") == 1 )
+	ASSERT( (null ** 1.0) == 0 )
+
+	ASSERT( (null - null) == 0 )
+
+	ASSERT( (null * null) == 0 )
```
wixoaGit, line 13: This is a runtime error in DM v515: `runtime error: type mismatch: null ** 2`

This doesn't cause that however:
```
var/a = null
world.log << (a * null)
```

I think v515 made `null ** 2` an invalid operation and the compiler is turning `null * null` into that.

wixoaGit (changes requested):

wixoaGit:
Closing in favor of #1488

PR #887, opened 2022-10-30
bitwise operation edge cases/fixes

mumencoder:


wixoaGit:
This PR was abandoned a long while ago.

Another attempt at this would need to be aware of BYOND's differences in compile-time evaluation and runtime evaluation. We should decide whether to replicate this.

For example:
```dm
world.log << (~ -42) // BYOND evaluates at compile-time
world.log << (~ pick(-42)) // BYOND evaluates at runtime
```
gives (as of v515.1616),
```
16777215
41
```

PR #892, opened 2022-10-30
Ports the final remaining DTT tests

ike709:
Ports the `tree` directory of DTT tests, which is the last of the DTT tests that we're missing.

I skipped the dot abuse tests (`.....` and friends) because almost all of the tests are broken in BYOND and we still haven't entirely figured out how we're going to implement this dot behavior.

All tests were manually verified in MoMMI/BeeBot but there's still room for human error.

### DIFF
--- a/Content.Tests/DMProject/Tests/Tree/Const/const_fn_badarg.dm
+++ b/Content.Tests/DMProject/Tests/Tree/Const/const_fn_badarg.dm
@@ -1,0 +1,9 @@
+// COMPILE ERROR
+
+/proc/somered()
+	return 127
+
+var/const/reddish = rgb(somered(),0,255) // error: constant initializer required
+
+/proc/RunTest()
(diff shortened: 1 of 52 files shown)


Content.Tests/DMProject/Tests/Tree/Const/const_fn_badarg.dm
```diff
@@ -0,0 +1,9 @@
+// COMPILE ERROR
+
+/proc/somered()
+	return 127
+
+var/const/reddish = rgb(somered(),0,255) // error: constant initializer required
+
+/proc/RunTest()
+	var/const/reddish = rgb(somered(),0,0) // error: constant initializer required
```
wixoaGit, line 9: Only one of these erroring would lead the test to pass, which could hide an issue if the other one ends up compiling fine.

wixoaGit (changes requested):

wixoaGit (approved):

PR #896, opened 2022-11-03
Implements compiler configuration through new preproc directive: `#pragma`

Altoids1:
image

## Summary
The syntax for `#pragma` that I am going with is probably `#pragma errorName Value`, where errorName can be either the error code or enum name for the error code, and Value is either `disabled`, `notice`, `warning`, or `error`.

This will allow us to vary the behaviour of the compiler, adding more warnings for bad behaviour, with users having the option to forbid the bad behaviour entirely.

## Changelog
- Argument parsing is now a bit smarter, and supports setting arguments to values.
- Fixed OD accepting macro functors that had several variadic arguments.
- New directive unlocked! `#pragma`, which allows both us and the user to decide the elevation of various error emissions.

### DIFF
--- a/DMCompiler/Compiler/DM/DMParserHelper.cs
+++ b/DMCompiler/Compiler/DM/DMParserHelper.cs
@@ -8,6 +8,20 @@ using System.Text;
 
 namespace DMCompiler.Compiler.DM {
     public partial class DMParser : Parser<Token> {
+
+        /// <summary>
+        /// A special override of Error() since, for DMParser, we know we are in a compilation context and can make use of error codes.
+        /// </summary>
+        /// <remarks>
+        /// Should only be called AFTER <see cref="DMCompiler"/> has built up its list of pragma configurations.
+        /// </remarks>
+        /// <returns> True if this will raise an error, false if not. You can use this return value to help improve error emission around this (depending on how permissive we're being)</returns>
+        protected bool Error(WarningCode code, string message) {
+            ErrorLevel level = DMCompiler.CodeToLevel(code);
+            Emissions.Add(new CompilerEmission(level, code, Current().Location, message));
+            return level == ErrorLevel.Error;
+        }
+
         protected bool PeekDelimiter() {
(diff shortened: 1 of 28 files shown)


DMCompiler/Compiler/DMPreprocessor/DMPreprocessor.cs
```diff
@@ -541,6 +575,76 @@ private List<Token> GetLineOfTokens()
...
+            if (warningTypeToken.Type != TokenType.DM_Preproc_Identifier) {
+                DMCompiler.Emit(WarningCode.BadDirective, warningNameToken.Location, $"Warnings can only be set to disabled, notice, warning, or error");
+                return;
+            }
+            switch(warningTypeToken.Text.ToLower()) {
+                case "disabled":
+                case "disable":
+                    DMCompiler.SetPragma(warningCode, ErrorLevel.Disabled);
+                    break;
+                case "notice":
```
ike709, line 622:
```suggestion
                case "notice":
                case "info":
```
pretty sure I've seen this somewhere

wixoaGit (changes requested):

DMCompiler/Compiler/DM/DMParserHelper.cs
```diff
@@ -8,6 +8,20 @@
...
     public partial class DMParser : Parser<Token> {
+
+        /// <summary>
+        /// A special override of Error() since, for DMParser, we know we are in a compilation context and can make use of error codes.
+        /// </summary>
+        /// <remarks>
+        /// Should only be called AFTER <see cref="DMCompiler"/> has built up its list of pragma configurations.
+        /// </remarks>
+        /// <returns> True if this will raise an error, false if not. You can use this return value to help improve error emission around this (depending on how permissive we're being)</returns>
+        protected bool Error(WarningCode code, string message) {
```
ike709, line 19: Maybe this should be called `TryError()` or something to distinguish it?

wixoaGit (approved):

PR #897, opened 2022-11-04
Improve accuracy of world initialization sequence

SpaceManiac:
Splits DreamManager.Initialize into two halves:
1. PreInitialize
    1. JSON is loaded
    2. World instance is created
    3. Compiled-in turfs and areas are added to `world.contents`, and their `<init>` is called, but ALL `New` are suppressed
        * Datums created during these `<init>` never have their `New` called
2. StartWorld
    1. `<global init>` is spawned
    2. All areas in the world have their `New()` spawned, in order
        * If `<global init>` created any areas, their New is called a second time
    4. All turfs in the world have their `New()` spawned, in reverse order
    5. All objects and mobs in compiled-in maps are `new`ed in the typical way, in order
    6. `world.New()` is spawned

Other improvements:
- Debugger calls `PreInitialize` as soon as it knows the JSON filename, but calls `StartWorld` only on ConfigurationDone
  - This will allow "stop on entry" support in the future
- Compiled-in areas are distinguished based on their full prototype, not just their typepath
- Turfs are added to `area.contents`

There are still some inaccuracies:
- Compiler should assign root-level `/var/static`s before root-level `/var`s
- The various threads spawned during StartWorld should be combined into one thread that calls procs as though with `waitfor = FALSE`


wixoaGit:
This appears to greatly slow down map loading.

Loading Box Station on my [stripped /tg/ fork](https://github.com/wixoaGit/tgstation) before:
`world.log: Loaded Station in 9.2s!`
After:
`world.log: Loaded Station in 326.4s!`

Haven't looked much into it yet, but I suspect this is due to the adding/removing of items on `area.contents`. Maybe this could instead be done with a special list similar to [`datum.vars`](https://github.com/OpenDreamProject/OpenDream/blob/45d41d6dbcb0b0e1e4a6c67ba710cdf7fb565106/OpenDreamRuntime/Objects/DreamList.cs#L187).

SpaceManiac:
Good call. The big contributors were UnionValue on `area.contents` and `world.contents`, which I really should have seen coming. My machine loads Box Station much slower than yours on OD master + your /tg/ branch (98.5s) but with this PR fixed it now loads about the same (104s).

SpaceManiac:
Ah, compiling OpenDreamServer in release mode gets me ~10.5s to load your Box both on OD master and this PR.

OpenDreamRuntime/Objects/DreamObject.cs
```diff
@@ -21,11 +21,11 @@ public class DreamObject {
             thread.PushProcState(procState);
 
             if (thread.Resume() == DreamValue.Null) {
-                thread.HandleException(new InvalidOperationException("DreamObject.InitSpawn called a yielding proc!"));
+                //thread.HandleException(new InvalidOperationException("DreamObject.InitSpawn called a yielding proc!"));
             }
```
wixoaGit, line 25: This can be removed entirely

wixoaGit (changes requested):

wixoaGit (approved):

PR #900, opened 2022-11-09
Add `--define` flag to DMCompiler

SpaceManiac:
Purpose: serve /tg/station's need to sometimes `#define CBT`, `CIBUILDING`, `AUTOWIKI`, `TGS` at build time, currently satisfied by generating a `.m.dme` file with the `#define`s prepended.

Interface:
- `--define FOO` to define a macro with no expansion, for `#ifdef`
- `--define FOO=10+7` to define a macro which expands to some tokens, for `#if`

Also:
- Exit DMCompiler with status 1 if the command-line arguments were invalid
- Fix multiple-file inputs being processed in reverse order

### DIFF
--- a/DMCompiler/Compiler/DMPreprocessor/DMPreprocessor.cs
+++ b/DMCompiler/Compiler/DMPreprocessor/DMPreprocessor.cs
@@ -180,8 +180,11 @@ namespace DMCompiler.Compiler.DMPreprocessor {
             _defines.Add(key, new DMMacro(null, list));
         }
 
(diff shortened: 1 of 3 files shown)


wixoaGit:
This does not override defines that exist within the DM code. Might be desirable for it to do so?

SpaceManiac:
I could see it being useful to override DMStandard defines. I don't think it should get special priority over `#define`s in the codebase proper - it's most useful to set conditional flags or values that the codebase is already prepared to understand.

PR #906, opened 2022-11-13
Paradise Interface Fixes

amylizzle:
Some fixes to our interface code to make it act more byondish. 

The first was a mistake from my DMF 0x0 fix, which treated `pos=none` as `pos=0x0`.
The second is some lovely jank where Paradise was setting `anchor2=-1,-1` which by DM ref is undefined behaviour, but seems to basically just be equivalent to `anchor2=none`. 
The third was some out-of-order components, which I thought broke BYOND's interface code in my tests, but maybe in some cases it just gets fixed silently? Now OD fixes it not-silently.

Anyway, Paradise's interface will now display unmodified and mostly correct.
image


amylizzle:
segfaulting, dnm

OpenDreamClient/Interface/Controls/ControlWindow.cs
```diff
@@ -179,8 +196,31 @@ public void CreateChildControls(IDreamInterfaceManager manager)
                     ControlDescriptorTab => new ControlTab(controlDescriptor, this),
                     _ => throw new Exception($"Invalid descriptor {controlDescriptor.GetType()}")
                 };
+                // Can't have out-of-order components, so make sure they're ordered properly
+                if(ChildControls.Count > 0)
+                {
```
wixoaGit, line 201:
```suggestion
                if(ChildControls.Count > 0) {
```

OpenDreamClient/Interface/Controls/ControlWindow.cs
```diff
@@ -179,8 +196,31 @@ public void CreateChildControls(IDreamInterfaceManager manager)
...
+                    var prevPos = ChildControls[ChildControls.Count-1].Pos.GetValueOrDefault();
+                    var curPos = control.Pos.GetValueOrDefault();
+                    if(prevPos.X <= curPos.X && prevPos.Y <= curPos.Y)
+                        ChildControls.Add(control);
+                    else
+                    {
```
wixoaGit, line 207:
```suggestion
                    else {
```

OpenDreamClient/Interface/Controls/ControlWindow.cs
```diff
@@ -179,8 +196,31 @@ public void CreateChildControls(IDreamInterfaceManager manager)
...
+                        Logger.Warning($"Out of order component {control.Name}. Elements should be defined in order of position. Attempting to fix automatically.");
+                        int i = 0;
+                        while(i < ChildControls.Count)
+                        {
```
wixoaGit, line 211:
```suggestion
                        while(i < ChildControls.Count) {
```

OpenDreamClient/Interface/Controls/ControlWindow.cs
```diff
@@ -179,8 +196,31 @@ public void CreateChildControls(IDreamInterfaceManager manager)
...
+                            prevPos = ChildControls[i].Pos.GetValueOrDefault();
+                            if(prevPos.X <= curPos.X && prevPos.Y <= curPos.Y)
+                                i++;
+                            else
+                                break;
+                        }
+                        ChildControls.Insert(i, control);
+                    }
+                }
+                else
```
wixoaGit, line 221:
```suggestion
                } else
```

wixoaGit (changes requested):

PR #907, opened 2022-11-13
Move some fixed tests around

ike709:
I tossed the broken tests into the regular tests to see if any pass, and these did.


wixoaGit:
The `Statements/Control Flow/sleep/no_paren.dm` test fails sometimes due to `world.realtime` precision.

wixoaGit (approved):

PR #909, opened 2022-11-13
Unicode test correction?

TiviPlus:
Dunno if this is intended but the test is wrong when compared to byond parity


wixoaGit:
This is intended. The difference exists here because our strings are encoded in UTF-16 instead of BYOND's UTF-8.

An explanation for this exists in our [Differences Between OpenDream and BYOND](https://github.com/OpenDreamProject/OpenDream/wiki/Differences-Between-OpenDream-and-BYOND#text-encoding) wiki page.

PR #913, opened 2022-11-14
Fixes turn and adds tests

redmoogle:
if statement didnt account for any other number above 720 or below -720

Adds comprehensive test coverage


DMCompiler/DMStandard/_Standard.dm
```diff
@@ -327,19 +327,21 @@ proc/get_dir(atom/Loc1, atom/Loc2)
 			if (Angle != 0)
 				return pick(NORTH, SOUTH, EAST, WEST)
```
wixoaGit, line 328: An invalid dir with non-zero angle can also return random diagonal directions.
An invalid dir with a zero angle will return that same dir, unless it isn't a number in which case it's an exception.
```suggestion
			if (Angle != 0)
				return pick(NORTH, SOUTH, EAST, WEST, NORTHEAST, SOUTHEAST, SOUTHWEST, NORTHWEST)
			else if (!isnum(Dir))
				CRASH("Invalid Dir \"[json_encode(Dir)]\"")
			else
				return Dir
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #915, opened 2022-11-15
list(a = x, (b)= y) support

willox:
I wrap `DMASTIdentifier` nodes with `DMASTIdentifierWrapped` if they are surrounded in brackets.

Spooky... is this workable?


wixoaGit (approved):

wixoaGit:
I looked through usages of `DMASTIdentifier` and found that `goto (label)` works (along with probably `break` and `continue`), but that's probably a separate issue.

Nothing else really popped out at me as a possible problem.

PR #920, opened 2022-11-16
Parser lookahead fix

mumencoder:


willox:
I think ReuseToken requires the following:

```csharp
    if (_lookahead.Count > 0) {
        _lookahead.Peek().Pop();
    }
```

willox:
I also think that AcceptPosition should be:
```csharp
    _lookahead.Pop();
```

mumencoder:
> I think ReuseToken requires the following:
> 
> ```cs
>     if (_lookahead.Count > 0) {
>         _lookahead.Peek().Pop();
>     }
> ```

This does mostly fix the combination of ReuseToken and the lookahead function calls. Testing showed that it is still important for an Advance to be the next state change after calls to both SavePosition and RestorePosition. Also an Accept cannot directly follow a Reuse.

wixoaGit:
Nothing uses these methods and honestly I'd rather get rid of them

PR #923, opened 2022-11-17
Fixes `<<=` and `>>=` use on complicated references, fixes `null << 1` not being `0`

### LINKED ISSUE #922: `>>=` and `<<=` operators are broken
`>>=` and `<<=` are broken 

https://github.com/OpenDreamProject/OpenDream/blob/15de2c2c7621a032946073a672c4a7e462ec096a/DMCompiler/DM/Expressions/Binary.cs#L607-L610

https://github.com/OpenDreamProject/OpenDream/blob/15de2c2c7621a032946073a672c4a7e462ec096a/DMCompiler/DM/Expressions/Binary.cs#L594-L597

invalid use of references, no test case because i am playing Tom Clancy's Rainbow Six® Siege

Altoids1:
Fixes #922.
## Summary
I just went ahead and added new opcodes for these, in symmetry with basically all of the other compound assignments.

&&= and ||= also could use some clean-up but I'm going to elect to leave that for after the Reference refactor gets merged, or whatever.

## Changelog
- Using `<<=` and `>>=` on complicated references no longer fails to compile.
- OD now "correctly" evaluates `null << 1` as `0`.

### DIFF
--- a/Content.Tests/DMProject/Tests/Operators/bitshift_assign.dm
+++ b/Content.Tests/DMProject/Tests/Operators/bitshift_assign.dm
@@ -15,7 +15,12 @@
 	for(var/i = 1; i < 6; i += 1)
 		a[i] <<= 1
 		ASSERT(a[i] == (i << 1))
+		a[i] >>= 1
+		ASSERT(a[i] == i)
 	//Make sure cursed behaviour still works in the assignment situation
 	var/n = null
 	n <<= 1
+	ASSERT(n == 0)
+	n = null
+	n >>= 1
 	ASSERT(n == 0)
--- a/Content.Tests/DMProject/Tests/Operators/bitshift_assign.dm
+++ b/Content.Tests/DMProject/Tests/Operators/bitshift_assign.dm
@@ -1,0 +1,21 @@
+
+// Issue OD#922: https://github.com/OpenDreamProject/OpenDream/issues/922
+
+/proc/RunTest()
+	//Easy references
+	var/x = 1
+	var/y = 2
+	x <<= 5
+	y <<= 5
+	x >>= 4
+	y >>= 5
+	ASSERT(x == y)
+	//Compound references
+	var/list/a = list(1,2,3,4,5)
+	for(var/i = 1; i < 6; i += 1)
+		a[i] <<= 1
+		ASSERT(a[i] == (i << 1))
+	//Make sure cursed behaviour still works in the assignment situation
+	var/n = null
+	n <<= 1
+	ASSERT(n == 0)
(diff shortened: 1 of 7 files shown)


Content.Tests/DMProject/Tests/Operators/bitshift_assign.dm
```diff
@@ -0,0 +1,21 @@
...
+	ASSERT(x == y)
+	//Compound references
+	var/list/a = list(1,2,3,4,5)
+	for(var/i = 1; i < 6; i += 1)
+		a[i] <<= 1
+		ASSERT(a[i] == (i << 1))
+	//Make sure cursed behaviour still works in the assignment situation
+	var/n = null
+	n <<= 1
+	ASSERT(n == 0)
```
wixoaGit, line 21: Maybe test `>>=` here too

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -787,6 +790,27 @@ private static void HandleSuffixPronoun(ref StringBuilder formattedString, ReadO
...
+            DMReference reference = state.ReadReference();
+            DreamValue second = state.Pop();
+            DreamValue first = state.GetReferenceValue(reference, peek: true);
+            DreamValue result;
+            switch (first.Type) {
+                case DreamValue.DreamValueType.DreamObject when first == DreamValue.Null:
+                    result = new DreamValue(0);
+                    break;
+                case DreamValue.DreamValueType.Float when second.Type == DreamValue.DreamValueType.Float:
+                    result = new DreamValue(first.GetValueAsInteger() << second.GetValueAsInteger());
```
wixoaGit, line 804: `GetValueAsInteger()` is deprecated. This should be converted to use the `TryGetValueAs[]()` variants instead of checking the type in a `switch`.

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -802,6 +826,26 @@ private static void HandleSuffixPronoun(ref StringBuilder formattedString, ReadO
...
+            DMReference reference = state.ReadReference();
+            DreamValue second = state.Pop();
+            DreamValue first = state.GetReferenceValue(reference, peek: true);
+            DreamValue result;
+            switch (first.Type) {
+                case DreamValue.DreamValueType.DreamObject when first == DreamValue.Null:
+                    result = new DreamValue(0);
+                    break;
+                case DreamValue.DreamValueType.Float when second.Type == DreamValue.DreamValueType.Float:
+                    result = new DreamValue(first.GetValueAsInteger() >> second.GetValueAsInteger());
```
wixoaGit, line 839: `GetValueAsInteger()` is deprecated.

wixoaGit (changes requested):

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
Altoids1: What I wanna use here is MustGetValueAsInteger but that isn't in yet, so

wixoaGit (approved):

PR #925, opened 2022-11-17
revamps dereference ast nodes/emitted code

willox:
Now an operation like `a.b.c().d[5]` is treated as a base expression (`a`) and a chain of operations to perform on the base expression (`.b`, `.c()`, `.d`, `[5]`) at the AST level.

This allows short-circuiting operators (`?.`, `?:`, `?[`) to skip the entire dereference expression as expected, so `a?.b.c.d` should evaluate as null without error if `a` is null.

~~Drafting atm because I need to improve the new instructions' behaviour when short-circuiting, `a?.b << c` is broken.~~

~~I'll probably rename all the "Deref" stuff back to "Dereference" too, it's like this because I didn't remove the old code until this all worked (unrebased commits available at https://github.com/willox/OpenDream/tree/dereference)~~


willox:
@SpaceManiac this merge commit moves your logic from https://github.com/OpenDreamProject/OpenDream/pull/814 to work at runtime and not infer based on variable names, seems like the correct approach

SpaceManiac:
I have not reviewed this PR yet, but I now agree that treating `vars` specially at compile-time is not the right approach

DMCompiler/DM/Visitors/DMVisitorExpression.cs
```diff
@@ -434,56 +437,250 @@ public void VisitIdentifier(DMASTIdentifier identifier)
...
+                        break;
+
+                    case DMASTDereference.OperationKind.Index:
+                    case DMASTDereference.OperationKind.IndexSafe:
+                        // Passing the path here is cursed, but one of the tests seems to suggest we want that?
+                        operation.Index = DMExpression.Create(_dmObject, _proc, astOperation.Index, prevPath);
+                        operation.Path = null;
+
+                        // This is cursed too
+                        // prevPath = prevPath;
```
wixoaGit, line 624: What was this commented line for?

OpenDreamRuntime/Procs/ProcDecoder.cs
```diff
@@ -126,6 +127,9 @@ public struct ProcDecoder {
             case DreamProcOpcode.PushType:
             case DreamProcOpcode.DebugLine:
             case DreamProcOpcode.MassConcatenation:
+            case DreamProcOpcode.JumpIfNullNoPop:
+            case DreamProcOpcode.JumpIfTrueReference:
+            case DreamProcOpcode.JumpIfFalseReference:
```
wixoaGit, line 132: These two should go with `JumpIfNullDereference` below.

`Initial` also no longer takes an operand and should be removed from this switch.

wixoaGit (changes requested):

DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -1589,38 +1594,31 @@ public DMASTProcStatementLabel Label(DMASTIdentifier expression)
...
-                            conditional = deref.Conditional;
-                            deref = expr as DMASTDereference;
-                        } while (deref != null);
+                        var lastOp = deref.Operations[^1];
 
-                        if (deref == null && type == DereferenceType.Search) {
-                            b = expr;
-                        } else {
-                            Error("Expected ':'");
+                        if (lastOp.Kind == DMASTDereference.OperationKind.FieldSearch) {
```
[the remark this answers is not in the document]
willox: pretty much impossible until the parser can probably look ahead

DMCompiler/DM/Visitors/DMVisitorExpression.cs
willox: I was expecting some weird fuckery with the inferred type of `L[x]` but I think I was wrong since it didn't hit any unit tests.

willox:
todo
- [ ] make sure all the uses of endLabel in the EmitReference function work properly (primarily sus places are enumerators or other loops)
- [ ] perfect ternary behaviour as far as feasible, i have another branch with a bunch of work on this already to merge over, it's unlikely that all cases can be covered with the current potential of the parser to rewind back through a variabler amt of tokens
- [ ] ???

willox:
added a bunch of changes to make ternary closer to byond 
image

willox:
The stuff I just hard-pushed away can be found in https://github.com/willox/OpenDream/tree/dereference-rebased-citest

willox:
I've merged master in to this and rectified some fucky behaviour with our "nested" inferred types. https://github.com/OpenDreamProject/OpenDream/pull/1071 is currently reverted (it's not incorrect, but it is incompatible with the changes in this PR and I'm not reimplementing it now.)

wixoaGit (approved):

PR #928, opened 2022-11-19
macros that contain preprocessor directives now work

willox:
Fixes the following cases: https://github.com/OpenDreamProject/OpenDream/blob/f83c003b230d16bf527947b3eb5d575a8769b433/Content.Tests/DMProject/Tests/Preprocessor/EmptyAndDirectiveMacros.dm#L3-L9


willox:
Needs some more work to get the following to work:

```
#define EMPTY
#define DEFINE #define

EMPTY DEFINE A 1
```

willox:
Turns out byond isn't quite that forgiving when it comes to whitespace in front of macro invocations. There is an edge-case I am not handling:

```
// Works in BYOND
#define EMPTY
#define DEFINE #define
EMPTY DEFINE A 1

// Does not work in BYOND (note the space before the DEFINE invocation)
#define EMPTY
#define DEFINE #define
 DEFINE A 1
```

It seems like the first EMPTY invocation determines the indentation level of the file, I think to track something like that we'd need an extra variable to track the state similar to `_currentLineWhitespaceOnly`.

Anyway, I'm finished with this PR since it probably handles the important cases.

Altoids1:
Goon enjoyers rejoice

amylizzle:
goon enjoyers un-joice D:

Issue #857, opened 2022-10-25
`ftime()` not implemented


ZeWaka:
Returns either creation or last-modified time based on the given argument

https://www.byond.com/docs/ref/#/proc/ftime

PR #929, opened 2022-11-20
Implement `ftime` proc (#857)

UARTman:
As my first contribution to the project, I've implemented `ftime()` proc (#857). 

In implementing it I made a couple of assumptions about how OpenDream's API works:

- I assumed that calling `.IsTruthy()` on an absent optional parameter will give `false`
- I assumed that the time expected is in the same format as the one `world/realtime` returns, namely, that I need to take the milliseconds passed from 1.1.2000 and divide them by 100.
- I assumed that what is considered a filename by BYOND is a valid path in the eyes of the `FileInfo` constructor.

If these assumptions all check out, the proc should work properly. I couldn't test, since I'm using Linux and setting up BYOND under WINE would be a pain. But if needed, I might use my Windows machine to test this.


wixoaGit:
> I assumed that calling .IsTruthy() on an absent optional parameter will give false

An absent parameter will be null or the default value if the `DreamProcParameter` has one. `IsTruthy()` returns false for null so this is correct.

The 2nd and 3rd assumptions appear to be correct.

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -630,6 +630,23 @@ public static DreamValue NativeProc_clamp(DreamObject instance, DreamObject usr,
             return new DreamValue(0);
         }
 
+        [DreamProc("ftime")]
+        [DreamProcParameter("File", Type = DreamValueType.String)]
+        [DreamProcParameter("IsCreationTime", Type = DreamValueType.Float)]
+        public static DreamValue NativeProc_ftime(DreamObject instance, DreamObject usr, DreamProcArguments arguments) {
```
wixoaGit, line 636: This is missing a `SetGlobalNativeProc()` call in [DreamProcNative.cs](https://github.com/OpenDreamProject/OpenDream/blob/b36a2b68c62c90e86b53bc4e3ffc184d6996dc76/OpenDreamRuntime/Procs/Native/DreamProcNative.cs). Without this it will run the proc defined in DM, which is just an empty proc.

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -630,6 +630,23 @@ public static DreamValue NativeProc_clamp(DreamObject instance, DreamObject usr,
...
+            DreamValue file = arguments.GetArgument(0, "File");
+            DreamValue isCreationTime = arguments.GetArgument(1, "IsCreationTime");
+
+            if (file.TryGetValueAsString(out var rscPath)) {
+                var fi = new FileInfo(rscPath);
+                if (isCreationTime.IsTruthy()) {
+                    return new DreamValue((fi.CreationTime - new DateTime(2000, 1, 1)).Milliseconds / 100);
```
wixoaGit, line 643: This should be using `TotalMilliseconds`
```suggestion
                    return new DreamValue((fi.CreationTime - new DateTime(2000, 1, 1)).TotalMilliseconds / 100);
```

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -630,6 +630,23 @@ public static DreamValue NativeProc_clamp(DreamObject instance, DreamObject usr,
...
+                }
+                return new DreamValue((fi.LastWriteTime - new DateTime(2000, 1, 1)).Milliseconds / 100);
```
wixoaGit, line 645:
```suggestion
                return new DreamValue((fi.LastWriteTime - new DateTime(2000, 1, 1)).TotalMilliseconds / 100);
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #934, opened 2022-11-22
Fixes lateral proc overrides, OD now errors on missing types promised by var definitions, and more!

### LINKED ISSUE #933: Lateral proc overrides do not behave correctly in OD
OD currently doesn't really differentiate between (and understand the parsing of) proc overrides and proc definitions. Instead, it emits compile-times like these:
```
Error at code\modules\antagonists\changeling\powers\transform.dm:167:7: Type /datum/antagonist/changeling already has a proc named "select_dna"
```
Whenever code like this is discovered:
```dm
/datum/antagonist/changeling/select_dna(var/prompt, var/title)
// ... and then later on in the file ...
/datum/antagonist/changeling/proc/select_dna(var/prompt, var/title)
```

Altoids1:
Fixes #933, and a bunch of other things that don't have issues on the repo, AFAIK.
## Summary
The main goal was to fix a range of errors that Yogs is having on master, where a proc override was appearing before its definition on the same type.

Looking into that ended up involving doing a whole mess of other bugfixes to get everything working. `/datum/do/re/mi` wasn't a valid type for some reason, some random unit tests that we were passing suddenly failed, and I had some strange issues with performance.

So this PR is more of a shotgun blast of everything necessary to squash that issue and get Yogs working. `DMObjectBuilder` grows ever more complex, I guess.

## Changelog
- Procs declared in a proc\n\tfoo() format are now properly informed of whether they are a definition or an override.
- Procs that override a proc on the exact same path now behave correctly and do not compiletime when discovered out-of-order.
- Global procs now behave better when being overridden.
- OD now properly has an emission when a definition for an overridden proc is not discovered.
- Using the text `while`, `do`, or `for` in a type path (as in /datum/do/re/mi) is now valid and behaves as expected.
- Removed a bad unit test and added a few more!


Altoids1:
This has https://github.com/OpenDreamProject/OpenDream/pull/935 as a dependency.

ike709:
CI should pass now if you rerun it. The unused vars in stripped TG have been removed.

Altoids1:
Lookin' good 😎

wixoaGit (changes requested):

wixoaGit (approved):

PR #946, opened 2022-12-02
Proper `/icon` support

### LINKED ISSUE #719: Need to be able to load a specific state/dir of a `DreamIcon`
Stuff like JSON serializing an icon or `browse_rsc()`ing an icon need to be able to LoadResource only the specific state/dir of a `DreamIcon`.

wixoaGit:
These changes fix up the `/icon` implementation quite a bit and lay a lot of the foundation needed for more complex operations.

- Icons can now contain icon states from multiple different images
- Icons can now generate their final .dmi file with the `GenerateDMI()` method
- These .dmi files can be sent to the client for use & display
- Resources now internally use IDs rather than file paths

Fixes #719

### DIFF
--- a/OpenDreamRuntime/DreamConnection.cs
+++ b/OpenDreamRuntime/DreamConnection.cs
@@ -235,7 +235,7 @@ namespace OpenDreamRuntime
                     msg.ResourceId = soundResource?.Id;
                     if (soundResource?.ResourcePath is { } resourcePath) {
                         if (resourcePath.EndsWith(".ogg"))
-                            msg.Format = MsgSound.FormatType.Wav;
+                            msg.Format = MsgSound.FormatType.Ogg;
                         else if (resourcePath.EndsWith(".wav"))
                             msg.Format = MsgSound.FormatType.Wav;
                         else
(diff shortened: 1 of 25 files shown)


OpenDreamRuntime/DreamConnection.cs
```diff
@@ -223,16 +224,25 @@ public void HandleMsgPromptResponse(MsgPromptResponse message)
...
+                        } else if (file != DreamValue.Null) {
+                            throw new ArgumentException($"Cannot output {value}", nameof(value));
+                        }
                     }
 
-                    Session.ConnectedClient.SendMessage(msg);
+                    msg.ResourceId = soundResource?.Id;
+                    if (soundResource?.ResourcePath is { } resourcePath) {
+                        if (resourcePath.EndsWith(".ogg"))
+                            msg.Format = MsgSound.FormatType.Wav;
```
wixoaGit, line 238:
```suggestion
                            msg.Format = MsgSound.FormatType.Ogg;
```

OpenDreamShared/Resources/DMIParser.cs
```diff
@@ -120,11 +79,77 @@ public sealed class ParsedDMIState {
...
+                    directions = new(1);
+
+                    if (!Directions.TryGetValue(dir.Value, out var frames))
+                        frames = Array.Empty<ParsedDMIFrame>();
+
+                    // Getting only one direction will give it to you with AtomDirection.South
+                    directions.Add(AtomDirection.South, frames);
+                }
+
+                if (frame != null) { // Only copy a specified frame
```
wixoaGit, line 142:
```suggestion
                if (frame != null) { // Only get a specified frame
```

PR #962, opened 2022-12-13
Macros that contain preprocessor directives now work

amylizzle:
Adopted #928 with @willox's permission.

Goon enjoyers re-rejoice!


Content.Tests/DMProject/Tests/Preprocessor/EmptyAndDirectiveMacros.dm
```diff
@@ -0,0 +1,26 @@
...
+	DEFINE NAME VALUE
+#define DEFINE_NESTED_WITH_STATEMENT(NAME, VALUE) \
+	var/test = 4\
+	DEFINE NAME VALUE
+
+DEFINE A 1
+DEFINE_WITH_EMPTY B 2
+DEFINE_NESTED(C, 3)
+EMPTY DEFINE D 4
+ DEFINE E 5
```
wixoaGit, line 15: BYOND doesn't like the space at the beginning of this line

Content.Tests/DMProject/Tests/Preprocessor/EmptyAndDirectiveMacros.dm
```diff
@@ -0,0 +1,26 @@
...
+DEFINE_NESTED_WITH_STATEMENT(F, 6)
```
wixoaGit, line 16: Using this macro makes BYOND throw a fit. I think it causes it to treat every token after it as an attempt at declaring a var?

DMCompiler/Compiler/DMPreprocessor/DMPreprocessor.cs
```diff
@@ -21,6 +21,9 @@ public sealed class DMPreprocessor : IEnumerable<Token> {
         //Every include pushes a new lexer that gets popped once the included file is finished
         private readonly Stack<DMPreprocessorLexer> _lexerStack =  new(8); // Capacity Note: TG peaks at 4 at time of writing
 
+        private Stack<Token> _bufferedWhitespace = new();
+        private bool _currentLineContainsNonWhitespace = false;
+        private bool _canUseDirective = true;
         private readonly HashSet<string> _includedFiles = new(5120); // Capacity Note: TG peaks at 4860 at time of writing
         private readonly Stack<Token> _unprocessedTokens = new(8192); // Capacity Note: TG peaks at 6802 at time of writing
         private bool _currentLineWhitespaceOnly = true;
```
wixoaGit, line 29: This is no longer used and should be removed.
```suggestion
```

DMCompiler/Compiler/DMPreprocessor/DMPreprocessor.cs
```diff
@@ -135,12 +142,20 @@ public sealed class DMPreprocessor : IEnumerable<Token> {
...
+                        }
+
+                        while (_bufferedWhitespace.TryPop(out var whitespace)) {
+                            yield return whitespace;
                         }
+                        _currentLineContainsNonWhitespace = true;
+                        _canUseDirective = false;
+
+                        yield return token;
                         break;
```
wixoaGit, line 157: Could fall-through to one of the cases below to remove the duplicated code
```suggestion
                        // Otherwise treat it like any other normal token
                        goto case TokenType.DM_Preproc_Punctuator;
```

wixoaGit (changes requested):

Content.Tests/DMProject/Tests/Preprocessor/EmptyAndDirectiveMacros.dm
amylizzle: That seems like a BYOND bug? swapping `DEFINE` for `#define` compiles just fine. I can't really see any impact in making it more permissive.

Content.Tests/DMProject/Tests/Preprocessor/EmptyAndDirectiveMacros.dm
amylizzle: Yeah, I have no idea what the hell BYOND is doing here. Seems like it might be a manifestation of the previous bug with whitespace before a `DEFINE` macro, since `DEFINE_NESTED` works.

amylizzle:
Reported the weird byond behaviour as a bug: http://www.byond.com/forum/post/2845642

Content.Tests/DMProject/Tests/Preprocessor/EmptyAndDirectiveMacros.dm
amylizzle: Okay, after a bit of coderbus discussion, it turns out the issue here is that BYOND eats newlines in multiline macro definitions, and we don't. So for parity, I guess I should make this also eat newlines, in which case the macro should be defined with `var test = 4;` with the semicolon

DMCompiler/Compiler/DMPreprocessor/DMPreprocessor.cs
```diff
@@ -165,7 +165,7 @@ public sealed class DMPreprocessor : IEnumerable<Token> {
                             yield return whitespace;
                         }
                         _currentLineContainsNonWhitespace = true;
-                        _canUseDirective = false;
+                        _canUseDirective = true;
```
wixoaGit, line 168: Why'd you switch this to true?

amylizzle: Because its entirely valid in BYOND to do `var/thing = "stuff" #define ONE 1`

wixoaGit: I’m fairly sure that’s an error.

This change also leads to `_canUseDirectives` never being false.

amylizzle: Ah, you are right, `_canUseDirective` should be set back to true if a `;` is encountered, but that's it

DMCompiler/Compiler/DMPreprocessor/DMPreprocessor.cs
```diff
@@ -135,27 +141,46 @@ public sealed class DMPreprocessor : IEnumerable<Token> {
...
+                    case TokenType.DM_Preproc_Identifier: {
+                        if (TryMacro(token)) {
+                            break;
+                        }
+
+                        // Otherwise treat it like any other normal token
+                        goto case TokenType.DM_Preproc_Number;
+                    }
+                    case TokenType.DM_Preproc_Punctuator:
+                        if(token.Text == ";") {
```
wixoaGit, line 153: Make this its own punctuator token.

Once you do that you can also merge this with the `case TokenType.DM_Preproc_Punctuator_RightParenthesis` below by replacing the `_canUseDirective = false;` with `_canUseDirective = (token.Type == TokenType.DM_Preproc_Punctuator_Semicolon);`.

wixoaGit (changes requested):

OpenDreamShared/Compiler/Token.cs
```diff
@@ -29,6 +29,7 @@ public enum TokenType : byte {
         DM_Preproc_Pragma,
         DM_Preproc_Punctuator,
         DM_Preproc_Punctuator_Colon,
+        DM_Preproc_Punctuator_Semicolon,
```
wixoaGit, line 32: Keep this in alphabetical order, move this below `DM_Preproc_Punctuator_RightParenthesis`

wixoaGit (changes requested):

wixoaGit (approved):

PR #664, opened 2022-04-12
Faster Typechecking - It's Finished Edition

ike709:
Closes #587

I wanted this to get in before I nuke DreamPath. Credit to @ZeWaka for doing basically all the work.

PR #963, opened 2022-12-14
Faster Typechecking - It's Finished Again Edition

wixoaGit:
Basically #664 which is a continuation of #587

- `DreamObjectDefinition.IsSubtypeOf()` was reworked to take a `IDreamObjectTree.TreeEntry` instead of a `DreamPath` and now operates in O(1) time.
- `IDreamObjectTree` is now IoC-resolvable instead of sitting in `IDreamManager`.
- `IDreamObjectTree.Atom` now takes the place of `DreamPath.Atom` (I hope to replace more uses of `DreamPath` with this later).
- DM enumerators now implement our own `IDreamValueEnumerator` interface instead of `IEnumerator<DreamValue>` to speed up enumeration.

This greatly speeds up anything that does a lot of inheritance checking, most notably the filtered DreamObject enumerator.

`Enumerate` opcode before:

image
After:

image

### DIFF
--- a/OpenDreamRuntime/Objects/DreamObjectDefinition.cs
+++ b/OpenDreamRuntime/Objects/DreamObjectDefinition.cs
@@ -8,7 +8,7 @@ namespace OpenDreamRuntime.Objects {
 
         public DreamPath Type => _treeNode.Path;
         public DreamObjectDefinition? Parent => _treeNode.ParentEntry?.ObjectDefinition;
-        public IDreamMetaObject MetaObject = null;
+        public IDreamMetaObject? MetaObject = null;
         public int? InitializationProc;
         public readonly Dictionary<string, int> Procs = new();
         public readonly Dictionary<string, int> OverridingProcs = new();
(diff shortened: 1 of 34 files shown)


ZeWaka (commented):
Much better implementation than mine!

OpenDreamRuntime/Objects/DreamObjectDefinition.cs
```diff
@@ -1,16 +1,14 @@
...
-    public sealed class DreamObjectDefinition
-    {
+    public sealed class DreamObjectDefinition {
         [Dependency] private readonly IDreamManager _dreamMan = default!;
 
-        public DreamPath Type;
-        public IDreamMetaObject? MetaObject = null;
+        public DreamPath Type => _treeNode.Path;
+        public DreamObjectDefinition? Parent => _treeNode.ParentEntry?.ObjectDefinition;
+        public IDreamMetaObject MetaObject = null;
```
wixoaGit, line 11:
```suggestion
        public IDreamMetaObject? MetaObject = null;
```

PR #967, opened 2022-12-16
Add disassembly support to the debugger

SpaceManiac:
- Implement the Disassemble request
  - Move instruction stringification from DMDisassembler to shared ProcDecoder
  - Use hex addresses for jump targets for consistency with VSC's display
  - VSC requires this to be modelled as infinitely scrolling, so pad with 1s and ~0s
- Implement instruction stepping (but not instruction breakpoints)
  - Include the DMProcState value stack in the Scopes list when non-empty
  - Add ToString and variable inspection for DreamProcArguments

To access in VSC, use "Open Disassembly View" in either the Call Stack or editor context menu while paused.


OpenDreamRuntime/Procs/DebugAdapter/DreamDebugManager.cs
```diff
@@ -105,10 +106,56 @@ private struct ActiveBreakpoint {
...
+                    Reason = StoppedEvent.ReasonFunctionBreakpoint,
+                    HitBreakpointIds = hit,
+                });
+                return;
+            }
+        }
+
+        // Stop if we're instruction stepping.
+        bool stoppedOnStep = false;
+        switch (_threadStepModes.GetValueOrDefault(state.Thread.Id)) {
```
wixoaGit, line 136: This extra dictionary index on every instruction is slowing things down. Is there some other way to do this?

wixoaGit (changes requested):

OpenDreamRuntime/Procs/DebugAdapter/DreamDebugManager.cs
ike709: Or worse case scenario only do it in debug?

SpaceManiac: - It could be stored on DreamThread directly, if you're willing for debugger state to leak there.
- We could skip the check if no debug client is attached, or if _threadStepModes is empty, or similar.
- I would not recommend putting it behind `#if DEBUG` since some may want to debug their DM code without debugging OD itself.

wixoaGit: I'm happy with that DreamThread idea.

PR #972, opened 2022-12-17
Implements trimtext

### LINKED ISSUE #856: `trimtext()` not implemented
Trims whitespace from both ends of a text string.

All [Unicode](https://www.byond.com/docs/ref/info.html#/{notes}/Unicode) whitespace characters are counted, whether they can cause a break or not.

https://www.byond.com/docs/ref/#/proc/trimtext

PaulRitter:
fixes #856


OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -2321,5 +2321,13 @@ public static DreamValue NativeProc_viewers(DreamObject instance, DreamObject us
             connection.WinSet(winsetControlId, winsetParams);
             return DreamValue.Null;
         }
+
+        [DreamProc("trimtext")]
+        [DreamProcParameter("text", Type = DreamValueType.String)]
+        public static DreamValue NativeProc_trimtext(DreamObject instance, DreamObject usr,
```
wixoaGit, line 2327: Native procs are not automatically detected. They must be set, which you can do in [this file](https://github.com/OpenDreamProject/OpenDream/blob/8af1680d1eae960ba9a117a82faa637bcacaab29/OpenDreamRuntime/Procs/Native/DreamProcNative.cs).

You also need to make a stub for it in [DMStandard](https://github.com/OpenDreamProject/OpenDream/blob/8af1680d1eae960ba9a117a82faa637bcacaab29/DMCompiler/DMStandard/_Standard.dm) so the compiler is aware of its existence.

wixoaGit (changes requested):

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -2321,5 +2321,13 @@ public static DreamValue NativeProc_viewers(DreamObject instance, DreamObject us
...
+            DreamProcArguments arguments)
+        {
+            return new DreamValue(arguments.GetArgument(0, "text").GetValueAsString().Trim());
```
ike709, line 2330:
```suggestion
            return new DreamValue(arguments.GetArgument(0, "text").MustGetValueAsString().Trim());
```
`Get` is obsolete, all calls need to use `MustGet` or `TryGet`

wixoaGit: `TryGetValueAsString()` would be preferable here so you return `DreamValue.Null`, which is what BYOND does.

wixoaGit (approved):

PR #975, opened 2022-12-18
Add `TryGetNamedArgument` and `TryGetPositionalArgument` to `DreamProcArguments`

redmoogle:
Adds some useful helpers to try and get values without having to worry about certain derefs

### DIFF
--- a/OpenDreamRuntime/Procs/DreamProcArguments.cs
+++ b/OpenDreamRuntime/Procs/DreamProcArguments.cs
@@ -37,34 +37,22 @@ namespace OpenDreamRuntime.Procs {
             return DreamValue.Null;
         }
 
-        public bool TryGetNamedArgument(string argumentName, [NotNullWhen(true)] out DreamValue? value) {
+        public bool TryGetNamedArgument(string argumentName, out DreamValue value) {
             if(NamedArguments == null) {
                 value = DreamValue.Null;
                 return false;
             }
 
-            if(NamedArguments.TryGetValue(argumentName, out DreamValue lookup)) {
-                value = lookup;
-                return true;
-            }
-
-            value = DreamValue.Null;
-            return false;
+            return NamedArguments.TryGetValue(argumentName, out value);
         }
 
-        public bool TryGetPositionalArgument(int index, [NotNullWhen(true)] out DreamValue? value) {
+        public bool TryGetPositionalArgument(int index, out DreamValue value) {
             if(OrderedArguments == null) {
                 value = DreamValue.Null;
                 return false;
             }
 
-            if(OrderedArguments.TryGetValue(index, out DreamValue lookup)) {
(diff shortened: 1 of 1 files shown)


OpenDreamRuntime/Procs/DreamProcArguments.cs
```diff
@@ -34,5 +36,35 @@ namespace OpenDreamRuntime.Procs {
             }
             return DreamValue.Null;
         }
+
+        public bool TryGetNamedArgument(string argumentName, [NotNullWhen(true)] out DreamValue? value) {
```
wixoaGit, line 40: You never actually set this to `null` (`DreamValue.Null` is still a value).
```suggestion
        public bool TryGetNamedArgument(string argumentName, out DreamValue value) {
```

OpenDreamRuntime/Procs/DreamProcArguments.cs
```diff
@@ -34,5 +36,35 @@ namespace OpenDreamRuntime.Procs {
...
+            if(NamedArguments.TryGetValue(argumentName, out DreamValue lookup)) {
+                value = lookup;
+                return true;
+            }
+
+            value = DreamValue.Null;
+            return false;
+        }
+
+        public bool TryGetPositionalArgument(int index, [NotNullWhen(true)] out DreamValue? value) {
```
wixoaGit, line 55:
```suggestion
        public bool TryGetPositionalArgument(int index, out DreamValue value) {
```

OpenDreamRuntime/Procs/DreamProcArguments.cs
```diff
@@ -34,5 +36,35 @@ namespace OpenDreamRuntime.Procs {
...
+                return false;
+            }
+
+            if(NamedArguments.TryGetValue(argumentName, out DreamValue lookup)) {
+                value = lookup;
+                return true;
+            }
+
+            value = DreamValue.Null;
+            return false;
```
wixoaGit, line 52: This can be shortened to one line
```suggestion
            return NamedArguments.TryGetValue(argumentName, out value);
```

OpenDreamRuntime/Procs/DreamProcArguments.cs
```diff
@@ -34,5 +36,35 @@ namespace OpenDreamRuntime.Procs {
...
+                return false;
+            }
+
+            if(OrderedArguments.TryGetValue(index, out DreamValue lookup)) {
+                value = lookup;
+                return true;
+            }
+
+            value = DreamValue.Null;
+            return false;
```
wixoaGit, line 67:
```suggestion
            return OrderedArguments.TryGetValue(index, out value);
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #976, opened 2022-12-18
one-sided const-eval for `&&` and `||`

PaulRitter:
what it says on the tin

### DIFF
--- a/DMCompiler/DM/Expressions/Binary.cs
+++ b/DMCompiler/DM/Expressions/Binary.cs
@@ -381,17 +381,13 @@ namespace DMCompiler.DM.Expressions {
             : base(location, lhs, rhs) { }
 
         public override bool TryAsConstant(out Constant constant) {
-            var lhsIsConst = LHS.TryAsConstant(out var lhs);
-            var rhsIsConst = RHS.TryAsConstant(out var rhs);
-
-            if (lhsIsConst && rhsIsConst) {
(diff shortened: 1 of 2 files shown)


DMCompiler/DM/Expressions/Binary.cs
```diff
@@ -381,13 +381,22 @@ public Or(Location location, DMExpression lhs, DMExpression rhs)
             : base(location, lhs, rhs) { }
 
         public override bool TryAsConstant(out Constant constant) {
```
wixoaGit, line 383: It would more closely mirror the behavior of `a || b` if you did this instead:

1. Check if LHS is constant and truthy, setting `constant = lhs` if it is.
2. Otherwise, set `constant = rhs` if RHS is constant.

```cs
if (LHS.TryAsConstant(out var lhs) && lhs.IsTruthy()) {
    constant = lhs;
    return true;
}

if (RHS.TryAsConstant(out var rhs)) {
    constant = rhs;
    return true;
}

constant = null;
return false;
```

Something similar could be done for `And.TryAsConstant()`. `Constant.Or()` and `Constant.And()` could be removed since these are the only uses of them.

PaulRitter: good point, that'd further reduce the or in cases where both werent true.

wixoaGit (approved):

PR #978, opened 2022-12-18
dead code detection

PaulRitter:
not sure if i like this implementation as it makes the procbuilder more noisy imo, but this was the first solution to the problem i came up with

i could prepare this pr for when const-folding eventually becomes a thing, but im not sure if the emit signatures & logic will change for that, so i held off on that for now and added two todos


ike709:
We already talked about it on Discord but I figured I should mention here that it currently seems to be broken due to switch cases.

ike709:
Fails CI

wixoaGit:
I'm not a fan of this solution. I think once we keep a control flow graph, it would be better to add dead code detection through that.

PR #984, opened 2022-12-20
implements try catch

### LINKED ISSUE #424: Implement `throw` properly and actually catching errors properly
http://www.byond.com/docs/ref/#/proc/throw

Basic try/catch support got added in https://github.com/wixoaGit/OpenDream/issues/227

But we're still missing `throw` and also it'll never actually *catch* exceptions at the moment.

PaulRitter:
fixes #424


OpenDreamRuntime/Procs/AsyncNativeProc.cs
```diff
@@ -102,7 +102,8 @@ public sealed class State : ProcState {
                 Result = await _taskFunc(this);
             }
 
-            protected override ProcStatus InternalResume() {
+            public override ProcStatus Resume()
+            {
```
wixoaGit, line 106:
```suggestion
            public override ProcStatus Resume() {
```

OpenDreamRuntime/DreamThread.cs
```diff
@@ -203,7 +212,41 @@ public sealed class DreamThread {
...
+                    } catch (DMRuntime dmRuntime) {
+                        if (TryCatchException(dmRuntime.Value)) continue;
+                        HandleException(dmRuntime);
+                        status = ProcStatus.Returned;
+                    } catch (DMError dmError) {
+                        CancelAll();
+                        HandleException(dmError);
+                        status = ProcStatus.Cancelled;
+                    } catch (Exception exception) {
+                        if (TryCatchException(new DreamValue(exception.Message))) continue;
```
wixoaGit, line 244: BYOND appears to turn any exception that isn't from a `throw` into an [/exception object](https://www.byond.com/docs/ref/#/exception).

OpenDreamRuntime/Procs/DMProc.cs
```diff
@@ -189,6 +192,9 @@ sealed class DMProcState : ProcState {
         public int ArgumentCount;
         public string? CurrentSource;
         public int CurrentLine;
+        public int CatchPosition = NoTryCatch;
+        public int CatchVarIndex = NoTryCatch;
```
wixoaGit, line 196: These should be reset in either `Dispose()` or `Initialize()`

Proc states can be saved for later reuse, so it's good to not keep values between uses.

OpenDreamRuntime/Procs/DMProc.cs
```diff
@@ -331,6 +348,30 @@ sealed class DMProcState : ProcState {
             return thread;
         }
 
+        public void StartTryBlock(int catchPosition, int catchVarIndex = NoTryCatch) {
+            CatchPosition = catchPosition;
+            CatchVarIndex = catchVarIndex;
```
wixoaGit, line 353: This wouldn't support nested try/catches

OpenDreamRuntime/Procs/NativeProc.cs
```diff
@@ -49,7 +49,8 @@ public sealed class State : ProcState {
                 Arguments = arguments;
             }
 
-            protected override ProcStatus InternalResume() {
+            public override ProcStatus Resume()
+            {
```
wixoaGit, line 53:
```suggestion
            public override ProcStatus Resume() {
```

wixoaGit (changes requested):

wixoaGit:
The `Try` and `TryNoValue` opcodes will need added to the `ProcDecoder` class, which is used by both the debugger and the decompiler.

PaulRitter:
> The `Try` and `TryNoValue` opcodes will need added to the `ProcDecoder` class, which is used by both the debugger and the decompiler.

Im unsure where i need to add these opcodes in the file, as well as how to test it. Can you ping me on discord with instructions?

OpenDreamRuntime/DreamThread.cs
```diff
@@ -86,16 +86,32 @@ public abstract class DreamProc {
...
+        }
+
+        public DMRuntime(DreamValue value) : base($"'throw' thrown ({value})") {
+            Value = value;
+        }
+    }
+
+    sealed class DMCrashRuntime : DMRuntime {
+
+        public DMCrashRuntime(string message) : base(new DreamValue(message)) { }
```
wixoaGit, line 104: Using this base constructor will end up putting `'throw' thrown` at the beginning of the exception message. Probably meant to use the other one?
```suggestion
        public DMCrashRuntime(string message) : base(message) { }
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #985, opened 2022-12-20
Make `DreamList._values` null by default

### LINKED ISSUE #629: Don't allocate `DreamList._values` until necessary
A continuation of https://github.com/wixoaGit/OpenDream/pull/628 would be to set `DreamList._values` to null whenever it's not actually storing any dreamvalues.

PaulRitter:
also adjusted the GetValues call because the alternative would have been allocating and returning an empty list on each call
fixes #629


ike709:
See https://github.com/PaulRitter/OpenDream/pull/2

ike709:
Fails CI

PaulRitter:
i dont know why the build is failing. the file in question does import system.linq which provides the two methods it claims are not found. also the linenumbers of the errors are wrong. seems like gh actions fuckery, can you rerun the workflows?

redmoogle:
> i dont know why the build is failing. the file in question does import system.linq which provides the two methods it claims are not found. also the linenumbers of the errors are wrong. seems like gh actions fuckery, can you rerun the workflows?

pull master and fix the missing linq using

OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectList.cs
```diff
@@ -1,4 +1,5 @@
-﻿using OpenDreamRuntime.Procs;
+﻿using System.Linq;
```
PaulRitter, line 1: @redmoogle its not missing though

PaulRitter:
mastermerge cant hurt tho

OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectList.cs
PaulRitter: nvm

OpenDreamRuntime/Objects/DreamList.cs
```diff
@@ -60,32 +65,43 @@ public class DreamList : DreamObject {
...
+                return GetValue(keyInteger - 1); //1-indexed
             }
             if (_associativeValues == null)
                 return DreamValue.Null;
 
             return _associativeValues.TryGetValue(key, out DreamValue value) ? value : DreamValue.Null;
         }
 
+        public virtual DreamValue GetValue(int key) {
+            if (_values == null) throw new IndexOutOfRangeException();
```
Insineer, line 77: This PR adds more additional checks making the code more complex and increasing the room for bugs in the future. But, conversely, the optimization it brings up is very questionable and requires to be profiled (see #629).

I would suggest closing this PR and related issue until we will see any optimization problems in any performance test with profiling.

Insineer (changes requested):

OpenDreamRuntime/Objects/DreamList.cs
wixoaGit: I agree, the topic of whether or not this is actually an optimization has come up a few times and I don't see proof that it is.

Turning `GetValues()` into an enumerable is good though, cuts down on large allocations in a few places. Would be good to get that in another PR.

PR #988, opened 2022-12-21
Optimize `world.contents`

### LINKED ISSUE #633: `DreamList.RemoveValue()` is really slow
So slow that it's at the top of the profiler.

wixoaGit:
Fixes #633 

A list of all atoms is now held in `DreamMapManager` instead of a DreamList on the `world` object. This speeds up the creation/deletion of atoms.

There is also now a special enumerator for `world.contents` which greatly speeds up loops like `for(var/atom/A in world)` by preventing an unnecessary copy of the huge list.

The `CreateFilteredEnumerator` opcode during Paradise initialization,

Before:
image
After:
image

### DIFF
--- a/OpenDreamRuntime/DreamMapManager.cs
+++ b/OpenDreamRuntime/DreamMapManager.cs
@@ -57,7 +57,7 @@ namespace OpenDreamRuntime {
 
         public Vector2i Size { get; private set; }
         public int Levels => _levels.Count;
-        public List<DreamObject> AllAtoms { get; } = new();
+        public List<DreamObject> AllAtoms { get; } = new(410000); // ~400K elements after Paradise init
 
         private readonly List<Level> _levels = new();
         private readonly Dictionary<DreamObject, (Vector2i Pos, Level Level)> _turfToTilePos = new();
--- a/OpenDreamRuntime/DreamMapManager.cs
+++ b/OpenDreamRuntime/DreamMapManager.cs
@@ -51,13 +51,13 @@ namespace OpenDreamRuntime {
         };
 
         [Dependency] private readonly IMapManager _mapManager = default!;
-        [Dependency] private readonly IDreamManager _dreamManager = default!;
         [Dependency] private readonly IDreamObjectTree _objectTree = default!;
         [Dependency] private readonly IEntitySystemManager _entitySystemManager = default!;
         private ServerAppearanceSystem _appearanceSystem = default!;  // set in Initialize
 
         public Vector2i Size { get; private set; }
         public int Levels => _levels.Count;
+        public List<DreamObject> AllAtoms { get; } = new();
 
         private readonly List<Level> _levels = new();
         private readonly Dictionary<DreamObject, (Vector2i Pos, Level Level)> _turfToTilePos = new();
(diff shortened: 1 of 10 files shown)


OpenDreamRuntime/DreamMapManager.cs
```diff
@@ -51,13 +51,13 @@ public sealed class Cell {
...
 
         [Dependency] private readonly IMapManager _mapManager = default!;
-        [Dependency] private readonly IDreamManager _dreamManager = default!;
         [Dependency] private readonly IDreamObjectTree _objectTree = default!;
         [Dependency] private readonly IEntitySystemManager _entitySystemManager = default!;
         private ServerAppearanceSystem _appearanceSystem = default!;  // set in Initialize
 
         public Vector2i Size { get; private set; }
         public int Levels => _levels.Count;
+        public List<DreamObject> AllAtoms { get; } = new();
```
ike709, line 60: Probably worth pre-allocating a large initial capacity.

wixoaGit: The size of this list after Paradise init was 400252. I'll go ahead and have this pre-allocate 410,000 elements.

OpenDreamRuntime/Objects/DreamList.cs
```diff
@@ -414,4 +414,43 @@ public sealed class DreamFilterList : DreamList {
             return appearance;
         }
     }
+
+    // world.contents list
+    // Operates on a list of all atoms
+    public sealed class WorldContentsList : DreamList {
```
PaulRitter, line 420: won't this still allocate a unused list due to inheriting dreamlist?

wixoaGit: It would. So do all the other special lists, it would be a good idea in the future to make DreamList into an abstract class or something similar.

OpenDreamRuntime/DreamMapManager.cs
```diff
@@ -51,13 +51,13 @@ public sealed class Cell {
...
 
         [Dependency] private readonly IMapManager _mapManager = default!;
-        [Dependency] private readonly IDreamManager _dreamManager = default!;
         [Dependency] private readonly IDreamObjectTree _objectTree = default!;
         [Dependency] private readonly IEntitySystemManager _entitySystemManager = default!;
         private ServerAppearanceSystem _appearanceSystem = default!;  // set in Initialize
 
         public Vector2i Size { get; private set; }
         public int Levels => _levels.Count;
+        public List<DreamObject> AllAtoms { get; } = new(410000); // ~400K elements after Paradise init
```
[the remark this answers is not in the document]
wixoaGit: It would be. I'm still not entirely convinced this is a worthwhile optimization though, still need to measure its impact.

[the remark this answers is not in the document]
wixoaGit: I profiled this. Creating 400K atoms only allocates an extra 8MB when not pre-allocated. It also spends a total time of 1-2ms adding to the list either way.

There doesn't seem to be a good reason to be worried about pre-allocating this list.

PR #991, opened 2022-12-22
Make opcodes a dictionary

amylizzle:
Makes opcodes into a dictionary of `DreamProcOpcode` to `DMOpcodeHandler` so we can easily check that opcodes haven't been deleted or are still in use. 

Performance impact is negligible, since it's converted to a static array at runtime.

Should just make things a little more maintainable.


OpenDreamRuntime/Procs/DMProc.cs
```diff
@@ -202,6 +199,21 @@ sealed class DMProcState : ProcState {
 
         public override (string?, int?) SourceLine => (CurrentSource, CurrentLine);
 
+        /// Static initialiser for maintainer friendly OpcodeHandlers to performance friendly _opcodeHandlers
+        static DMProcState()
+        {
```
wixoaGit, line 204: Keep braces on the same line

wixoaGit (changes requested):

wixoaGit (approved):

PR #999, opened 2022-12-26
Implement a sane alternative to `.....`

### LINKED ISSUE #729: Implement `.....`
image

ike709:
Waka said that they're open to just using a sane alternative to `.....` in gooncode instead of trying to replicate it. So I did that.

Resolves #729

### DIFF
--- a/Content.Tests/DMProject/Tests/Procs/opendream_procpath.dm
+++ b/Content.Tests/DMProject/Tests/Procs/opendream_procpath.dm
@@ -1,0 +1,13 @@
+
+// opendream_procpath is equivalent to ..... in BYOND
+
+/atom/movable/proc/foo()
+	return opendream_procpath
+
+/proc/bar()
+	return opendream_procpath
+
+/proc/RunTest()
+	var/atom/movable/A = new
+	ASSERT(A.foo() == "/atom/movable/proc/foo")
+	ASSERT(bar() == "/proc/bar")
--- a/Content.Tests/DMProject/Tests/Procs/opendream_procpath.dm
+++ b/Content.Tests/DMProject/Tests/Procs/opendream_procpath.dm
@@ -9,5 +9,5 @@
 
 /proc/RunTest()
 	var/atom/movable/A = new
-	ASSERT(A.foo() == "/atom/movable/proc/foo")
-	ASSERT(bar() == "/proc/bar")
+	ASSERT(A.foo() == /atom/movable/proc/foo)
+	ASSERT(bar() == /proc/bar)
--- a/DMCompiler/DM/Expressions/LValue.cs
+++ b/DMCompiler/DM/Expressions/LValue.cs
@@ -41,17 +41,6 @@ namespace DMCompiler.DM.Expressions {
         }
     }
 
-    // opendream_procpath
-    class OpenDreamProcpath : LValue {
-        public OpenDreamProcpath(Location location, DreamPath? path)
-            : base(location, path)
-        {}
-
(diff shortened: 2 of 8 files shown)


ZeWaka:
Could even gate this behind a pragma if necessary

DMCompiler/DM/Visitors/DMVisitorExpression.cs
```diff
@@ -94,6 +94,9 @@ public void VisitIdentifier(DMASTIdentifier identifier)
                 case "args":
                     Result = new Expressions.Args(identifier.Location);
                     break;
+                case "opendream_procpath": // Our saner alternative to .....
```
Altoids1, line 97: Can I hit you with `"_od_builtin_procpath"`? Might communicate that this is a compiler built-in a little better.

Altoids1 (changes requested):

DMCompiler/DM/Visitors/DMVisitorExpression.cs
ike709: Everyone can bikeshed it on Discord to their heart's content and then when the group reaches consensus I'll use whatever everyone agrees with.

ike709: Considering the formatting of other OD stuff like `opendream_compiletimereadonly`, and the lack of any real interest from anyone else in arguing over the name, I'm kind of leaning towards just keeping it as-is tbh.

DMCompiler/DM/Expressions/LValue.cs
```diff
@@ -41,6 +41,17 @@ public Src(Location location, DreamPath? path)
         }
     }
 
+    // opendream_procpath
+    class OpenDreamProcpath : LValue {
```
wixoaGit, line 45: This shouldn't be an LValue since it's not an assignable value. This could probably be a normal `DMExpression` that pushes the proc's path at compile-time. Maybe give it a `TryAsConstant()` too?

wixoaGit (changes requested):

OpenDreamRuntime/Procs/DMProc.cs
```diff
@@ -585,6 +586,9 @@ sealed class DMProcState : ProcState {
 
                     throw new Exception($"Cannot get index {index} of {indexing}");
                 }
+                case DMReference.Type.OpenDreamProcpath: {
+                    return new DreamValue(Proc.OwningType == DreamPath.Root ? "/proc/" + Proc.ToString() : Proc.ToString());
```
wixoaGit, line 590: This value should be a `DreamProc` instead of a string.

ZeWaka: for the reasoning here, see the discord

Content.Tests/DMProject/Tests/Procs/opendream_procpath.dm
```diff
@@ -0,0 +1,13 @@
...
+/atom/movable/proc/foo()
+	return opendream_procpath
+
+/proc/bar()
+	return opendream_procpath
+
+/proc/RunTest()
+	var/atom/movable/A = new
+	ASSERT(A.foo() == "/atom/movable/proc/foo")
+	ASSERT(bar() == "/proc/bar")
```
ike709, line 13:
```suggestion
	ASSERT(A.foo() == /atom/movable/proc/foo)
	ASSERT(bar() == /proc/bar)
```

wixoaGit (approved):

PR #1016, opened 2023-01-04
Allow for numeric components in macro define paths

amylizzle:
Allows macros to define paths like `/obj/thing_3/`


DMCompiler/Compiler/DM/DMLexer.cs
```diff
@@ -30,6 +30,11 @@ public sealed class DMLexer : TokenLexer {
             "..."
         };
 
+        private static readonly List<TokenType> ValidIdentifierComponents = new(2) {
+            TokenType.DM_Preproc_Identifier,
+            TokenType.DM_Preproc_Number
```
ike709, line 35: Does checking 2 types in 1 spot really justify moving this to a list?

amylizzle: Not really, but I was thinking this would be a much better place to handle the "operator[]" nonsense that I did kinda hackily in the operator overloads PR.

Which is to say, I intend to add a bunch more things to this list.

Content.Tests/DMProject/Tests/Preprocessor/macro_numeric_path.dm
```diff
@@ -0,0 +1,15 @@
+/obj/thing_1/dodaa
+	name = @#underscore 1 test#
```
wixoaGit, line 2: What does this have to do with the test?

wixoaGit (changes requested):

Content.Tests/DMProject/Tests/Preprocessor/macro_numeric_path.dm
amylizzle: nothing, the erroring code was using a rawstring, and I didn't see a need to remove it.

wixoaGit: This should be removed to keep the test focused on the macro problem. If you want to test raw strings, a new test for it could be created under `Tests/Text/`.

amylizzle: Sure thing, I'll do it now

wixoaGit: I meant the `/obj/thing_1/dodaa` in general

amylizzle: oh, the error was occurring when a path was defined as `/obj/thing_##NUM` where `##NUM` was numeric. It didn't error on fully defined paths like `/obj/thing_1`. I was just making sure I didn't alter that behaviour.

wixoaGit (approved):

PR #1017, opened 2023-01-04
Patchworks unit tests so that more actually work

redmoogle:
Yeah something isnt being reset right but looking through the debugger doesnt yield much useful info

_basic.dm is needed due to `ArgsEqual.dm` having proc definitions (which whatever bug is causing this is taking the procs from the first json loaded)


redmoogle:
seems to only fix it for single runs and not a full test, looks like the proclist will have to be cleared every time then remade (basically just call Preinitialize for every test or add a way to clear all procs then readd natives)

redmoogle:
nvm it actually just works fine just the extra test I tried to port over dont

Content.Tests/DMProject/Tests/List/ListCut.dm
```diff
@@ -0,0 +1,5 @@
+﻿/proc/RunTest()
+	var/list/A = list(1,2,3)
+	ASSERT(A.len == 3)
+	A.Cut()
+	ASSERT(A.len == 0)
```
ike709, line 5: Pretty sure we still don't have a test that confirms cutting an _empty_ list isn't an out of bounds exception.

Content.Tests/DMProject/Tests/List/ListCut.dm
```diff
@@ -0,0 +1,6 @@
+﻿/proc/RunTest()
+	var/list/A = list(1,2,3)
+	ASSERT(A.len == 3)
+	A.Cut()
+	A.Cut()
```
ike709, line 5:
```suggestion
```

I'll just do it properly myself sometime, I give up.

Content.Tests/DMTests.cs
```diff
@@ -114,24 +111,25 @@ public void OneTimeSetup()
...
                     return DreamValue.Null;
                 }
             });
 
+
             var Watch = new Stopwatch();
             Watch.Start();
 
-            // Tick until our inner call has finished
+             // Tick until our inner call has finished
```
wixoaGit, line 132: Unnecessary space created
```suggestion
            // Tick until our inner call has finished
```

Content.Tests/DMTests.cs
```diff
@@ -114,24 +111,25 @@ public void OneTimeSetup()
...
-                    result = await callTask;
+                    retValue = await callTask;
                     return DreamValue.Null;
                 } else {
                     Assert.Fail($"No global proc named RunTest");
                     return DreamValue.Null;
                 }
             });
 
+
```
wixoaGit, line 128: Unnecessary line created
```suggestion
```

Content.Tests/DMProject/environment.dme
```diff
@@ -6,7 +6,7 @@
 // BEGIN_PREFERENCES
 // END_PREFERENCES
 // BEGIN_INCLUDE
-#include "code.dm"
 #include "interface.dmf"
 #include "map.dmm"
+#include "code.dm"
```
wixoaGit, line 11: The includes in a DME are typically kept in alphabetical order

Content.Tests/DMProject/code.dm
```diff
@@ -1,63 +1,2 @@
...
-
-/world/proc/switch_const()
-	var/a = 137
-	switch(a)
-		if(20)
-			. = 500
-		if(136 | 1)
-			. = 1
-		if(/datum, /mob)
-			. = 300
```
wixoaGit, line 63: No equivalent test was created to replace this

wixoaGit (changes requested):

Content.Tests/DMProject/Tests/Const/Const6.dm
```diff
@@ -0,0 +1,10 @@
+﻿/proc/RunTest()
+	var/a = 137
+	switch(a)
+		if(20)
+			. = 500
+		if(136 | 1)
+			. = 1
+		if(/datum, /mob)
+			. = 300
+	ASSERT(a == 137)
```
wixoaGit, line 10: This should do `ASSERT(. == 1)` because this aims to verify that `if(136 | 1)` is executing. `a` is not meant to ever change.
```suggestion
	ASSERT(. == 1)
```

Content.Tests/DMProject/code.dm
```diff
@@ -1,63 +0,0 @@
-// TODO Move all of these into test files
```
wixoaGit, line 1: Might as well delete this file too, if it's empty and nothing is using it.

Content.Tests/DMTests.cs
```diff
@@ -169,8 +169,7 @@ private static IEnumerable<object[]> GetTests()
             DMTestFlags testFlags = DMTestFlags.NoError;
 
             using (StreamReader reader = new StreamReader(sourceFile)) {
-                string firstLine = reader.ReadLine();
-
+                string firstLine = reader.ReadLine() ?? throw new System.Diagnostics.UnreachableException("reader.ReadLine() returned null");
```
wixoaGit, line 172: This is not necessarily unreachable, is it? Could be an empty file.

wixoaGit (changes requested):

Content.Tests/DMTests.cs
redmoogle: Throw a error pointing to the file when its null then? I kinda want the compiler to not null warn for this

wixoaGit: Could return early if the file is empty
```cs
string? firstLine = reader.ReadLine();
if (firstLine == null)
    return testFlags;
```

wixoaGit: An error works too though

wixoaGit (approved):

PR #1020, opened 2023-01-07
refactors argumentlists to not allocate on the stack

PaulRitter:
deposits the value list in this structure on the stack (listing from top to bottom)

```
[orderedArgVal1]
...
[orderedArgValO]
[namedArgName1]
[namedArgValue1]
...
[namedArgNameN]
[namedArgValueN]
```

the count of both ordered and named args is written to the bytecode after the opcode. if an arglist is used (which we dont know the counts of on compiletime), the values (-2, 1) are written to the bytecode. (the exact values were used because when piped into the grow/shrinkstack logic, they resolve to a stacksize change of 0)

when an arglist is used, the list ontop of the stack will be resolved into dreamprocargs.

unsure about the shrink/growstack statements since i havent done that yet but i think i used them correctly


wixoaGit:
I'm not sure if this actually provides much of an optimization, if any. The bulk of PushArguments comes from allocating the structures within DreamProcArguments itself, not from pushing it onto the stack (particularly since pushing to the stack is cheap; we can predict its maximum size and it's already pooled).

A more worthwhile optimization would need to reduce the allocations of the OrderedArguments List and NamedArguments Dictionary. Object-pooling those two or even DreamProcArguments itself would likely see a huge benefit.

wixoaGit:
I'm closing this due to the reasons mentioned above.

PR #1041, opened 2023-01-20
implements winclone

### LINKED ISSUE #281: `winclone()` not implemented
http://www.byond.com/docs/ref/#/proc/winclone

PaulRitter:
resolves #281

not tested ingame yet bc i dont know which codebases use this to what extend, and i didnt have time to setup my own testing project atm.

i am slightly confused by the byond ref since it first specifies that the only valid inputs to winclone are `a window, pane, menu, or macro set in the world's skin file`, however later down the spec they talk about `Menus or macros must be assigned to a window with winset() before they will work.`, implying you can clone single macros too. I opted to interpret this as a typo.

## Deviations from the byond spec implementation
- `The original object as it exists in the skin file (not its current state) is used as a template to build the clone.`: Right now elementdescriptors **are mutable** but are never mutated, so while theoretically possible, this isnt implemented. Is there anything stopping us from making elementdescriptors readonly (with some necessary exceptions, namely Name and isVisible)?


OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -357,25 +362,81 @@ sealed class DreamInterfaceManager : IDreamInterfaceManager {
...
+                    break;
+                case MacroSetDescriptor interfaceMacroSet:
+                    _macroManager.LoadMacroSet(interfaceMacroSet);
+                    break;
+            }
+        }
+
         private void LoadInterface(InterfaceDescriptor descriptor) {
-            InterfaceDescriptor = descriptor;
+            InterfaceDescriptor = descriptor.Copy(_serializationManager);
```
wixoaGit, line 429: The InterfaceDescriptor is never modified, there should be no need to copy it.

wixoaGit, line 429: The InterfaceDescriptor is never modified, there should be no need to copy it.

OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -357,25 +362,81 @@ sealed class DreamInterfaceManager : IDreamInterfaceManager {
             }
         }
 
+        public void WinClone(string controlId, string cloneId) {
+            ElementDescriptor elementDescriptor = null;
+            // window/pane
+            foreach (var windowDescriptor in InterfaceDescriptor.WindowDescriptors) {
+                if (windowDescriptor.Name == controlId) {
+                    elementDescriptor = windowDescriptor.WithName(_serializationManager, cloneId);
+                    ((WindowDescriptor)elementDescriptor).IsVisible = false; // per byond spec
```
wixoaGit, line 371: Make the `ControlWindow.WindowDescriptor` property public and use that instead of casting here.

OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -357,25 +362,81 @@ sealed class DreamInterfaceManager : IDreamInterfaceManager {
             }
         }
 
+        public void WinClone(string controlId, string cloneId) {
```
wixoaGit, line 365: This misses the ability to clone UI controls (buttons, labels, etc).

wixoaGit (changes requested):

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -2495,5 +2495,27 @@ public static DreamValue NativeProc_viewers(DreamObject instance, DreamObject us
...
         }
+
+        [DreamProc("winclone")]
+        [DreamProcParameter("player", Type = DreamValueType.DreamObject)]
+        [DreamProcParameter("window_name", Type = DreamValueType.String)]
+        [DreamProcParameter("clone_name", Type = DreamValueType.String)]
+        public static DreamValue NativeProc_winclone(DreamObject instance, DreamObject usr, DreamProcArguments arguments) {
+            DreamValue player = arguments.GetArgument(0, "player");
+            var windowName = arguments.GetArgument(1, "window_name").MustGetValueAsString();
+            var cloneName = arguments.GetArgument(2, "clone_name").MustGetValueAsString();
```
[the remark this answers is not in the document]
PaulRitter: how would i throw an error here? regular throw? which exception should i use?

OpenDreamClient/Interface/DreamInterfaceManager.cs
PaulRitter: the `windowDescriptor` variable is already typed correctly, however due to the WithName method returning ElementDescriptor (a constraint caused by it being defined on the elementdescriptor type), the elementDescriptor variable has to be cast here

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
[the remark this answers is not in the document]
wixoaGit: Any exception will be caught and treated as a DM runtime. `ArgumentException` might fit here.

PaulRitter:
all reviews should be adressed

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -2489,5 +2489,31 @@ public static DreamValue NativeProc_viewers(DreamObject instance, DreamObject us
             connection.WinSet(winsetControlId, winsetParams);
             return DreamValue.Null;
         }
+
+        [DreamProc("winclone")]
+        [DreamProcParameter("player", Type = DreamValueType.DreamObject)]
+        [DreamProcParameter("window_name", Type = DreamValueType.String)]
+        [DreamProcParameter("clone_name", Type = DreamValueType.String)]
+        public static DreamValue NativeProc_winclone(DreamObject instance, DreamObject usr, DreamProcArguments arguments) {
```
wixoaGit, line 2497: This needs to be moved above `winexists` to be kept in alphabetic order. There also needs to be a `SetGlobalNativeProc()` call in `DreamProcNative.cs` for this to be used.

OpenDreamShared/Network/Messages/MsgWinClone.cs
```diff
@@ -0,0 +1,24 @@
+﻿using Lidgren.Network;
+using Robust.Shared.Network;
+using Robust.Shared.Serialization;
+
+namespace OpenDreamShared.Network.Messages;
+
+public sealed class MsgWinClone : NetMessage {
```
wixoaGit, line 7: This needs a `_netManager.RegisterNetMessage<MsgWinClone>();` in `DreamManager.Connections.cs` before it can be sent to clients.

OpenDreamShared/Network/Messages/MsgWinClone.cs
```diff
@@ -0,0 +1,24 @@
...
+    public override MsgGroups MsgGroup => MsgGroups.EntityEvent;
+
+    public string ControlId;
+    public string CloneId;
+
+    public override void ReadFromBuffer(NetIncomingMessage buffer, IRobustSerializer serializer)
+    {
```
wixoaGit, line 14:
```suggestion
    public override void ReadFromBuffer(NetIncomingMessage buffer, IRobustSerializer serializer) {
```

OpenDreamShared/Network/Messages/MsgWinClone.cs
```diff
@@ -0,0 +1,24 @@
...
+        ControlId = buffer.ReadString();
+        CloneId = buffer.ReadString();
+    }
+
+    public override void WriteToBuffer(NetOutgoingMessage buffer, IRobustSerializer serializer)
+    {
```
wixoaGit, line 20:
```suggestion
    public override void WriteToBuffer(NetOutgoingMessage buffer, IRobustSerializer serializer) {
```

OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -390,26 +395,74 @@ sealed class DreamInterfaceManager : IDreamInterfaceManager {
...
+                    elementDescriptor = ((WindowDescriptor)elementDescriptor).WithVisible(_serializationManager, false); // per byond spec
+                    break;
+                }
+            }
+
+            // menu
+            if (elementDescriptor == null) {
+                foreach (var menuDescriptor in InterfaceDescriptor.MenuDescriptors) {
+                    if (menuDescriptor.Name == controlId) {
+                        elementDescriptor = _serializationManager.CreateCopy(menuDescriptor);
```
wixoaGit, line 413: This ends up not setting the name of the new element. I tried using `WithName()`, but that ended up creating an `ElementDescriptor` instead of a `MenuDescriptor`.

OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -390,26 +395,74 @@ sealed class DreamInterfaceManager : IDreamInterfaceManager {
...
+                        break;
+                    }
+                }
+            }
+
+            // macro set
+            if (elementDescriptor == null) {
+                foreach (var macroSetDescriptor in InterfaceDescriptor.MacroSetDescriptors) {
+                    if (macroSetDescriptor.Name == controlId) {
+                        elementDescriptor = _serializationManager.CreateCopy(macroSetDescriptor);
```
wixoaGit, line 423: This ends up not setting the name of the new element.

OpenDreamClient/Interface/Descriptors/InterfaceDescriptor.cs
```diff
@@ -13,14 +13,25 @@ public sealed class InterfaceDescriptor {
...
+        return new InterfaceDescriptor(serializationManager.CreateCopy(WindowDescriptors),
+            serializationManager.CreateCopy(MacroSetDescriptors), serializationManager.CreateCopy(MenuDescriptors));
+    }
 }
 
 [Virtual, ImplicitDataDefinitionForInheritors]
 public class ElementDescriptor {
     [DataField("type")]
-    public string Type;
+    public readonly string Type;
```
wixoaGit, line 26: The serialization manager is throwing an exception when attempting to copy descriptors because this is null. Macro sets and menus currently have no type, the correct fix would probably be to give them one.

OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -390,26 +395,74 @@ sealed class DreamInterfaceManager : IDreamInterfaceManager {
             }
         }
 
+        public void WinClone(string controlId, string cloneId) {
+            ElementDescriptor elementDescriptor = null;
+            // window/pane
+            foreach (var windowDescriptor in InterfaceDescriptor.WindowDescriptors) {
+                if (windowDescriptor.Name == controlId) {
+                    elementDescriptor = windowDescriptor.WithName(_serializationManager, cloneId);
+                    elementDescriptor = ((WindowDescriptor)elementDescriptor).WithVisible(_serializationManager, false); // per byond spec
```
wixoaGit, line 404: This cast fails because `WithName()` gives an `ElementDescriptor`

wixoaGit (changes requested):

wixoaGit (approved):

Issue #903, opened 2022-11-11
`pick()` parity bug


ike709:
In BYOND if you pass an empty list it'll runtime with a `pick() from empty list` error. OD doesn't replicate this behavior from what I can tell (I'm at work right now so I can't investigate super thoroughly).

PowerfulBacon:
It doesn't runtime if you do pick(null) however

wixoaGit:
I think it would error in OD because [this](https://github.com/OpenDreamProject/OpenDream/blob/45d41d6dbcb0b0e1e4a6c67ba710cdf7fb565106/OpenDreamRuntime/Procs/DMOpcodeHandlers.cs#L1837) would end up doing `values[0]` on an empty list. A more friendly exception would be welcome though.

Insineer:
> I think it would error in OD because [this](https://github.com/OpenDreamProject/OpenDream/blob/45d41d6dbcb0b0e1e4a6c67ba710cdf7fb565106/OpenDreamRuntime/Procs/DMOpcodeHandlers.cs#L1837) would end up doing `values[0]` on an empty list. A more friendly exception would be welcome though.

Yeah, that's right. It actually threw an exception, so I just added one with more clear error text.

PR #1046, opened 2023-01-23
fix(PickUnweighted): throw exception on empty list()

Insineer:
fix https://github.com/OpenDreamProject/OpenDream/issues/903

### DIFF
--- a/OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
+++ b/OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
@@ -1834,14 +1834,15 @@ namespace OpenDreamRuntime.Procs {
                 List<DreamValue> values;
                 if (value.TryGetValueAsDreamList(out DreamList list)) {
                     values = list.GetValues();
-                    if (values.Count == 0)
-                        throw new Exception("pick() from empty list");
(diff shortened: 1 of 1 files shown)


OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -1834,6 +1834,8 @@ private static void HandleSuffixPronoun(ref StringBuilder formattedString, ReadO
                 List<DreamValue> values;
                 if (value.TryGetValueAsDreamList(out DreamList list)) {
                     values = list.GetValues();
+                    if (values.Count == 0)
+                        throw new Exception("pick() from empty list");
```
wixoaGit, line 1838: This should be moved outside here and above the `picked = ...` line so that `pick(arglist(list()))` is caught as well.

wixoaGit (changes requested):

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
Insineer: Yeah, that's right, thanks. Fixed

wixoaGit (approved):

PR #1047, opened 2023-01-25
feat(DMObjectBuilder): support late const definition

### LINKED ISSUE #803: 'Unknown identifier' with /var/const

### LINKED ISSUE #1011: Incorrect processing order of const object vars
It seems constant vars in objects aren't getting processed in the right order.  Perhaps related to #803

Test Case:
```
/datum/foo
  var/other_var = MY_CONSTANT
  var/const/MY_CONSTANT = 1
```

BYOND: Compilation Successful, behaves as expected

OpenDream: `Unknown identifier "MY_CONSTANT"`

Insineer:
Add support for late const definitions.

1. Add new `UnknownIdentifierException`
2. Put definition statements failed to compile with `UnknownIdentifierException` to "danglingStatements" object's dictionary
3. Process all statements failed because of unknown identifier again, when identifier is finally defined

fix #1011
fix #803


Altoids1:
Does this handle the recursive case? Such as
```dm
var/const/A = B
var/const/B = C
var/const/C = A
```

Insineer:
> Does this handle the recursive case? Such as
> 
> ```dm
> var/const/A = B
> var/const/B = C
> var/const/C = A
> ```

No, of course :)
BYOND can't compile this too:

```
code.dm:177:error: B:  compile failed (possible infinite cross-reference loop)
code.dm:178:error: C:  compile failed (possible infinite cross-reference loop)
code.dm:179:error: A:  compile failed (possible infinite cross-reference loop)
```

And that makes sense, cause which values A, B and C should take as a result?

Altoids1:
It's fine that it doesn't emit as descriptive an error, I was just worried that the compiler would spin or something in such a situation.

A new unit test to show that we error in this situation would be a good addition to this PR :^)

Insineer:
> A new unit test to show that we error in this situation would be a good addition to this PR :^)

Done

Content.Tests/DMProject/Tests/Const/Const_LateDefinition.dm
```diff
@@ -0,0 +1,13 @@
...
+var/const/late_defined_const2 = late_defined_const
+var/const/late_defined_const = 1
+
+/obj/TestObj
+	var/a = 1 + const_var
+	var/const/const_var = 2
+
+/proc/RunTest()
+	var/obj/TestObj/o = new
+	return o.const_var + global_list[1]
```
wixoaGit, line 13: A test's return value isn't used for anything if there isn't a `// RETURN TRUE` comment at the top of the file. Instead, `ASSERT()` that the values are what they should be.
```suggestion
	ASSERT(o.const_var == 2)
	ASSERT(global_list[1] == 1)
```

DMCompiler/DM/DMObject.cs
```diff
@@ -29,8 +30,8 @@ class DMObject {
 
         public bool IsRoot => Path == DreamPath.Root;
 
-        public List<DMASTObjectVarOverride>? danglingOverrides = null; // Overrides waiting for the LateVarDef event to happen
-
+        // Statements waiting for LateVarDef event to happen
+        public Dictionary<string, List<DMASTStatement>> danglingStatementsByUndefinedNames = new();
```
wixoaGit, line 34: Rename this to `DanglingStatementsByUndefinedNames` to fit the naming convention used throughout the rest of the project.

wixoaGit (changes requested):

Content.Tests/DMProject/Tests/Const/Const_LateDefinition.dm
Insineer: Done

wixoaGit (approved):

PR #1048, opened 2023-01-28
Unbreak Some Tests

ike709:
I ran all the Broken Tests and these are now passing.


wixoaGit:
The `no_paren.dm` test appears to assert that no time has passed since the sleep, which is obviously incorrect. I think it only passes in BYOND due to `world.realtime`'s lack of precision, and would probably fail sporadically because of it.

We also don't support using `sleep` without parentheses, it's likely being parsed as a label and an expression.

wixoaGit (approved):

PR #1058, opened 2023-02-11
Fix a bug with DreamFilters

amylizzle:
DreamFilters were losing their type information because a compile time optimisation was replacing the serialisation manager's `CreateCopy()` with a static typed `CreateCopy<DreamFilter>()`. 

This just forces it to be dynamic, allowing filter parameters to work again.

### DIFF
--- a/OpenDreamRuntime/Objects/DreamList.cs
+++ b/OpenDreamRuntime/Objects/DreamList.cs
@@ -367,6 +367,8 @@ namespace OpenDreamRuntime.Objects {
             if (!value.TryGetValueAsDreamObjectOfType(_objectTree.Filter, out var filterObject))
                 throw new Exception($"Cannot add {value} to filter list");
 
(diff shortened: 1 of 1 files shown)


OpenDreamRuntime/Objects/DreamList.cs
```diff
@@ -367,7 +367,7 @@ public sealed class DreamFilterList : DreamList {
             if (!value.TryGetValueAsDreamObjectOfType(_objectTree.Filter, out var filterObject))
                 throw new Exception($"Cannot add {value} to filter list");
 
-            DreamFilter filter = DreamMetaObjectFilter.DreamObjectToFilter[filterObject];
+            dynamic filter = DreamMetaObjectFilter.DreamObjectToFilter[filterObject];
```
ike709, line 370: Could you drop a comment to deter regressions?

wixoaGit (approved):

PR #1063, opened 2023-02-14
Adds support for weird line endings like `if(1).` or `do:`, adds pragma for it

Altoids1:
## Summary
Noticed this while looking through OpenSS13.

## Changelog
- Fixed not being able to parse certain block inits with weird but BYOND-support characters after them, like `if(1).` or whatever.
- Added a pragma for emitting style warnings against weird extra characters in if, while, or do expressions.


Altoids1:
Closing this because I realize that the problem is more an issue with how `.` and `:` as single characters do not constitute a valid (albeit no-op) expression in the language.

PR #1081, opened 2023-03-11
Fix whitespace before `#include`

### LINKED ISSUE #1050: Indentation before #include causes undefined var error
## Description

OpenDream fails to parse first object definition in the file, when file's include line begins with indentation.

## Steps to reproduce

Add folowing line starting with indentation to TestGame .dme file:

```
    #include "exodus_announcements.dm"
```

Example "exodus_announcements.dm" file content:

```
/obj/TestObj
	a = "The Emergency Shuttle has docked with the station. You have approximately %ETD% to board the Emergency Shuttle."
```

OpenDream compiles with error:

```
Error OD0405 at exodus_announcements.dm:2:4: Cannot override undefined var "a"
```

At the same time, BYOND compiles it without any issues.

## Initial research

I've taken a look to this problem in the debugger and found that problem is in DMLexer. DMObjectBuilder tries to process `/obj/TestObj/a` override statement, and encounters wrong statement's ObjectPath:

image

ObjectPath should be `/obj/TestObj` instead of `/datum/map` which is simply previous processed block's ObjectPath.

## Workaround

Removing indentation before #include file's line solves the problem.

ike709:
Closes #1050 

I did test it locally but I already had the file open on github so it was easier to webedit than push a commit.

### DIFF
--- a/DMCompiler/Compiler/DMPreprocessor/DMPreprocessor.cs
+++ b/DMCompiler/Compiler/DMPreprocessor/DMPreprocessor.cs
@@ -247,7 +247,7 @@ namespace DMCompiler.Compiler.DMPreprocessor {
 
         public void PreprocessFile(string includeDir, string file) {
             string filePath = Path.Combine(includeDir, file).Replace('\\', Path.DirectorySeparatorChar);
-            string source = File.ReadAllText(filePath);
+            string source = "\n" + File.ReadAllText(filePath); // Cope with whitespace before #includes by prepending a newline
             source = source.Replace("\r\n", "\n");
             source += '\n';
 


out-of-phaze:
Confirmed to fix the issue we had with this on Nebula.

ike709:
Now that I think about it, I need to test if this causes error line numbers to be off by one.

out-of-phaze:
> Now that I think about it, I need to test if this causes error line numbers to be off by one.

Debugger line numbers were off by one but I don't know if that's related or if it occurred before I locally merged this PR.

ike709:
This isn't the ideal fix and I cba to revisit this right now.

PR #1094, opened 2023-03-15
Fix `flist()` on nonexistent dirs

### LINKED ISSUE #1088: `flist()` on a nonexistant directory throws a runtime
BYOND instead gives you an empty list.

ike709:
Closes #1088 

At first I considered just checking if the dir exists first, but then I realized most flist() calls will presumably be for existing dirs, so that's added unnecessary overhead to every call whereas try/catch only adds overhead on calls for nonexistent dirs.

### DIFF
--- a/Content.Tests/DMProject/Tests/Builtins/flist.dm
+++ b/Content.Tests/DMProject/Tests/Builtins/flist.dm
@@ -3,5 +3,5 @@
 
 /proc/RunTest()
 	// Assert that a nonexistent dir returns an empty list
-	var/list/L = flist("./woiguowejsiojioeh") // If you create this dir, I'll kill you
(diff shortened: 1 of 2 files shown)


Content.Tests/DMProject/Tests/Builtins/flist.dm
```diff
@@ -0,0 +1,7 @@
+
+// TODO: Test that it lists files in a dir correctly. Right now we only test flist() on a nonexistent dir
+
+/proc/RunTest()
+	// Assert that a nonexistent dir returns an empty list
+	var/list/L = flist("./woiguowejsiojioeh") // If you create this dir, I'll kill you
```
ike709, line 6:
```suggestion
	var/list/L = flist("woiguowejsiojioeh/") // If you create this dir, I'll kill you
```

wixoaGit (approved):

PR #1099, opened 2023-03-16
`TryGetValueAsDreamObject` no longer returns deleted atoms

### LINKED ISSUE #1066: `MustGetValueAsDreamObject()` can return null if the object is deleted
Since `TryGetValueAsDreamObject()` calls `Must`, this causes `Try` to return `true` while setting the value of the out var to `null`.

Try should probably  return false when the object is deleted.

ike709:
Closes #1066


ike709:
I got codebaited by amy for a non-issue.

PR #1103, opened 2023-03-18
Increase the max stack depth to 400

ike709:
image


out-of-phaze:
More accurately, it's 400 for the `loop_checks = TRUE` warning. BYOND's actual hard stack limit is like... 3606-3609, according to Kapu.

wixoaGit:
> More accurately, it's 400 for the `loop_checks = TRUE` warning. BYOND's actual hard stack limit is like... 3606-3609, according to Kapu.

400 is likely a large enough cap, if we run into legitimate issues with it we can raise it again.

wixoaGit (changes requested):

wixoaGit (approved):

PR #1108, opened 2023-03-19
Fix incorrect Entered/Exited code

out-of-phaze:
This code called Entered and Exited on *all the contents of an entered turf,* which is incorrect and led to problems. It should only call Crossed/Uncrossed instead.


wixoaGit (changes requested):
There is also an incorrect loop here that should be fixed
```
// Ensure the atoms on the turf also permit this exit
for (var/atom/movable/exiting in loc)
	if (!exiting.Exit(src, NewLoc))
		return FALSE
```

This should be calling `Uncross()`, not `Exit()`

wixoaGit:
I'll merge this and open another PR for the above issue.

PR #1109, opened 2023-03-19
Add support for icon file overlays

out-of-phaze:
Adds support for adding overlays via `overlays += 'path/to/icon.dmi'`.


wixoaGit:
#1147 added this as part of assembling all the appearance-copying code into one method.

PR #1111, opened 2023-03-19
Fix null reference exception debugger crashes

out-of-phaze:
I encountered some crashes due to null DreamThreads in the list returned from InspectThreads and figured I'd PR a quick fix for it. CC @SpaceManiac to see if this is a sane solution or if it happened due to a configuration issue, user error, etc. in the first place, since I don't really know what I'm doing here.


wixoaGit:
This issue is fixed by [this change](https://github.com/OpenDreamProject/OpenDream/pull/1370/commits/5190585545693cff983cff6b1c3c87c20cdb3340#diff-b49e9b7620f3133848ff8f43df621cb135db79807888461565ea65babfe17a8aR65-R81) in #1370.

Still not an ideal fix as there shouldn't be nulls in the first place, but the fix is a lot closer to the source of the issue.

PR #1114, opened 2023-03-19
Add basic winshow implementation

out-of-phaze:
Winshow is effectively just a shorthand for using winset to set `is_visible=true`. This implements it as literally just that.


wixoaGit:
Just want to note that I expected `winshow()` in byond to at least verify that the target element is a window, but apparently not. It will also set the visibility of non-window controls.

wixoaGit (approved):

PR #1064, opened 2023-02-19
Implements all of the undocumented matrix constructors that targets use, adds compiletime linting of `/matrix()` calls

Altoids1:
Fixes #1055, but not completely, since I skipped over the undocumented matrix opcodes that our targets do not use.

## Summary
Amy recently pointed out that Paradise (and apparently also Goon) use this esoteric /matrix signature whose documentation only exists as an incomplete blog post written by Lummox 8 years ago.

Being as we target them, here is some support for that. Despite my laziness, I did end up converting some /matrix procs into natives to make it easier for these odd signatures to call into them.

While I was here, I also added linting against /matrix() calls within the compiler. We now correctly compiletime when /matrix() is called with too many arguments, as BYOND does. In addition, I've added a new pragma for detecting other bad /matrix calls (such as calling with only 5 arguments, or calling with 2-4 arguments w/o a matrix opcode). Another step in OD becoming a nice and articulate compiler to work with :^)

## Changelog
- `/matrix()` is now a native method, and supports several new undocumented signatures used by targets.
- `/matrix.Invert()` and `/matrix.Scale()` are now native methods.
- The compiler now lints against `/matrix()` calls, making sure they seem valid.
- Added the rest of the weird matrix defines into DMStandard.
- Some methods in DreamMetaObjectMatrix now avoid metaobject queries when getting and setting, probably improving performance a schmidge.

PR #1117, opened 2023-03-19
Rewrite `turn()`, `/matrix.Turn`, and `/matrix.Multiply` builtins in C# 

ZeWaka:
This rewrites three procs that were previously written in DM to be written in C#.

## Reasoning

This is mainly done for performance reasons, but also comes with many other benefits including more comprehensive debugging, better failure modes, access to actual types, and all the various goodies.

Builds off work done in #1064
See self-review for more annotations on specific code sections.
Dev testing and unit tests all passed.

## Performance comparisons (Release):
| Impl  | Operation (800,000 loops) | Time |
| ------------- | ------------- | ------------- |
| DM | Matrix.Multiply | 77ds |
| C# | Matrix.Multiply | 65ds |

**15.6%** increase in perf 
| Impl  | Operation (800,000 loops) | Time |
| ------------- | ------------- | ------------- |
| DM | Matrix.Turn | 80ds |
| C# | Matrix.Turn | 49ds |

**38.75%** increase in perf
| Impl  | Operation (10,000,000 loops) | Time |
| ------------- | ------------- | ------------- |
| DM | /proc/turn() | 101ds |
| C# | /proc/turn() | 44ds |

**56.43%** increase in perf

<details>
<summary>Test Code (borrowed from a unit test):</summary>
<pre>
/proc/one()
	var/matrix/M = matrix(1, 2, 3, 4, 5, 6)
	var/matrix/N = matrix(7, 8, 9, 10, 11, 12)
	M.Multiply(N)
	if(M ~! matrix(39, 54, 78, 54, 75, 108))
		CRASH("Unexpected matrix/Multiply result: [json_encode(M)]")
/proc/two()
	var/matrix/M = matrix(1, 2, 3, 4, 5, 6)
	M.Turn(90)
	if(M ~! matrix(4, 5, 6, -1, -2, -3))
		CRASH("Unexpected matrix/Turn result: [json_encode(M)]")
/proc/three()
	var/x = pick(NORTH, SOUTH, EAST, WEST, NORTHEAST, SOUTHEAST, SOUTHWEST, NORTHWEST)
	some_global = turn(x, pick(45, 1, 39, 90, -90, 20, 40, 10, -3))
</pre>
</details>

### DIFF
--- a/OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectIcon.cs
+++ b/OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectIcon.cs
@@ -72,4 +72,16 @@ sealed class DreamMetaObjectIcon : IDreamMetaObject {
         ObjectToDreamIcon.Add(icon, dreamIcon);
         return dreamIcon;
     }
+
+    public static DreamObject CloneIcon(IDreamObjectTree ObjectTree, DreamObject icon) {
+        // var newIcon = ObjectTree.CreateObject(icon.ObjectDefinition.TreeEntry);
+        // newIcon.InitSpawn(new Procs.DreamProcArguments(args));
+        //TODO: actually clone the icon
+        return icon;
+    }
+
+    public static DreamIcon TurnIcon(DreamIcon icon, float angle) {
+        //TODO: actually rotate the icon clockwise x degrees
+        return icon;
+    }
 }
--- a/OpenDreamRuntime/Procs/Native/DreamProcNative.cs
+++ b/OpenDreamRuntime/Procs/Native/DreamProcNative.cs
@@ -9,6 +9,7 @@ namespace OpenDreamRuntime.Procs.Native {
             DreamProcNativeRoot.MapManager = IoCManager.Resolve<IDreamMapManager>();
             DreamProcNativeRoot.ObjectTree = objectTree;
 
+            DreamProcNativeIcon.ObjectTree = objectTree;
             DreamProcNativeMatrix.ObjectTree = objectTree;
 
             objectTree.SetGlobalNativeProc(DreamProcNativeRoot.NativeProc_abs);
--- a/OpenDreamRuntime/Procs/Native/DreamProcNative.cs
+++ b/OpenDreamRuntime/Procs/Native/DreamProcNative.cs
@@ -116,6 +116,7 @@ namespace OpenDreamRuntime.Procs.Native {
             objectTree.SetGlobalNativeProc(DreamProcNativeRoot.NativeProc_time2text);
             objectTree.SetGlobalNativeProc(DreamProcNativeRoot.NativeProc_trimtext);
             objectTree.SetGlobalNativeProc(DreamProcNativeRoot.NativeProc_trunc);
+            objectTree.SetGlobalNativeProc(DreamProcNativeRoot.NativeProc_turn);
             objectTree.SetGlobalNativeProc(DreamProcNativeRoot.NativeProc_typesof);
             objectTree.SetGlobalNativeProc(DreamProcNativeRoot.NativeProc_uppertext);
             objectTree.SetGlobalNativeProc(DreamProcNativeRoot.NativeProc_url_decode);
--- a/OpenDreamRuntime/Procs/Native/DreamProcNative.cs
+++ b/OpenDreamRuntime/Procs/Native/DreamProcNative.cs
@@ -9,6 +9,8 @@ namespace OpenDreamRuntime.Procs.Native {
             DreamProcNativeRoot.MapManager = IoCManager.Resolve<IDreamMapManager>();
             DreamProcNativeRoot.ObjectTree = objectTree;
 
+            DreamProcNativeMatrix.ObjectTree = objectTree;
+
             objectTree.SetGlobalNativeProc(DreamProcNativeRoot.NativeProc_abs);
             objectTree.SetGlobalNativeProc(DreamProcNativeRoot.NativeProc_alert);
             objectTree.SetGlobalNativeProc(DreamProcNativeRoot.NativeProc_animate);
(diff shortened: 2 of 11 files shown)


OpenDreamRuntime/Procs/Native/DreamProcNativeMatrix.cs
```diff
@@ -0,0 +1,76 @@
...
+        DreamValue angleArg = arguments.GetArgument(0, "angle");
+        if (!angleArg.TryGetValueAsFloat(out float angle)) {
+            return new DreamValue(src); // Defaults to input on invalid angle
+        }
+        return _NativeProc_TurnInternal(src, usr, angle);
+    }
+
+    /// <summary> Turns a given matrix a given amount of degrees clockwise. </summary>
+    /// <returns> Returns a new matrix which has been rotated </returns>
+    public static DreamValue _NativeProc_TurnInternal(DreamObject src, DreamObject usr, float angle) {
```
out-of-phaze, line 64: This should also handle things like `turn(icon, 90)`.

ZeWaka: It *should*, yes. It doesn't currently.
I'll add a stub but I'm not adding that functionality in this PR - this is intended to just convert the existing builtin.

OpenDreamRuntime/Procs/Native/DreamProcNative.cs
```diff
@@ -9,6 +9,9 @@ static class DreamProcNative {
             DreamProcNativeRoot.MapManager = IoCManager.Resolve<IDreamMapManager>();
             DreamProcNativeRoot.ObjectTree = objectTree;
 
+            DreamProcNativeIcon.ObjectTree = objectTree;
+            DreamProcNativeMatrix.ObjectTree = objectTree;
```
ZeWaka, line 13: I think this is the right way to handle getting the ObjectTree into the native procs?

OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectIcon.cs
```diff
@@ -72,4 +72,16 @@ sealed class DreamMetaObjectIcon : IDreamMetaObject {
...
+        // var newIcon = ObjectTree.CreateObject(icon.ObjectDefinition.TreeEntry);
+        // newIcon.InitSpawn(new Procs.DreamProcArguments(args));
+        //TODO: actually clone the icon
+        return icon;
+    }
+
+    public static DreamIcon TurnIcon(DreamIcon icon, float angle) {
+        //TODO: actually rotate the icon clockwise x degrees
+        return icon;
+    }
```
ZeWaka, line 86: Out of scope, existing functionality doesn't implement these either. Stubbed for future completion.

OpenDreamRuntime/Procs/Native/DreamProcNativeIcon.cs
```diff
@@ -94,5 +97,22 @@ static class DreamProcNativeIcon {
...
+            DreamValue angleArg = arguments.GetArgument(0, "angle");
+            if (!angleArg.TryGetValueAsFloat(out float angle)) {
+                return new DreamValue(src); // Defaults to input on invalid angle
+            }
+            return _NativeProc_TurnInternal(src, usr, angle);
+        }
+
+        /// <summary> Turns a given icon a given amount of degrees clockwise. </summary>
+        /// <returns> Returns a new icon which has been rotated </returns>
+        public static DreamValue _NativeProc_TurnInternal(DreamObject src, DreamObject usr, float angle) {
```
ZeWaka, line 113: This is the best way I could think of resolving how there needs to be a `DreamProcArguments` passed from the DM Code-Interpreting-Runtime side, but we can just pass floats around in internal C# code. Without having to construct useless argument objects that get immediately dropped.

OpenDreamShared/Dream/AtomDirection.cs
```diff
@@ -2,7 +2,7 @@
 using System;
 
 namespace OpenDreamShared.Dream {
-    [Serializable, NetSerializable]
+    [Serializable, NetSerializable, Flags]
```
ZeWaka, line 5: Not sure why this wasn't applied.

OpenDreamRuntime/Procs/Native/DreamProcNative.cs
wixoaGit: Yep

Though in the future I want to avoid the static vars by storing them on the proc state (see #1065)

OpenDreamShared/Dream/AtomDirection.cs
wixoaGit: Wasn't aware of its existence when I first wrote this

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -2719,6 +2719,81 @@ public static DreamValue NativeProc_sorttext(DreamObject instance, DreamObject u
...
+
+            // If Dir is actually a matrix, call /matrix.Turn
+            if (dirArg.TryGetValueAsDreamObjectOfType(ObjectTree.Matrix, out var matrix)) {
+                // Clone matrix here since it's specified to return a new one
+                DreamObject clonedMatrix = DreamMetaObjectMatrix.MatrixClone(ObjectTree, matrix);
+                return DreamProcNativeMatrix._NativeProc_TurnInternal(clonedMatrix, usr, angle);
+            }
+
+            if (!dirArg.TryGetValueAsInteger(out int possibleDir)) {
+                throw new ArgumentException($"Invalid Dir for Turn: \"{dirArg.ToString()}\"");
```
wixoaGit, line 2751: A non-number `Dir` defaults to 0 in BYOND.

OpenDreamRuntime/Procs/Native/DreamProcNativeMatrix.cs
```diff
@@ -11,6 +14,28 @@ internal static class DreamProcNativeMatrix {
         return new DreamValue(src);
     }
 
+    [DreamProc("Multiply")]
+    [DreamProcParameter("Matrix2")] // or "n"
```
wixoaGit, line 18:
```suggestion
    [DreamProcParameter("Matrix2", Type = DreamValueType.DreamObject | DreamValueType.Float)] // or "n"
```

wixoaGit (changes requested):

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
ZeWaka: Fixed.

wixoaGit (approved):

PR #1121, opened 2023-03-21
Misc DMF fixes

ike709:
See commit names.

### DIFF
--- a/OpenDreamClient/Interface/DMF/DMFParser.cs
+++ b/OpenDreamClient/Interface/DMF/DMFParser.cs
@@ -203,12 +203,19 @@ public sealed class DMFParser : Parser<char> {
             }
 
             Token attributeValue = Current();
-            if (!Check(TokenType.DMF_Value))
-                Error($"Invalid attribute value ({attributeValue.Text})");
+            if (!Check(TokenType.DMF_Value)) {
+                if (Check(TokenType.DMF_Semicolon)) {
+                    token = "none";
+                } else {
+                    Error($"Invalid attribute value ({attributeValue.Text})");
+                }
+            } else {
+                token = attributeValue.Text;
+            }
 
             Newline();
             key = attributeToken.Text;
-            token = attributeValue.Text;
+
             return true;
         }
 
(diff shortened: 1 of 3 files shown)


OpenDreamClient/Interface/DMF/DMFParser.cs
```diff
@@ -203,12 +203,19 @@ public sealed class DMFParser : Parser<char> {
             }
 
             Token attributeValue = Current();
-            if (!Check(TokenType.DMF_Value))
-                Error($"Invalid attribute value ({attributeValue.Text})");
+            if (!Check(TokenType.DMF_Value)) {
```
ike709, line 206: I'm not sure whether the changes to this file work as intended.

OpenDreamClient/Interface/InterfaceMacro.cs
```diff
@@ -38,8 +38,7 @@ public sealed class InterfaceMacroSet : InterfaceElement {
     public override void AddChild(ElementDescriptor descriptor) {
         if (descriptor is not MacroDescriptor macroDescriptor)
             throw new ArgumentException($"Attempted to add a {descriptor} to a macro set", nameof(descriptor));
-
-        Macros.Add(macroDescriptor.Name, new InterfaceMacro(_inputContextName, macroDescriptor, _entitySystemManager, _inputManager, _inputContext));
+        Macros[macroDescriptor.Name] = new InterfaceMacro(_inputContextName, macroDescriptor, _entitySystemManager, _inputManager, _inputContext);
```
ike709, line 41: Previously we'd get an error that the element is already present in the dictionary. I'm not entirely sure whether this fix matches BYOND behavior.

OpenDreamClient/Interface/DMF/DMFParser.cs
wixoaGit: A value of `none` leads to the attribute being ignored entirely. This would work if it was the initial value in a DMF, though it'll get ignored in a winset. I'm actually not sure how this would be fixed given how we use ISerializationManager.

Should also note a newline does the same thing as a semicolon here.

wixoaGit (changes requested):

wixoaGit:
`command=.me` syntax was fixed in #1189

ike709:
Closing since this is old and stale.

PR #1122, opened 2023-03-21
Implement json encoding a proc

ike709:
BYOND just turns it into a string.

image


OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -1251,6 +1251,8 @@ static class DreamProcNativeRoot {
                 writer.WriteStringValue(text);
             else if (value.TryGetValueAsType(out var type))
                 writer.WriteStringValue(HttpUtility.JavaScriptStringEncode(type.Path.PathString));
+            else if (value.TryGetValueAsProc(out var proc))
+                writer.WriteStringValue(HttpUtility.JavaScriptStringEncode(proc.ToString()));
```
wixoaGit, line 1255: WriteStringValue already escapes strings
```suggestion
                writer.WriteStringValue(proc.ToString());
```

The other two uses of `JavaScriptStringEncode()` should also be removed.

wixoaGit (changes requested):

wixoaGit (approved):

PR #1123, opened 2023-03-21
Speed up `/atom` and `/atom/movable` init times

wixoaGit:
Speed up `/atom/proc/<init>` and `/atom/movable/proc/<init>` by setting the built-in list vars inside their meta objects.

### DIFF
--- a/OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectAtom.cs
+++ b/OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectAtom.cs
@@ -31,6 +31,12 @@ namespace OpenDreamRuntime.Objects.MetaObjects {
             VerbLists[dreamObject] = new VerbsList(_objectTree, dreamObject);
             _filterLists[dreamObject] = new DreamFilterList(_objectTree.List.ObjectDefinition, dreamObject);
 
+            // TODO: These should use their own special list types
+            dreamObject.SetVariable("overlays", new(_objectTree.CreateList()));
+            dreamObject.SetVariable("underlays", new(_objectTree.CreateList()));
+            dreamObject.SetVariableValue("vis_locs", new(_objectTree.CreateList()));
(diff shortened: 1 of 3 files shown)


ike709:
Might want to drop a comment mentioning that the lists are initialized internally.

wixoaGit:
> Might want to drop a comment mentioning that the lists are initialized internally.

Will do once I'm at my computer again tomorrow.

OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectAtom.cs
```diff
@@ -31,6 +31,12 @@ sealed class DreamMetaObjectAtom : IDreamMetaObject {
             VerbLists[dreamObject] = new VerbsList(_objectTree, dreamObject);
             _filterLists[dreamObject] = new DreamFilterList(_objectTree.List.ObjectDefinition, dreamObject);
 
+            // TODO: These should use their own special list types
+            dreamObject.SetVariable("overlays", new(_objectTree.CreateList()));
+            dreamObject.SetVariable("underlays", new(_objectTree.CreateList()));
+            dreamObject.SetVariableValue("vis_locs", new(_objectTree.CreateList()));
+            dreamObject.SetVariableValue("vis_contents", new(_objectTree.CreateList()));
```
ike709, line 38: But in the DM file these two still do 
```
var/list/vis_locs = list() as opendream_unimplemented
var/list/vis_contents = list() as opendream_unimplemented
```
?

I also don't see where you internally create a list for `locs` or `content`

wixoaGit: Oops, must have missed that while fixing a merge conflict

`locs` is already created on each access inside DreamMetaObjectMovable

`contents` is different depending on the type of atom, they all have their own getter

PR #1128, opened 2023-03-27
Fix inferred paths in appends and lint for suspicious `list += new`

### LINKED ISSUE #1126: Lint for `list += new`
```dm
var/list/mob/mobs = list()
mobs += new
```

This actually translates to `mobs += list()`, because typed lists aren't real.

ike709:
Closes #1126 

The inferred path behavior now matches `VisitAssign()`.


wixoaGit (changes requested):

wixoaGit:
This appears long abandoned, and I think a different method of detecting this issue would be better.

PR #1131, opened 2023-03-30
Improves error emission of switch blocks, adds new pragma

### LINKED ISSUE #996: `else if()` Switch Case Detection Causes Code Skips
Code:
image

Error:
```
Error OD0000 at code\datums\components\construction.dm:94:12: Expected "if" or "else", "else if" is not permitted as a switch case
Warning OD0000 at code\datums\components\construction.dm:99:5: Error recovery had to skip to code\datums\components\construction.dm:114:7
```

It detects the bad switch case like it should, but it should recover without a code skip.

Altoids1:
Fixes #996.

## Summary
Quick and easy audit of our switch parsing, to make use of the new(ish) pragma system.

## Changelog
- Errors in switch parsing are now less likely to cause large code skips in the parser
- New style pragma added: `SuspiciousSwitchCase`, for when user codebases use `else if` as a switch case (which is valid but is read as an 'else' case block with an if block inside it)
- DMParser calling `base.Error()` is now marked as obsolete. All calls of that nature should be rewritten to make use of the pragma system and not necessitate throwing, whenever possible (to minimize code skips).


DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -1364,7 +1364,7 @@ public partial class DMParser : Parser<Token> {
                     DMASTExpression? expression = Expression();
                     if (expression == null) {
                         if (expressions.Count == 0) {
-                            Error("Expected an expression");
+                            Error(WarningCode.BadExpression, "Expected an expression");
                         } else {
                             //Eat a trailing comma if there's at least 1 expression
                             break;
```
wixoaGit, line 1370: This needs to break out of the loop, now that the `Error()` call isn't throwing.
```suggestion
                            Error(WarningCode.BadExpression, "Expected an expression");
                        }
                        
                        //Eat a trailing comma
                        break;
```

DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -1375,7 +1375,7 @@ public partial class DMParser : Parser<Token> {
                         var loc = Current().Location;
                         Whitespace();
                         DMASTExpression? rangeEnd = Expression();
-                        if (rangeEnd == null) Error("Expected an upper limit");
+                        if (rangeEnd == null) Error(WarningCode.BadExpression, "Expected an upper limit");
```
wixoaGit, line 1378: This needs to have a fallback for `rangeEnd` now that this isn't throwing.

wixoaGit (changes requested):

wixoaGit (approved):

PR #1133, opened 2023-04-01
Embedded `winset`

### LINKED ISSUE #1125: "Embedded winsets" are not supported
In a browser control, when a link like `byond://winset?id=elementId&attribute=value` is followed, it should perform a winset instead of a topic call.

This is used in tgpanel to make itself visible on modern tg.

ike709:
Fixes #1125 

TG's usage won't work until https://github.com/OpenDreamProject/OpenDream/pull/1121 is also merged.

Tested it with the following tossed in a verb (requires the other aforementioned PR to work):
`usr << browse("<a href=\"byond://winset?size=100x10&command=.quit\">Quit</a>", "window=quitbutton")`

### DIFF
--- a/OpenDreamClient/Interface/Controls/ControlBrowser.cs
+++ b/OpenDreamClient/Interface/Controls/ControlBrowser.cs
@@ -80,7 +80,19 @@ namespace OpenDreamClient.Interface.Controls
                 context.DoCancel();
 
                 if (newUri.Host == "winset") { // Embedded winset. Ex: usr << browse("<a href=\"byond://winset?command=.quit\">Quit</a>", "window=quitbutton")
-                    _interfaceManager.WinSet(null, newUri.Query.Substring(1)); // Strip the question mark
+                    // Strip the question mark out before parsing
+                    var queryParams = HttpUtility.ParseQueryString(newUri.Query.Substring(1));
+
+                    // We need to extract the control element (if one was included)
+                    string? element = queryParams.Get("element");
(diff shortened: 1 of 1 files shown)


OpenDreamClient/Interface/Controls/ControlBrowser.cs
```diff
@@ -78,6 +79,23 @@ public ControlBrowser(ControlDescriptor controlDescriptor, ControlWindow window)
...
+                    // Strip the question mark out before parsing
+                    var queryParams = HttpUtility.ParseQueryString(newUri.Query.Substring(1));
+
+                    // We need to extract the control element (if one was included)
+                    string? element = queryParams.Get("element");
+                    queryParams.Remove("element");
+
+                    // Reassemble the query params without element then convert to winset syntax
+                    var query = queryParams.ToString();
+                    query = query!.Replace('&', ';');
```
wixoaGit, line 92: Could this cause issues with values that have `&` in them? Potentially something like `menu.name="&File"`

wixoaGit (changes requested):

OpenDreamClient/Interface/Controls/ControlBrowser.cs
ike709:
```suggestion
                    query = query!.Replace('&', ';'); // TODO: More robust parsing
```

wixoaGit (approved):

PR #1135, opened 2023-04-05
Implements `nameof()`

### LINKED ISSUE #1115: nameof() is not supported but call()() uses post-515 behavior
Prior to BYOND 515, passing a proc path to `call()()` would resolve it to the deepest override of that proc on the object it was being called on. This reportedly happened because BYOND would silently convert the proc path argument into the name of the leaf node of the path, making `call(new /mob, /atom/movable/proc/do_thing)()` equivalent to `call(new /mob, "do_thing")()` and calling a `/mob/do_thing()` override if one exists. After BYOND 515, passing a proc path will call that exact proc, i.e. `call(new /mob, /atom/movable/proc/do_thing)()` will only ever call the `/atom/movable` definition of that proc, even if a mob-level override exists.

The solution for this breaking change in BYOND 515 was the introduction of the `nameof` compile-time helper, which allows you to explicitly turn a path or variable into the name of its leaf node *at compile-time.* The pre-515 `call(new /mob, /atom/movable/proc/do_thing)()` functionality could then be replicated through `call(new /mob, nameof(/atom/movable/proc/do_thing))()`. This is preferable to just replacing it with `call(new /mob, "do_thing")()` because the former will emit a compiler error if `/atom/movable/proc/do_thing` is renamed or removed, while the latter will not, allowing for better error detection.

This is an issue in some targets, specifically Nebula, which is currently targeting BYOND 514. `INVOKE_ASYNC(A, /atom/proc/Bumped, src)` effectively expands to `call(A, /atom/proc/Bumped)(src)`, which is intended to call the latest override of `/atom/proc/Bumped` on `A`: for example, the airlock Bumped proc that causes airlocks to open when someone collides with them. However, in OpenDream and BYOND 515, they only call the no-op stub proc `/atom/proc/Bumped`, meaning any on-collide functionality fails to run.

A minimal test case for the 514 functionality:
```
/datum/proc/call_test()
    CRASH("This should not be called.")

var/global/was_called = FALSE
/atom/call_test()
    global.was_called = TRUE
    
/world/New()
    call(new /obj, /datum/proc/call_test)()
    ASSERT(global.was_called == TRUE)
```

ike709:
Closes #1115 

Let me know if you think of other expressions that I missed, but the ref isn't super detailed.

### DIFF
--- a/DMCompiler/DM/DMExpression.cs
+++ b/DMCompiler/DM/DMExpression.cs
@@ -69,7 +69,7 @@ namespace DMCompiler.DM {
             throw new CompileAbortException(Location, $"Cannot reference r-value");
         }
 
-        public virtual string EmitNameof(DMObject dmObject, DMProc proc) {
+        public virtual string GetNameof(DMObject dmObject, DMProc proc) {
             throw new CompileAbortException(Location, "nameof: requires a var, proc reference, or type path");
         }
 
--- a/DMCompiler/DM/DMExpression.cs
+++ b/DMCompiler/DM/DMExpression.cs
@@ -69,6 +69,10 @@ namespace DMCompiler.DM {
             throw new CompileAbortException(Location, $"Cannot reference r-value");
         }
 
+        public virtual string EmitNameof(DMObject dmObject, DMProc proc) {
+            throw new CompileAbortException(Location, "nameof: requires a var, proc reference, or type path");
+        }
+
         public virtual DreamPath? Path => null;
     }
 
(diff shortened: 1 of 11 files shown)


DMCompiler/DM/DMExpression.cs
```diff
@@ -69,6 +69,10 @@ public enum IdentifierPushResult {
             throw new CompileAbortException(Location, $"Cannot reference r-value");
         }
 
+        public virtual string EmitNameof(DMObject dmObject, DMProc proc) {
```
wixoaGit, line 72: `EmitNameof()` doesn't seem like a proper name since it's not actually emitting anything.

wixoaGit (changes requested):

DMCompiler/DM/DMExpression.cs
ike709: Suggestions?

wixoaGit (approved):

PR #1139, opened 2023-04-07
Enable nullable on the compiler, client, and shared

### LINKED ISSUE #658: Enable nullables in DMCompiler
It'd be convenient and also it'd be consistent with the rest of the project.

wixoaGit:
Closes #658


pali6:
Notably this also fixes several uncaught exceptions that currently happen when one attempts to clone a window.

PR #1145, opened 2023-04-12
Deep copy `atom.overlays` and `atom.underlays`

### LINKED ISSUE #1089: Assigning a list to `/atom.overlays` and `/atom.underlays` should copy it
Right now assigning them to a list creates a reference instead of a copy. This can have undesirable side effects such as the list being cleared when `overlays`/`underlays` is assigned again.

ike709:
Fixes #1089 

Is it really just that easy?


wixoaGit:
Closing in favor of #1212

PR #1158, opened 2023-04-16
Add an enter shortcut to input prompts

wixoaGit:
You can now press enter to finish input prompts.

Bothered me during the playtest.


wixoaGit:
You need to press enter twice for some reason

wixoaGit:
> You need to press enter twice for some reason

This is a bug coming from RobustToolbox/its windowing API. When the prompt window is closed that prevents it from getting the key-up event and releasing the `TextSubmit` keybind.

wixoaGit:
> This is a bug coming from RobustToolbox/its windowing API. When the prompt window is closed that prevents it from getting the key-up event and releasing the `TextSubmit` keybind.

I made a [RobustToolbox issue](https://github.com/space-wizards/RobustToolbox/issues/3971) for this. It's not an issue with this PR so I'll be merging this.

PR #1159, opened 2023-04-17
Add profiler groups to dream overlay

ElectroJr:
Makes it easier to get an overview of which parts of the rendering is slow. Also changes the lookup flags to use approximate lookups, seeing as lookup area is already arbitrarily scaled up.

OpenDreamClient_oMR5riM9mM

### DIFF
--- a/OpenDreamClient/Rendering/DreamViewOverlay.cs
+++ b/OpenDreamClient/Rendering/DreamViewOverlay.cs
@@ -7,6 +7,7 @@ using OpenDreamShared.Dream;
 using Robust.Shared.Console;
 using Robust.Shared.Prototypes;
 using OpenDreamShared.Rendering;
+using Robust.Shared.Profiling;
 
 namespace OpenDreamClient.Rendering;
 
@@ -19,6 +20,7 @@ sealed class DreamViewOverlay : Overlay {
     [Dependency] private readonly IEntityManager _entityManager = default!;
     [Dependency] private readonly IMapManager _mapManager = default!;
     [Dependency] private readonly IClyde _clyde = default!;
+    [Dependency] private readonly ProfManager _prof = default!;
     private EntityQuery<DMISpriteComponent> spriteQuery;
     private EntityQuery<TransformComponent> xformQuery;
     private EntityQuery<DreamMobSightComponent> mobSightQuery;
@@ -45,6 +47,7 @@ sealed class DreamViewOverlay : Overlay {
     private Stack<RendererMetaData> _rendererMetaDataRental = new();
     private Stack<RendererMetaData> _rendererMetaDataToReturn = new();
     private Matrix3 _flipMatrix;
+    private const LookupFlags Flags = LookupFlags.Approximate | LookupFlags.Dynamic | LookupFlags.Static | LookupFlags.Sundries | LookupFlags.Contained;
 
     public DreamViewOverlay() {
         IoCManager.InjectDependencies(this);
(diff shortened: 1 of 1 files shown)


OpenDreamClient/Rendering/DreamViewOverlay.cs
```diff
@@ -45,6 +47,7 @@ sealed class DreamViewOverlay : Overlay {
     private Stack<RendererMetaData> _rendererMetaDataRental = new();
     private Stack<RendererMetaData> _rendererMetaDataToReturn = new();
     private Matrix3 _flipMatrix;
+    private const LookupFlags Flags = LookupFlags.Approximate | LookupFlags.Dynamic | LookupFlags.Static | LookupFlags.Sundries | LookupFlags.Contained;
```
ElectroJr, line 50: I don't know if OD uses the engine's containers. If it doesn't (or only uses it for non-visible entities), then the `LookupFlags.Contained` should be removed, which should speed things up a bit.

OpenDreamClient/Rendering/DreamViewOverlay.cs
```diff
@@ -120,7 +125,13 @@ sealed class DreamViewOverlay : Overlay {
             return;
         Box2 screenArea = Box2.CenteredAround(eyeTransform.WorldPosition, args.WorldAABB.Size);
 
-        var entities = _lookupSystem.GetEntitiesIntersecting(args.MapId, screenArea.Scale(1.2f)); //the scaling is to attempt to prevent pop-in, by rendering sprites that are *just* offscreen
+        HashSet<EntityUid> entities;
+        using (_prof.Group("lookup")) {
+            //TODO use a sprite tree.
```
ElectroJr, line 130: I assume rendering performance isn't a big concern atm, but eventually this should use a proper `DMISpriteComponent` lookup tree instead of doing a general entity lookup. I just did a basic test, and of all the entities returned only ~ 60% actually had a `DMISpriteComponent`.

OpenDreamClient/Rendering/DreamViewOverlay.cs
wixoaGit: We don't. Would be a good idea to use them, though.

ElectroJr: I'll just remove the flag then.

wixoaGit:
Useful to know this exists!

Took me a while to figure out I could open up "UI" here though, since it's drawing a white arrow on white
image

OpenDreamClient/Rendering/DreamViewOverlay.cs
wixoaGit: It used to be that every entity had a `DMISpriteComponent`, but now that the client creates entities for opaque turfs that would be a good idea.

OpenDreamClient/Rendering/DreamViewOverlay.cs
```diff
@@ -45,6 +47,7 @@ sealed class DreamViewOverlay : Overlay {
     private Stack<RendererMetaData> _rendererMetaDataRental = new();
     private Stack<RendererMetaData> _rendererMetaDataToReturn = new();
     private Matrix3 _flipMatrix;
+    private const LookupFlags Flags = LookupFlags.Approximate | LookupFlags.Uncontained;
```
wixoaGit, line 50: The name seems a little vague, maybe `MapLookupFlags` would be better?

wixoaGit (changes requested):

ElectroJr (commented):
> Useful to know this exists!

PJB did good work with the profiler, its been really useful for figuring out what's causing frame stutters or lag. Unfortunately a lot of the colour scheme for the profiler was hard coded assuming a ss14 style dark background.
Content Client_LhDjmi6hr5

E.g., `ExpanderArrow` is just always white and `ProfTree` configures the alternating background colours in `ProfAltBackground`, but they are also hardcoded. It should probably be changed so that it can be customized via style sheets, but I cbf doing that at the moment.

wixoaGit (approved):

PR #1161, opened 2023-04-18
Fully Implement Num2Text

redmoogle:
Who shoves multiple behaviors (that are based on argument counts) into a proc

Should match up with this https://www.byond.com/docs/ref/#/proc/num2text


OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -1565,16 +1565,35 @@ internal static class DreamProcNativeRoot {
...
+            if(bundle.Arguments.Length == 1) {
+                return new DreamValue(floatNum.ToString("g6"));
+            }
+
+            if(bundle.Arguments.Length == 2) {
+                var sigFig = bundle.GetArgument(1, "A");
+                if(sigFig == DreamValue.Null) {
+                    return new DreamValue(floatNum.ToString("g6"));
+                }
+                return new DreamValue(floatNum.ToString('g' + sigFig.ToString()));
```
wixoaGit, line 1585: This is not what `DreamValue.ToString()` is for, it should only be used to display a value to the user. This could very easily break at another time and does not work for non-integer values.
```suggestion
                if(!bundle.GetArgument(1, "A").TryGetValueAsInteger(out var sigFig)) {
                    return new DreamValue(floatNum.ToString("g6"));
                }
                return new DreamValue(floatNum.ToString($"g{sigFig}"));
```

OpenDreamRuntime/Procs/Native/DreamProcNativeHelpers.cs
```diff
@@ -297,6 +305,32 @@ internal static partial class DreamProcNativeHelpers {
...
+        StringBuilder resString = new();
+        bool wasNegative = false; // Theres likely a better way
+
+        if(value < 0) {
+            wasNegative = true;
+            value = Math.Abs(value);
+        }
+
+        while (value > 0)
+        {
```
wixoaGit, line 322:
```suggestion
        while (value > 0) {
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #1166, opened 2023-04-18
Add line numbers to DM stack traces

wixoaGit:


wixoaGit:
The debugger uses this method so I want to make sure this doesn't break it before merging.

PR #1168, opened 2023-04-19
Fix `rand()` not being inclusive on the high end

wixoaGit:


OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -1874,12 +1874,12 @@ public static DreamValue NativeProc_oviewers(DreamObject instance, DreamObject u
             } else if (arguments.ArgumentCount == 1) {
                 arguments.GetArgument(0, "L").TryGetValueAsInteger(out var high);
 
-                return new DreamValue(DreamManager.Random.Next(high));
+                return new DreamValue(DreamManager.Random.Next(high)+1);
```
ike709, line 1877:
```suggestion
                return new DreamValue(DreamManager.Random.Next(high)+1); // rand() is inclusive on both ends
```

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -1874,12 +1874,12 @@ public static DreamValue NativeProc_oviewers(DreamObject instance, DreamObject u
...
             } else {
                 arguments.GetArgument(0, "L").TryGetValueAsInteger(out var low);
                 arguments.GetArgument(1, "H").TryGetValueAsInteger(out var high);
 
-                return new DreamValue(DreamManager.Random.Next(Math.Min(low, high), Math.Max(low, high)));
+                return new DreamValue(DreamManager.Random.Next(Math.Min(low, high), Math.Max(low, high)+1));
```
ike709, line 1882:
```suggestion
                return new DreamValue(DreamManager.Random.Next(Math.Min(low, high), Math.Max(low, high)+1)); // rand() is inclusive on both ends
```

PR #1171, opened 2023-04-19
Fix `turn()` with an invalid first arg

wixoaGit:
It was attempting to cast an int to a DreamValue directly. I also fixed a bug where it could return 0.

This was triggered by this code in Paradise, which is probably another bug in itself that needs looked into (`dir` was 0):
```
/obj/structure/disposalpipe/segment/Initialize(mapload)
	. = ..()
	if(icon_state == "pipe-s")
		dpdir = dir | turn(dir, 180)
	else
		dpdir = dir | turn(dir, -90)
	update()
```

### DIFF
--- a/OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
+++ b/OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
@@ -4,6 +4,7 @@ using OpenDreamRuntime.Resources;
 using OpenDreamShared.Dream;
 using Robust.Shared.Utility;
 using System.Collections.Specialized;
+using System.Diagnostics;
 using System.Globalization;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -2790,9 +2778,34 @@ public static DreamValue NativeProc_sorttext(DreamObject instance, DreamObject u
...
 
+            // Is the dir invalid?
+            if (dirAngle == null) {
+                // If Dir is invalid and angle is zero, 0 is returned
+                if (angle == 0) {
+                    return new DreamValue(0);
+                }
+
+                // Otherwise, it returns a random direction
+                // Can't just select a random value from AtomDirection since that contains AtomDirection.None
```
ZeWaka, line 2792: Ah, my bad. Thanks for the fix.

PR #1177, opened 2023-04-21
Implement verb invisibility

ike709:
Extremely basic testing shows that it works.


OpenDreamRuntime/DreamConnection.cs
```diff
@@ -127,6 +130,15 @@ public sealed class DreamConnection {
                         continue;
                     }
 
+                    // Don't send invisible verbs.
+                    sbyte seeVis = 127;
+                    if(_entityManager.TryGetComponent<DreamMobSightComponent>(Session?.AttachedEntity, out var mobSight)){
+                        seeVis = mobSight.SeeInvisibility;
+                    }
```
wixoaGit, line 137: Grabbing the component for every verb is slow. Move this outside the `AddVerbs()` method.

wixoaGit (changes requested):

OpenDreamRuntime/DreamConnection.cs
wixoaGit: With #1266 merged it'd be just as good to use `_mob.SeeInvisible` instead.

Does `/client.eye` affect verb invisibility? I figure not since `see_invisible` is a /mob var.

wixoaGit (approved):

PR #1179, opened 2023-04-21
minor improvements for fract and trunc

redmoogle:
cleaner code also probably more performant

### DIFF
--- a/OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
+++ b/OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
@@ -2724,7 +2724,7 @@ namespace OpenDreamRuntime.Procs.Native {
         public static DreamValue NativeProc_trunc(NativeProc.State state) {
             DreamValue arg = state.GetArgument(0, "n");
             if (arg.TryGetValueAsFloat(out var floatNum)) {
-                if (float.IsInfinity(floatNum) || float.IsNaN(floatNum)) {
+                if (!float.IsFinite(floatNum)) {
                     return arg;
                 }
                 return new DreamValue((int)floatNum);
--- a/OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
+++ b/OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
@@ -709,11 +709,14 @@ namespace OpenDreamRuntime.Procs.Native {
         [DreamProcParameter("n", Type = DreamValueType.Float)]
         public static DreamValue NativeProc_fract(NativeProc.State state) {
             DreamValue arg = state.GetArgument(0, "n");
-            if (arg.TryGetValueAsFloat(out float floatnum)) {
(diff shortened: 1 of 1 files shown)


OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -709,11 +709,14 @@ static class DreamProcNativeRoot {
...
-                if(float.IsInfinity(floatnum)) {
+            if (arg.TryGetValueAsFloat(out var floatNum)) {
+                if(float.IsInfinity(floatNum)) {
                     return new DreamValue(0);
                 }
-                return new DreamValue(floatnum - MathF.Truncate(floatnum));
+                if(float.IsNaN(floatNum)) {
+                    return arg;
+                }
+                return new DreamValue(floatNum - (int)floatNum);
```
wixoaGit, line 719: Why should `(int)floatNum` be preferred over `MathF.Truncate(floatNum)`?

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -2720,8 +2723,11 @@ public static DreamValue NativeProc_sorttext(NativeProc.State state)
         [DreamProcParameter("n", Type = DreamValueType.Float)]
         public static DreamValue NativeProc_trunc(NativeProc.State state) {
             DreamValue arg = state.GetArgument(0, "n");
-            if (arg.TryGetValueAsFloat(out float floatnum)) {
-                return new DreamValue(MathF.Truncate(floatnum));
+            if (arg.TryGetValueAsFloat(out var floatNum)) {
+                if (!float.IsFinite(floatNum)) {
+                    return arg;
+                }
+                return new DreamValue((int)floatNum);
```
wixoaGit, line 2730: Does this not just give the same values as `MathF.Truncate()`?

wixoaGit (changes requested):

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
redmoogle: they do roughly the same things do they not; subtraction shouldnt be a particularity expensive op. Although im not 100% sure on intcasting

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
redmoogle: Yes, but imo in this case its cleaner since its just turning it into a integer

wixoaGit: I don't see how this is cleaner and preferable over a one-line call to the `MathF` equivalent of the proc.

wixoaGit:
This doesn't affect the behavior or performance of these procs, nor is it better looking code in my opinion.

PR #1182, opened 2023-04-21
mob.ckey and client.ckey now conform to how byond generates them from key

pali6:
Specifically the `ckey = ckey(key)` invariant should now hold.

This might eventually lead to trouble if SS14 allows usernames which differ only in case / special characters. But that will likely need to be solved on the side of ss13 codebases which will need to replace uses of `ckey` with `key` or something? Idk.


OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectMob.cs
```diff
@@ -39,6 +39,7 @@ sealed class DreamMetaObjectMob : IDreamMetaObject {
             ParentType?.OnVariableSet(dreamObject, varName, value, oldValue);
 
             if (varName == "key" || varName == "ckey") {
+                // TODO: make this work with ckeys too
```
pali6, line 42: I can try to implement this in this PR but if that's wanted I'd like some guidance on what's a good place to store and preserve the ckey to key dictionary. My first instinct was to subclass the player manager but it's sealed and also idk how it all exactly works.

out-of-phaze:
For some reason, with this PR merged spawning in as a quantum mechanic (admin mob) no longer gives you control of it on Nebula, when it did before. Probably because it uses `Q.ckey = ckey`?

pali6:
Sounds likely. I'll wait for some advice for where to add the dictionary I mentioned and then implement that.

OpenDreamRuntime/Objects/MetaObjects/DreamMetaObjectMob.cs
wixoaGit: It may be better to just loop through DreamManager.Connections and ckey() each username to compare with. I don't think performance would be a huge concern here.

OpenDreamRuntime/Procs/Native/DreamProcNativeHelpers.cs
```diff
@@ -237,4 +238,17 @@ internal static class DreamProcNativeHelpers {
             return false;
         }
     }
+
+    /// <summary>
+    /// Returns the string with all non-alphanumeric characters (except @) removed, and all letters converted to lowercase.
+    /// Mirrors the behaviour of BYOND's ckey() proc.
+    /// </summary>
+    /// <param name="input"></param>
```
wixoaGit, line 246:
```suggestion
    /// <param name="input">The string to canonicalize</param>
```

wixoaGit (changes requested):
Some tests for `ckey()` would be nice

wixoaGit:
Should note that this also fixes the `who` command on Paradise for anybody who's username doesn't already fit the ckey format.

wixoaGit (approved):

PR #1184, opened 2023-04-23
Overlay/underlay tiebreaker, `FLOAT_LAYER`, & Image over/underlays fix

amylizzle:
Basically replaces the previous attempt with a tree traversal algorithm that should preserve proper underlay and overlay render order even when there are multiply nested under/overlays. Also adds another visual test to ensure that they're working right. 
Also implements `FLOAT_LAYER` and makes `/image` work right when used as overlays.
image


out-of-phaze:
With this PR merged, instead of the ordering on human mobs on Nebula being broken some of the time, it's broken all of the time. :(

out-of-phaze:
Tested again—after removing a bad Overlays/Underlays.Sort() call outside of renderer code, it's... exactly as broken with or without this PR. Darn.

amylizzle:
Works good now, if there's still a problem I think it must be something other than the ordering.

amylizzle:
also note: this is copy pasted code from atom overlays. It's gross I know, but it should be a one-for-one swap-out when someone does overlay lists properly.

DMCompiler/DMStandard/Types/Image.dm
```diff
@@ -1,17 +1,17 @@
 ﻿/image
 	parent_type = /datum
 
-	var/alpha
+	var/alpha = 255
```
wixoaGit, line 4: The defaults for `/image` don't come from here, they're whatever the default value in `IconAppearance` is.

I suppose setting it here is desirable for `initial()` though. Maybe leave a comment somewhere clarifying where the actual defaults come from?

amylizzle: So I should set these values in `IconAppearance` and `/image`?

amylizzle: These values are all already set in IconAppearance, so I just added the comment.

ike709:
CI fails.

amylizzle:
Finally got around to testing this - I can't replicate missing player sprites
image

out-of-phaze:
> Finally got around to testing this - I can't replicate missing player sprites image

Yeah, player sprites work fine on Nebula now, I think? The issue was some weirdness with something like a blank, no-args image inheriting icon from its overlay, so it was overlaying a duplicate torso on top of the correctly-generated player sprite.

OpenDreamClient/Rendering/DreamViewOverlay.cs
```diff
@@ -481,28 +488,30 @@ internal sealed class DreamViewOverlay : Overlay {
...
-                current.KeepTogetherGroup.AddRange(ProcessIconComponents(underlay, current.Position, uid, isScreen, current, keepTogether, underlayTiebreaker));
+                current.KeepTogetherGroup.AddRange(ProcessIconComponents(underlay, current.Position, uid, isScreen, ref tieBreaker, current, keepTogether));
             }
         }
 
+        tieBreaker++;
+        current.TieBreaker = tieBreaker;
+
+
         //overlays - colour, alpha, and transform are inherited, but filters aren't
```
wixoaGit, line 506:
```suggestion
        tieBreaker++;
        current.TieBreaker = tieBreaker;

        //overlays - colour, alpha, and transform are inherited, but filters aren't
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #1188, opened 2023-04-25
URL-decode embedded winset

pali6:


OpenDreamClient/Interface/Controls/ControlBrowser.cs
```diff
@@ -89,6 +89,7 @@ public ControlBrowser(ControlDescriptor controlDescriptor, ControlWindow window)
 
                     // Reassemble the query params without element then convert to winset syntax
                     var query = queryParams.ToString();
+                    query = HttpUtility.UrlDecode(query);
```
wixoaGit, line 92: Shouldn't this be placed below the `Replace('&', ';')` call? Could break winsets with `&` in them.

wixoaGit (changes requested):

OpenDreamClient/Interface/Controls/ControlBrowser.cs
pali6: I don't think so? Placing it below that will turn e.g. `&gt;` into `;gt;` first so it doesn't get URL decoded

wixoaGit: Ah, true. File that under the "more robust parsing" then.

wixoaGit (approved):

PR #1190, opened 2023-04-25
Support for subdirectory paths for browse_rsc

pali6:
Trying to send a file to a path like `foo/bar/page.html` currently fails. ~~This PR makes it so the `foo` and `foo/bar` directories are created properly.~~ This PR strips `foo/bar/` from the path and creates `page.html` directly in the base cache directory, just like BYOND.

### DIFF
--- a/OpenDreamClient/Resources/DreamResourceManager.cs
+++ b/OpenDreamClient/Resources/DreamResourceManager.cs
@@ -56,7 +56,7 @@ namespace OpenDreamClient.Resources {
         }
 
         private void RxBrowseResource(MsgBrowseResource message) {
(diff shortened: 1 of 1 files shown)


pali6:
Wait actually this isn't how BYOND does it. Boo. Stupid BYOND. It actually just puts everything in the base directory.

pali6:
Now it's correct, but I hate it more!

ike709:
> Wait actually this isn't how BYOND does it. Boo. Stupid BYOND. It actually just puts everything in the base directory.

Please document this silliness in the code with a comment.

wixoaGit (approved):

Issue #1142, opened 2023-04-11
Unit tests should not reference `/atom` or its descendants


wixoaGit:
Some tests in `Content.Tests` currently use `/atom` and some of its descendants. These unit tests aren't ran in a fully initialized OpenDream server, and nearly none of `/atom`'s functionality works.

They should be changed to use `/datum` and `/atom` usage should be reserved for `Content.IntegrationTests`.

PR #1193, opened 2023-04-26
Assignment of DreamObject-s into usr

pali6:


Content.Tests/DMProject/Tests/Procs/usr_assignment.dm
```diff
@@ -0,0 +1,9 @@
+/proc/change_usr()
+	ASSERT(usr == null)
+	usr = new/mob
```
wixoaGit, line 3: Unit tests should not reference `/mob` (see #1142)

Set it to a `/datum` instead
```suggestion
	usr = new/datum
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #1194, opened 2023-04-26
Fix prob

pali6:
Currently prob(x) returns true with probability x+1. And also current prob implementation doesn't support non-integers correctly.
This PR fixes those issues by using RT's Prob and also adds a unit test.


ZeWaka (commented):
Could have a shortcut for 0? Probably doesn't matter though...

pali6:
> Could have a shortcut for 0? Probably doesn't matter though... 

Hmm I think I disagree. I could imagine code that somewhat relies on prob advancing the rng by one step.

wixoaGit (approved):

PR #1195, opened 2023-04-27
Implementation of text2path for proc paths and related fixes

pali6:
Implements `text2path` for proc paths. In order to do this correctly we need to know whether a proc was declared as `proc/` or `verb/` so this information is now passed from the compiler and stored in proc classes. Proc path stringification now also correctly uses `"proc/"` and `"verb/"` based on this.

Also fixes incorrect values for `text2path("")` and `text2path("/")`

### DIFF
--- a/OpenDreamRuntime/Objects/DreamObjectTree.cs
+++ b/OpenDreamRuntime/Objects/DreamObjectTree.cs
@@ -473,7 +473,7 @@ namespace OpenDreamRuntime.Objects {
         public bool TryGetGlobalProc(string name, [NotNullWhen(true)] out DreamProc? globalProc);
         public TreeEntry GetTreeEntry(DreamPath path);
         public TreeEntry GetTreeEntry(int typeId);
-        public bool TryGetTreeEntry(DreamPath path, out TreeEntry? treeEntry);
+        public bool TryGetTreeEntry(DreamPath path, [NotNullWhen(true)] out TreeEntry? treeEntry);
         public DreamObjectDefinition GetObjectDefinition(int typeId);
         public IEnumerable<TreeEntry> GetAllDescendants(TreeEntry treeEntry);
         public DreamValue GetDreamValueFromJsonElement(object value);
--- a/OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
+++ b/OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
@@ -2667,8 +2667,7 @@ namespace OpenDreamRuntime.Procs.Native {
 
             string? procName = null;
             if (isProcPath) {
-                // we strip the leading slash here
-                procName = path.FromElements(procElementIndex + 1).ToString()[1..];
+                procName = path.LastElement;
 
                 if (procElementIndex == 0) { // global procs
                     if (state.ObjectTree.TryGetGlobalProc(procName, out var globalProc) && globalProc.IsVerb == isVerb)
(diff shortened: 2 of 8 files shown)


OpenDreamRuntime/Objects/DreamObjectTree.cs
```diff
@@ -473,7 +473,7 @@ public sealed class TreeEntry {
         public bool TryGetGlobalProc(string name, [NotNullWhen(true)] out DreamProc? globalProc);
         public TreeEntry GetTreeEntry(DreamPath path);
         public TreeEntry GetTreeEntry(int typeId);
-        public bool TryGetTreeEntry(DreamPath path, out TreeEntry? treeEntry);
+        public bool TryGetTreeEntry(DreamPath path, [NotNullWhen(true)] out TreeEntry? treeEntry);
```
pali6, line 476: `NotNullWhen` is already in the method implementation. I just mirrored it here. A bit surprised that such mismatch gets no warnings during compilation.

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -2648,19 +2648,52 @@ public static DreamValue NativeProc_sorttext(NativeProc.State state)
         [DreamProc("text2path")]
         [DreamProcParameter("T", Type = DreamValueType.String)]
         public static DreamValue NativeProc_text2path(NativeProc.State state) {
-            if (!state.GetArgument(0, "T").TryGetValueAsString(out var text)) {
+            if (!state.GetArgument(0, "T").TryGetValueAsString(out var text) || text.Length == 0) {
```
wixoaGit, line 2651: Strings that are just whitespace also return null
```suggestion
            if (!state.GetArgument(0, "T").TryGetValueAsString(out var text) || string.IsNullOrWhiteSpace(text)) {
```

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -2648,19 +2648,52 @@ public static DreamValue NativeProc_sorttext(NativeProc.State state)
...
+                if (procElementIndex != -1)
+                    isVerb = true;
+            }
+
+            bool isProcPath = procElementIndex != -1;
+
+            string? procName = null;
+            if (isProcPath) {
+                // we strip the leading slash here
+                procName = path.FromElements(procElementIndex + 1).ToString()[1..];
```
wixoaGit, line 2671: `LastElement` works better here
```suggestion
                procName = path.LastElement;
```

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -2648,19 +2648,52 @@ public static DreamValue NativeProc_sorttext(NativeProc.State state)
...
+
+                if (procElementIndex == 0) { // global procs
+                    if (state.ObjectTree.TryGetGlobalProc(procName, out var globalProc) && globalProc.IsVerb == isVerb)
```
wixoaGit, line 2674:
```suggestion
                    if (procName != null && state.ObjectTree.TryGetGlobalProc(procName, out var globalProc) && globalProc.IsVerb == isVerb)
```

wixoaGit (changes requested):

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
pali6: This is already covered because `TryGetTreeEntry` or something returns false. It just throws some exception on empty string. But can't be hurt to have a bit of defensive programming I guess.

wixoaGit (approved):

PR #1200, opened 2023-05-01
Support for ANY macro keybind

pali6:
PR adds support for the `ANY` keybind in BYOND.

Remaining issues:
 - Both here and in BYOND the ANY macro and the specific macro (if there are both) fire. However, in BYOND it seems that the order of them firing is determined by the position in the dmf macro definition while in this implementation the ANY macro always fires first. However, this is a fairly minor issue and I sure hope no one relies on it. (clueless)
 - The ANY keybind does only a rudimentary check for keyboard focus unlike regular keybinds which go through RT's whole pipeline. This might result in some differences in behaviour but I was not able to spot anything specific right now.

Misc changes:
 - keybind modifiers (ctrl, shift, alt) should now work as far as I can tell
 - the REP+UP check now correctly looks at the uppercased keybind
 - semicolon added to the possible keys, not documented in BYOND docs but works there. There are likely other such ommissions too, perhaps most ASCII characters?
 - keybind parsing slightly cleaned up

### DIFF
--- a/OpenDreamClient/Interface/InterfaceMacro.cs
+++ b/OpenDreamClient/Interface/InterfaceMacro.cs
@@ -232,6 +232,8 @@ public sealed class InterfaceMacro : InterfaceElement {
 
     private readonly IEntitySystemManager _entitySystemManager;
     private readonly IUserInterfaceManager _uiManager;
+    private readonly IInputCmdContext _inputContext;
+    private readonly IInputManager _inputManager;
 
     private readonly bool _isRepeating;
     private readonly bool _isRelease;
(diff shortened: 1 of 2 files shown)


OpenDreamClient/Interface/InterfaceMacro.cs
```diff
@@ -39,45 +43,263 @@ public sealed class InterfaceMacroSet : InterfaceElement {
...
-        _isRepeating = Name.Contains("+REP");
-        _isRelease = Name.Contains("+UP");
+        if (_isAny && (parsedKeybind.Shift || parsedKeybind.Ctrl || parsedKeybind.Alt || parsedKeybind.Rep))
+            throw new Exception("ANY can only be combined with the +UP modifier");
 
         if (_isRepeating && _isRelease)
             throw new Exception("A macro cannot be both +REP and +UP");
 
+        if (_isAny) {
+            inputManager.FirstChanceOnKeyEvent += FirstChanceKeyHandler;
```
wixoaGit, line 265: `FirstChanceKeyHandler` ends up running regardless of which macro set is enabled.

pali6:
Let me know if you think it's better to instead register / unregister the handlers as macro sets switch.

wixoaGit:
> Let me know if you think it's better to instead register / unregister the handlers as macro sets switch.

This fix is fine

wixoaGit (approved):

PR #1202, opened 2023-05-01
Const eval for some math procs

pali6:
Added only those that BYOND considers const-evaluable (no `round` for example).


pali6:
Huh, locally tests were passing. Will check what's wrong tomorrow. Also I should remove the native proc implementations now.

pali6:
The unit test uses approximate equality because of the fact that .NET's trigonometry functions can and do behave differently on different platforms. See the remark [here](https://learn.microsoft.com/en-us/dotnet/api/system.mathf.tan?view=net-5.0). There is apparently a 1.1920929E-07 difference between Windows and Linux (or at least between my computer and the github actions server). Maybe at some point we will have to use something else than MathF and Math to avoid this issue.

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -1618,6 +1618,166 @@ private static void HandleSuffixPronoun(ref StringBuilder formattedString, ReadO
...
+                case DreamValue.DreamValueType.DreamObject when baseValue == DreamValue.Null: baseValueFloat = 0.0f; break;
+                default: throw new Exception("Invalid argument type for log");
+            }
+
+            switch (value.Type) {
+                case DreamValue.DreamValueType.Float: state.Push(new DreamValue(MathF.Log(value.MustGetValueAsFloat(), baseValueFloat))); break;
+                case DreamValue.DreamValueType.DreamObject when value == DreamValue.Null: state.Push(new DreamValue(MathF.Log(0.0f, baseValueFloat))); break;
+                default: throw new Exception("Invalid argument type for log");
+            }
+
```
wixoaGit, line 1753: `UnsafeGetValueAsFloat()`
```suggestion
            state.Push(new DreamValue(MathF.Log(value.UnsafeGetValueAsFloat(), baseValue.UnsafeGetValueAsFloat())));
```

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -1618,6 +1618,166 @@ private static void HandleSuffixPronoun(ref StringBuilder formattedString, ReadO
...
+        public static ProcStatus Abs(DMProcState state) {
+            DreamValue value = state.Pop();
+
+            switch (value.Type) {
+                case DreamValue.DreamValueType.Float: state.Push(new DreamValue(MathF.Abs(value.MustGetValueAsFloat()))); break;
+                case DreamValue.DreamValueType.DreamObject when value == DreamValue.Null: state.Push(new DreamValue(0.0f)); break;
+                default: throw new Exception("Invalid argument type for abs");
+            }
+
```
wixoaGit, line 1777: `UnsafeGetValueAsFloat()`
```suggestion
            state.Push(new DreamValue(MathF.Abs(value.UnsafeGetValueAsFloat())));
```

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -29,15 +29,7 @@ namespace OpenDreamRuntime.Procs.Native {
...
-    internal static class DreamProcNativeRoot {
-        [DreamProc("abs")]
-        [DreamProcParameter("A", Type = DreamValueTypeFlag.Float)]
-        public static DreamValue NativeProc_abs(NativeProc.Bundle bundle, DreamObject? src, DreamObject? usr) {
-            bundle.GetArgument(0, "A").TryGetValueAsFloat(out var number);
-
-            return new DreamValue(Math.Abs(number));
-        }
-
+    static class DreamProcNativeRoot {
```
wixoaGit, line 32: Don't remove `internal`
```suggestion
    internal static class DreamProcNativeRoot {
```

Content.Tests/DMProject/Tests/Expression/const_eval.dm
```diff
@@ -0,0 +1,56 @@
...
+	var/sqrttest = sqrt(123)
+	var/arcsintest = arcsin(sin(45))
+	var/arccostest = arccos(cos(123))
+	var/arctantest = arctan(tan(69))
+	var/log_test = log(10)
+	var/log10_test = log(10, 100)
+	var/arctan2_test = arctan(1, 3)
+	var/abs_test = abs(-213)
+
+#define EPSILON 1e-6
```
wixoaGit, line 15: Require less precision now that the arc procs are single-precision
```suggestion
#define EPSILON 4e-6
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #1203, opened 2023-05-01
Optimize the passing of arguments between proc calls

wixoaGit:
DreamProcArguments is now a ref struct that holds a ReadOnlySpan pointing to an array of DreamValues (a DMProc's stack in a majority of cases).

To allow for this they are no longer pushed to the stack and are instead operands to opcodes that call procs. The interpreter will also map named arguments to their position instead of putting them in a dictionary.

**Paradise init before:**
image
image
image
`[Master] Initializations complete within 68 seconds!` (after 3 runs)

**After:**
image
image
image
(PopProcArguments is PushArguments, PushProcArguments, and PushArgumentList rolled into one)
`[Master] Initializations complete within 61.7 seconds!` (also after 3 runs)


OpenDreamRuntime/DreamConnection.cs
```diff
@@ -219,13 +219,7 @@ public sealed class DreamConnection {
...
             }
 
-            DreamProcArguments topicArguments = new DreamProcArguments(new() {
-                new DreamValue(pTopic.Query),
-                new DreamValue(hrefList),
-                src
-            });
-
-            Client?.SpawnProc("Topic", topicArguments, Mob);
+            Client?.SpawnProc("Topic", Mob, new(pTopic.Query), new(hrefList), src);
```
wixoaGit, line 222:
```suggestion
            Client?.SpawnProc("Topic", usr: Mob, new(pTopic.Query), new(hrefList), src);
```

OpenDreamRuntime/Procs/DMProc.cs
```diff
@@ -441,8 +417,86 @@ sealed class DMProcState : ProcState {
...
 
-        public DreamProcArguments PopArguments() {
-            return Pop().MustGetValueAsProcArguments();
+        /// <summary>
+        /// Pops arguments off the stack and returns them in DreamProcArguments
+        /// </summary>
+        /// <param name="proc">The target proc we're calling. If null, named args or arglist() cannot be used.</param>
+        /// <param name="argumentsType">The source of the arguments</param>
+        /// <param name="argumentStackSize">The amount of items the arguments have on the stack</param>
+        /// <returns></returns>
```
wixoaGit, line 426:
```suggestion
        /// <returns>The arguments in a DreamProcArguments struct</returns>
```

PR #1208, opened 2023-05-03
Speed up some DreamObject casts

ike709:
Basic Paradise init comparisons show a consistent ~2 second speedup


ike709:
There's still more gains to be had.

ike709:
Meh any other optimizations can be future PRs.

wixoaGit (approved):

PR #1212, opened 2023-05-03
Get rid of DreamList event handlers

### LINKED ISSUE #1089: Assigning a list to `/atom.overlays` and `/atom.underlays` should copy it

### LINKED ISSUE #1210: Make overlay/underlay lists their own DreamList subtype
Removing the event handlers from DreamList would save 16 bytes per list

wixoaGit:
This moves `/atom.overlays`, `/atom.underlays`, `/area.contents`, `/client.screen`, and `args` into their own special DreamList types. This allows them to handle more operations than just adding/removing, and allows for the removal of the events on DreamList.

Resolves #1210
Fixes #1089

### DIFF
--- a/DMCompiler/DMStandard/Types/Client.dm
+++ b/DMCompiler/DMStandard/Types/Client.dm
@@ -1,6 +1,6 @@
 ﻿/client
 	var/list/verbs = null
-	var/list/screen = list()
+	var/list/screen = null
 	var/list/images = list() as opendream_unimplemented
 	var/list/vars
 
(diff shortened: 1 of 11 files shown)


out-of-phaze:
Does this do anything for `vars` (already a DreamList subtype, iirc) being terrible and not working properly (specifically, doing stuff with it like adding and subtracting doesn't create a new list like it should, it just gives null)? Do the new DreamList subtypes have that issue as well?

PR #1214, opened 2023-05-04
clamp rgb() arguments

pali6:
also adds a TODO about lowercase named arguments


pali6:
Tbh considering the HSL arguments and const-folding and upper/lowercaseness and null not being the same as absence of an argument this all should just be its own AST node instead of a proc.

wixoaGit (approved):

PR #1233, opened 2023-05-16
Implements \roman \Roman macros

redmoogle:
postPrefix is kinda cursed but also its the most sane way to do this.

### DIFF
--- a/OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
+++ b/OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
@@ -265,14 +265,31 @@ namespace OpenDreamRuntime.Procs {
 
             int[] numArr = new int[] { 1000, 500, 100, 50, 10, 5, 1 };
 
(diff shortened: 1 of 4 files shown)


OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -254,12 +254,39 @@ private static void HandleSuffixPronoun(ref StringBuilder formattedString, ReadO
...
+
+            var i = 0;
+            while(value != 0) {
+                if(value >= numArr[i]) {
+                    value -= numArr[i];
+                    formattedString.Append(arr[i]);
+                } else {
+                    i++;
+                }
+            }
```
wixoaGit, line 280: This needs to handle negative values.

While BYOND doesn't nicely handle infinity here, I think we should.

wixoaGit (changes requested):

wixoaGit (approved):

PR #1241, opened 2023-05-18
Adds 515 coordinate-style block() implementation

### LINKED ISSUE #855: coordinate version of `block()` not implemented
This is a version of `block()` that takes direct coordinates instead of turf references.
see: `block(StartX,StartY,StartZ, EndX=StartX,EndY=StartY,EndZ=StartZ)`

https://www.byond.com/docs/ref/#/proc/block

This might need handling in the compiler depending on implementation I think - unsure.
Bonus points: rewrite the existing `block()` to not be in softcode

ZeWaka:
title, https://www.byond.com/docs/ref/#/proc/block
fixes #855

Also organized the integration tests into their own files while i was there - can revert if needed, just thought the file was getting long


OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -178,21 +178,52 @@ static class DreamProcNativeRoot {
...
-                return new DreamValue(state.ObjectTree.CreateList());
-            if (!state.GetArgument(1, "End").TryGetValueAsDreamObjectOfType(state.ObjectTree.Turf, out var end))
-                return new DreamValue(state.ObjectTree.CreateList());
+            int x1;
+            int y1;
+            int z1;
+            int x2;
+            int y2;
+            int z2;
+            if (state.Arguments.Count <= 2) { // Is this the turf-style call?
```
wixoaGit, line 194: Using named arguments (`block(Start=..., End=...)`) will lead there to be 6 arguments because it'll fill the other four with null. It may be better to instead check if the first arguments is a turf.

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -178,21 +178,52 @@ static class DreamProcNativeRoot {
...
+                    return new DreamValue(state.ObjectTree.CreateList());
+                if (!state.GetArgument(1, "End").TryGetValueAsDreamObjectOfType(state.ObjectTree.Turf, out var end))
+                    return new DreamValue(state.ObjectTree.CreateList());
 
-            start.GetVariable("x").TryGetValueAsInteger(out var x1);
-            start.GetVariable("y").TryGetValueAsInteger(out var y1);
-            start.GetVariable("z").TryGetValueAsInteger(out var z1);
+                start.GetVariable("x").TryGetValueAsInteger(out x1);
+                start.GetVariable("y").TryGetValueAsInteger(out y1);
+                start.GetVariable("z").TryGetValueAsInteger(out z1);
```
wixoaGit, line 202: You should do `state.AtomManager.GetAtomPosition(start)` to avoid the slower path of getting 3 DreamObject vars.

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -178,21 +178,52 @@ static class DreamProcNativeRoot {
...
-            end.GetVariable("y").TryGetValueAsInteger(out var y2);
-            end.GetVariable("z").TryGetValueAsInteger(out var z2);
+                end.GetVariable("x").TryGetValueAsInteger(out x2);
+                end.GetVariable("y").TryGetValueAsInteger(out y2);
+                end.GetVariable("z").TryGetValueAsInteger(out z2);
+            } else { // coordinate-style
+                if (!state.GetArgument(0, "StartX").TryGetValueAsInteger(out x1)) {
+                    x1 = 1; // First three default to 1 when passed null or invalid
+                }
+                if (!state.GetArgument(1, "StartY").TryGetValueAsInteger(out y1)) {
```
wixoaGit, line 211: These are only used when an arg has a default value, but they should match the name used in `[DreamProcParameter]`

```suggestion
                if (!state.GetArgument(0, "Start").TryGetValueAsInteger(out x1)) {
                    x1 = 1; // First three default to 1 when passed null or invalid
                }
                if (!state.GetArgument(1, "End").TryGetValueAsInteger(out y1)) {
```

wixoaGit (changes requested):

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -178,38 +178,62 @@ static class DreamProcNativeRoot {
...
+                    startPos.Y = 1;
+                }
+                if (!state.GetArgument(2, "StartZ").TryGetValueAsInteger(out startPos.Z)) {
+                    startPos.Z = 1;
+                }
+                if (!state.GetArgument(3, "EndX").TryGetValueAsInteger(out endPos.X)) {
+                    endPos.X = startPos.X; // Last three default to the start coords if null or invalid
+                }
+                if (!state.GetArgument(4, "EndY").TryGetValueAsInteger(out endPos.Y)) {
+                    endPos.X = startPos.Y;
```
wixoaGit, line 214:
```suggestion
                    endPos.Y = startPos.Y;
```

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -178,38 +178,62 @@ static class DreamProcNativeRoot {
...
+                }
+                if (!state.GetArgument(5, "EndZ").TryGetValueAsInteger(out endPos.Z)) {
+                    endPos.X = startPos.Z;
```
wixoaGit, line 217:
```suggestion
                    endPos.Z = startPos.Z;
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #1242, opened 2023-05-20
feat(DMCompiler, Constant): support relative resource paths

### LINKED ISSUE #1087: Relative resource pathing does not work
DM can refer to resources using a path relative to the DM file. This does not work in OpenDream.

Insineer:
1. Support relative resource paths: resource paths can be set relatively to the .dm source file.
2. Emit compile error when resource file doesn't exist
3. Test loading of resource files

fix #1087

### DIFF
--- a/Content.Tests/DMProject/Tests/Expression/Constants/data/test.txt
+++ b/Content.Tests/DMProject/Tests/Expression/Constants/data/test.txt
@@ -1,0 +1,1 @@
+Test resource file's content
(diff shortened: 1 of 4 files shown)


wixoaGit:
I tested it in BYOND, and it turns out it doesn't actually search relative file paths. It appears the actual issue is [`FILE_DIR`](https://www.byond.com/docs/ref/#/DM/preprocessor/define/FILE_DIR).

wixoaGit (approved):

PR #1243, opened 2023-05-20
Displacement filter

amylizzle:
image


TestGame/icons/displace.dmi
wixoaGit, line 1: This needs added to the [solution file](https://github.com/OpenDreamProject/OpenDream/blob/master/OpenDream.sln#L15) so it's visible within the IDE

wixoaGit (changes requested):
The filter doesn't have the same affect as it does in BYOND
image
BYOND on the left

amylizzle:
Close, but no cigar 
image

amylizzle:
Done 
image

Minor differences are due to sampling algorithm and different `world.view` sizes I'm pretty sure. Sampling algo isn't exposed to content though, so we'll have to fix that in the upstream RT.

wixoaGit (approved):

PR #1244, opened 2023-05-20
fix(DMPreprocessor): skip whitespaces before #include

### LINKED ISSUE #1050: Indentation before #include causes undefined var error

Insineer:
fix #1050

### DIFF
--- a/Content.Tests/DMProject/Tests/Preprocessor/Include/test_module/test_file_to_include.dm
+++ b/Content.Tests/DMProject/Tests/Preprocessor/Include/test_module/test_file_to_include.dm
@@ -1,0 +1,3 @@
+// IGNORE
+
+var/const/test_var = "test_var_value"
--- a/Content.Tests/DMProject/Tests/Preprocessor/Include/test_module/test_file_to_include.dm
+++ b/Content.Tests/DMProject/Tests/Preprocessor/Include/test_module/test_file_to_include.dm
@@ -1,3 +1,3 @@
 // IGNORE
-
+// This file is used for #include testing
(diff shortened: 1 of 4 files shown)


Content.Tests/DMProject/Tests/Preprocessor/Include/test_module/test_file_to_include.dm
```diff
@@ -0,0 +1,3 @@
+// IGNORE
+
```
ike709, line 2:
```suggestion
// This file is used for #include testing
```

wixoaGit (approved):

PR #1245, opened 2023-05-20
Add a stylistic pragma for empty procs

ike709:
Adds a stylistic pragma for empty procs.

Hits on latest TG:
```
Warning OD3101 at code\game\atoms.dm:290:6: Empty proc detected - add an explicit "return" statement
Warning OD3101 at code\game\machinery\telecomms\broadcasting.dm:88:7: Empty proc detected - add an explicit "return" statement
```

I also marked `ProcessBlockInner()` as private because I could, and we'll no longer call it for completely empty (i.e. no set statements either) procs because what's the point.


DMCompiler/DM/Visitors/DMProcBuilder.cs
```diff
@@ -57,7 +57,18 @@ sealed class DMProcBuilder {
                 }
             }
 
-            ProcessBlockInner(procDefinition.Body, silenceEmptyBlockWarning : true);
+            if (procDefinition.Body.Statements.Length == 0) {
+                DMCompiler.Emit(WarningCode.EmptyProc, _proc.Location,"Empty proc detected - add an explicit \"return\" statement");
+                if (procDefinition.Body.SetStatements.Length > 0) { // We check this after because a proc with just set statements is still an empty proc
+                    ProcessBlockInner(procDefinition.Body, silenceEmptyBlockWarning : true);
+                } else {
+                    DMCompiler.Emit(WarningCode.EmptyBlock,_proc.Location,"Empty block detected");
```
wixoaGit, line 65: Doesn't make sense to emit both an `EmptyProc` and an `EmptyBlock`

wixoaGit (changes requested):

DMCompiler/DM/Visitors/DMProcBuilder.cs
```diff
@@ -57,7 +57,16 @@ sealed class DMProcBuilder {
...
+            if (procDefinition.Body.Statements.Length == 0) {
+                DMCompiler.Emit(WarningCode.EmptyProc, _proc.Location,"Empty proc detected - add an explicit \"return\" statement");
+                if (procDefinition.Body.SetStatements.Length > 0) { // We check this after because a proc with just set statements is still an empty proc
+                    ProcessBlockInner(procDefinition.Body, silenceEmptyBlockWarning : true);
+                }
+            } else {
+                ProcessBlockInner(procDefinition.Body, silenceEmptyBlockWarning : true);
+            }
+
+
```
wixoaGit, line 69: No reason to not just call `ProcessBlockInner()` regardless

```suggestion
            }
            
            ProcessBlockInner(procDefinition.Body, silenceEmptyBlockWarning : true);
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #1248, opened 2023-05-21
[RT] fix(text2num): speed up and allow any radix

### LINKED ISSUE #1247: text2num is too slow
**text2num** processes the empty text argument by throwing an exception. Result of this on the screenshot:

изображение

Around 5% of CPU resources in the 4 minutes long test of dynamic maps loading with dmm_suite.

Insineer:
* Rewrite implementation in a way that will not use exception throwing for empty input processing (valid according to BYOND).
* Make convertation logic the same as in BYOND (add unit tests for that logic enforcement)
* Allow any radix (up to 36), same as BYOND allows

fix #1247


ike709:
Did you profile again?

Insineer:
> Did you profile again?

No

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -2470,30 +2470,54 @@ public static DreamValue NativeProc_sorttext(NativeProc.State state)
             return new DreamValue(state.ResourceManager.SaveTextToFile(file, text) ? 1 : 0);
         }
 
+        private static int? StringToInteger(string value, int radix) {
```
wixoaGit, line 2473: BYOND also supports floats with a radix (`text2num("F.A", 16)` is 15.625)

This method should be moved to `DreamProcNativeHelpers.cs`

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -2470,30 +2470,54 @@ public static DreamValue NativeProc_sorttext(NativeProc.State state)
...
+            if (value == null || value.Length == 0)
```
wixoaGit, line 2474:
```suggestion
            if (string.IsNullOrWhiteSpace(value))
```

wixoaGit (changes requested):

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
Insineer: Alright. I've done more complex logic, which seems closer to BYOND implementation.

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
Insineer: I changed "string" to "ReadOnlySpan", so the string library function is no more working here.

Insineer:
> Did you profile again?

Checked text2num in profile again during working on further optimizations:

изображение

0.02% of CPU time compared to 5% before the fix. Moreover, the profile shows much fewer text2num calls cause they are so fast now that they fall between the profiler snapshots, and it does not see them. I even had to increase the number of snapshots per second for the function to appear in the profiler report.

amylizzle:
There is some moderately stupid behaviour BYOND has for this: `text2num("nanotrasen") == NaN`
I'm not sure if your implementation produces NaN anywhere, but I mention it anyway in case it's relied upon.

Insineer:
> There is some moderately stupid behaviour BYOND has for this: `text2num("nanotrasen") == NaN` I'm not sure if your implementation produces NaN anywhere, but I mention it anyway in case it's relied upon.

That's a really crazy one, but I supported this too

wixoaGit (approved):

PR #1252, opened 2023-05-22
Fix type filter for global datum loops

wixoaGit:
Loops like `for (var/datum/foo/D)` were enumerating every `/datum` instance rather than every `/datum/foo` instance.


firedhat (approved):
Fixes problem on my machine

PR #1264, opened 2023-05-28
Make a bunch of `GetValueAs` calls unsafe

ike709:
Essentially the only difference between `TryGet` and `MustGet` now is whether it returns false or throws an `InvalidCastException`.


OpenDreamRuntime/DreamValue.cs
```diff
@@ -259,15 +257,15 @@ public enum DreamValueType {
 
         public DreamList MustGetValueAsDreamList() {
             try {
-                return (DreamList) _refValue;
+                return (DreamList) _refValue!;
```
wixoaGit, line 260: Why can't we use `Unsafe.As<DreamList>()` here?

wixoaGit (approved):

PR #1266, opened 2023-06-02
Overhaul DreamObjects - Remove meta objects

wixoaGit:
This removes meta objects. What was previously on meta objects are now virtual methods on DreamObject itself, and engine types are implemented by inheriting DreamObject and overriding those. This removes tons of slow interface calls, and allows data to be stored directly on DreamObject rather than inside several separate dictionaries. You can also use more specific types like `DreamObjectMovable`, allowing for greater type safety.

`DreamValue.TryGetValueAsDreamObjectOfType()` was replaced with `DreamValue.TryGetValueAsDreamObject<T>() where T : DreamObject`

Second run Paradise init on my machine:
`[Master] Initializations complete within 42.5 seconds!`
Was previously ~53 seconds. BYOND takes ~35.


OpenDreamRuntime/AtomManager.cs
```diff
@@ -268,59 +252,60 @@ internal sealed class AtomManager : IAtomManager {
         }
 
         /// <summary>
-        /// Looks for an appearance and, if one does not exist, usually creates a new one. <br/>
-        /// If used with a turf, this will fail and throw some kinda KeyNotFoundException down the line.
+        /// Gets an atom's appearance. <br/>
```
wixoaGit, line 255:
```suggestion
        /// Gets an atom's appearance.
```

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -893,9 +889,9 @@ static class DreamProcNativeRoot {
 
             var arg = state.GetArgument(0, "Icon");
 
-            if (arg.TryGetValueAsDreamObjectOfType(state.ObjectTree.Icon, out var iconObj)) {
+            if (arg.TryGetValueAsDreamObject<DreamObjectIcon>(out var iconObj)) {
                 // Fast path for /icon, we don't need to generate the entire DMI
-                return new DreamValue(state.ObjectTree.CreateList(DreamMetaObjectIcon.ObjectToDreamIcon[iconObj].States.Keys.ToArray()));
+                return new DreamValue(state.ObjectTree.CreateList((iconObj).Icon.States.Keys.ToArray()));
```
wixoaGit, line 894:
```suggestion
                return new DreamValue(state.ObjectTree.CreateList(iconObj.Icon.States.Keys.ToArray()));
```

PR #1275, opened 2023-06-05
Fix operator being a valid var name

amylizzle:
Fix `var/mob/living/operator = usr` gives an `Invalid operator overload error`


Content.Tests/DMProject/Tests/Operators/Overloads/invalid_proc.dm
```diff
@@ -0,0 +1,11 @@
+// COMPILE ERROR
+
+//This results in an undefined proc error in BYOND also
+
+/datum/tester
+	proc/operator_()
+		return 555
+
+/proc/RunTest()
+	var/datum/tester = new()
```
wixoaGit, line 10: The type of this var is `/datum`, not `/datum/tester`.

Changing this to `var/datum/tester/tester = new()` causes the proc to work in BYOND.

amylizzle: oh whoops

wixoaGit (approved):

Issue #1267, opened 2023-06-02
Case Sensitivity on Resource Paths


amylizzle:
Introduced by #1242 - resource paths are case sensitive on case sensitive OS (ie, linux) when they shouldn't be. In BYOND, resource paths are case insensitive regardless of the OS.

Insineer:
Were resources with wrong-cased paths loaded by OpenDream on Linux before my fix? Or it just silently ignored them?

I've tried this with BYOND, and there are really case insensitive paths regardless of the OS. Seems like BYOND implements some specific files searching logic for Linux.

amylizzle:
Not sure but I'd guess they were ignored silently. Your fix just makes them not silent.

wixoaGit:
This was previously a runtime error (#591). #1242 just moved the error from runtime to compile-time.

wixoaGit:
Duplicate of #591

PR #1277, opened 2023-06-06
Explicit expression wrapping in AST

### LINKED ISSUE #1225: Assign being treated as named argument
TG contains this code: `if(isnull((smoothing_groups = GLOB.bitflag_lists[txt_signature])))`

It causes the following runtime: `[ERRO] world.log: Exception occurred: /proc/isnull has no argument named smoothing_groups`

According to wixoa, the compiler is treating that as if the parentheses aren't there and it's a named argument.

### LINKED ISSUE #1276: Unable to dereference from `new`
```dm

/datum/foo
	var/bar = 5

/proc/RunTest()
	ASSERT((new/datum/foo()).bar == 5)
```

image

martinlyra:
- Fixes #1225 

Adds the `DMASTExpressionWrapping` which has no dedicated visitor; instead, it reroutes the visitor directly to the wrapped expression's own visitor instead. If this is undesirable, do let me know. This change does decommission `DMASTIdentifierWrapped` as it is functionally the same but with its own visitor.

### DIFF
--- a/DMCompiler/Compiler/DM/DMParser.cs
+++ b/DMCompiler/Compiler/DM/DMParser.cs
@@ -1561,22 +1561,6 @@ namespace DMCompiler.Compiler.DM {
             }
         }
 
-        public DMASTExpression? ExpressionParameter() {
-            if(Current().Type == TokenType.DM_LeftParenthesis)
-            {
-                DMASTExpression? wrapped = Expression();
-
-                if (wrapped is DMASTAssign assignment)
-                    return new DMASTAssign(
-                        assignment.Location,
-                        new DMASTConstantNull(assignment.Location),
-                        assignment
-                    );
-                return wrapped;
-            }
-            return Expression();
-        }
-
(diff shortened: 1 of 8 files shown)


martinlyra:
Alright. After some more debugging, I was too quick to when I made the conclusion that #1267 was related to wrapping, when it was not.

DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -1561,6 +1561,22 @@ public partial class DMParser : Parser<Token> {
...
+                    return new DMASTAssign(
+                        assignment.Location,
+                        new DMASTConstantNull(assignment.Location),
+                        assignment
+                    );
+                return wrapped;
+            }
+            return Expression();
+        }
+
```
martinlyra, line 1579:
```suggestion
```

wixoaGit (approved):

PR #1278, opened 2023-06-06
Implement `Path` for `NewPath` class

### LINKED ISSUE #1276: Unable to dereference from `new`

martinlyra:
- Fixes #1276

### DIFF
--- a/Content.Tests/DMProject/Tests/Expression/Derefs/new_deref.dm
+++ b/Content.Tests/DMProject/Tests/Expression/Derefs/new_deref.dm
@@ -1,8 +1,6 @@
 
 //# issue 1276
 
-#define ASSERT(x) x
-
 /datum/foo
 	var/bar = 5
 
--- a/Content.Tests/DMProject/Tests/Expression/Derefs/new_deref.dm
+++ b/Content.Tests/DMProject/Tests/Expression/Derefs/new_deref.dm
@@ -1,0 +1,16 @@
+
+//# issue 1276
+
+#define ASSERT(x) x
+
+/datum/foo
+	var/bar = 5
+
+/proc/RunTest()
+	var/datum/foo/a
+	ASSERT((a = new/datum/foo()).bar == 5)
+
+	var/b
+	ASSERT((b = new/datum/foo().bar) == 5)
+
+	ASSERT(new/datum/foo().bar == 5)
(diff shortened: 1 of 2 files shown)


Content.Tests/DMProject/Tests/Expression/Derefs/new_deref.dm
```diff
@@ -0,0 +1,16 @@
+
+//# issue 1276
+
+#define ASSERT(x) x
```
wixoaGit, line 4: Why do you disable `ASSERT()` here?

DMCompiler/DM/Expressions/Builtins.cs
```diff
@@ -82,6 +82,8 @@ sealed class NewPath : DMExpression {
             proc.PushType(typeId);
             proc.CreateObject(argumentInfo.Type, argumentInfo.StackSize);
         }
+
+        public override DreamPath? Path => _targetPath;
```
wixoaGit, line 86: This code snippet compiles fine in BYOND, showing that this is actually treated as `(new /datum/test2):foo`
```
/datum/test/var/foo = "foo"
/datum/test2/var/bar = "bar"

/proc/main()
    world.log << ((new /datum/test2).foo)
```

You could achieve this by adding `case NewPath:` [here](https://github.com/OpenDreamProject/OpenDream/blob/master/DMCompiler/DM/Visitors/DMVisitorExpression.cs#L476). Though I wonder if there'd be no harm in the additional type safety this solution provides?

wixoaGit (changes requested):

Content.Tests/DMProject/Tests/Expression/Derefs/new_deref.dm
martinlyra: Ah, I was supposed to remove that. I was debugging the compiler, but since it loaded the standard library which would lead me into stacks that I was absolutely not interested in debugging. Since ASSERT is stdlib, I had to define it so I could debug the compiler on this particular test case. I was supposed to remove this `#define` before PR-ing it, which I did neglect to in another PR until I caught and fixed it myself.

DMCompiler/DM/Expressions/Builtins.cs
wixoaGit: This is fine as-is, the additional type safety is better and nothing depends on this to my knowledge.

wixoaGit (approved):

PR #1294, opened 2023-06-14
Fixes "operator" not working as a valid type name

Altoids1:
Previously, a phrase like "a/b/operator/c" would be interpreted by Path() as an imminent `a/b/operator/()` divison overload, to be followed by a `c` identifier token.

This was causing compile errors in Yogs, where NTSL makes use of a legit type called operator.

This allows Yogs to compile :^)
## Changelog
- Using `operator` as a type name is now allowed and behaves similarly to BYOND.

### DIFF
--- a/DMCompiler/Compiler/DM/DMParser.cs
+++ b/DMCompiler/Compiler/DM/DMParser.cs
@@ -366,7 +366,18 @@ namespace DMCompiler.Compiler.DM {
                     if (pathElement != null) {
                         if(pathElement == "operator") {
                             Token operatorToken = Current();
-                            if(Check(OperatorOverloadTypes)) {
+                            if(Current().Type == TokenType.DM_Slash) {
+                                //Up to this point, it's ambiguous whether it's a slash to mean operator/(), like the division operator overload
+                                //or "operator" just being used as a normal type name, as in a/operator/b/c/d
+                                Token peekToken = Advance();
+                                if (peekToken.Type == TokenType.DM_LeftParenthesis) { // Disambiguated as an overload
+                                    operatorFlag = true;
+                                    pathElement += operatorToken.PrintableText;
+                                } else { //Otherwise it's just a normal path, resume
+                                    ReuseToken(operatorToken);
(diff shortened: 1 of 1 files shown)


DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -366,7 +366,19 @@ public partial class DMParser : Parser<Token> {
...
+                            if(Current().Type == TokenType.DM_Slash) {
+                                //Up to this point, it's ambiguous whether it's a slash to mean operator/(), like the division operator overload
+                                //or "operator" just being used as a normal type name, as in a/operator/b/c/d
+                                Token peekToken = Advance();
+                                if (peekToken.Type == TokenType.DM_LeftParenthesis) { // Disambiguated as an overload
+                                    operatorFlag = true;
+                                    pathElement += operatorToken.PrintableText;
+                                } else { //Otherwise it's just a normal path, resume
+                                    ReuseToken(operatorToken);
+                                    Warning("Using \"operator\" as a path element is ambiguous", Current()); // NOTE: Maybe should be SoftReservedKeyword? Unsure!
```
Altoids1, line 378: Any opinions on marking this as SoftReservedKeyword? That'd be an error in the default config.

ike709: SoftReservedKeyword is better than a warning, but I don't know why we'd emit anything unless it impacts behavior

Altoids1: This can be spooky for behaviour, since `operator/()` has a few places of grammatical ambiguity in the language. I won't be completely surprised if something somewhere accidentally relies BYOND resolving this vexing parse one way instead of the other. A warning is deserved.

DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -366,7 +366,19 @@ public partial class DMParser : Parser<Token> {
...
+                                }
+                            }
+                            else if(Check(OperatorOverloadTypes)) {
```
wixoaGit, line 381:
```suggestion
                            } else if(Check(OperatorOverloadTypes)) {
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #1300, opened 2023-06-18
Minor DMTest suite refactor

martinlyra:
image

So I start with looking at how I can improve the test by giving me the information I need, and what tests are 'accidentally' fixed. Thankfully it was minor stuff.

- Moves four tests that were seemingly fixed by prior PRs
- Makes the DMTest explicitly log what file it is testing on and write pass when successful.

God I don't like working with NUnit, that shit is a fucking memetic black hole and I don't find the information I *need*.


Content.Tests/DMProject/Tests/Statements/Control Flow/sleep/no_paren.dm
wixoaGit, line 1: This test is actually not fixed, we don't support `sleep()` without parentheses yet. I think this is being parsed as a label named "sleep" and a statement containing an unused expression.

Its use of `world.realtime` is also not good, since it has very low precision and could spontaneously fail. The `ASSERT()` should also be doing `== 20` not `== 0` since 2 seconds should have passed.

wixoaGit (changes requested):

wixoaGit (approved):

PR #1312, opened 2023-06-19
Add a couple broken tests for operators

amylizzle:
Salvaged the runtime operator tests from op overloads work I did.

Right now both of these fail, so added them to the broken tests folder.

### DIFF
--- a/Content.Tests/DMProject/Broken Tests/Operators/valid_and_null.dm
+++ b/Content.Tests/DMProject/Broken Tests/Operators/valid_and_null.dm
@@ -1,0 +1,129 @@
+//simple test of all basic operators with valid and C(null) arguments
+//We can't be having const folding in here
+#define C(X) pick(X) 
+/proc/RunTest()
+	var/a = C(2)
+	ASSERT(!C(null) == C(1))
+	ASSERT(!!C(null) == C(0))
+
+	ASSERT(~C(1) == 16777214)
+	ASSERT(~C(0) == 16777215)
+	ASSERT(~C(null) == 16777215)
+
+	ASSERT(C(1) + C(1) == C(2))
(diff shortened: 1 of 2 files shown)


Content.Tests/DMProject/Broken Tests/Operators/valid_and_null.dm
```diff
@@ -0,0 +1,129 @@
...
+	ASSERT(C(2) ** C(null) == C(1))
+	ASSERT(C(null) ** C(2) == C(0))
+
+	ASSERT(C(2) * C(3) == 6)
+	ASSERT(C(2) * C(null) == C(0))
+	ASSERT(C(null) * C(2) == C(0))
+
+	ASSERT(C(4) / C(2) == C(2))
+	ASSERT(C(null) / C(2) == C(0))
+	ASSERT(C(2) / C(null) == C(2))
```
wixoaGit, line 60: In BYOND this gives `2` if const folded at compile-time but `Undefined operation: 2 / null` if done at runtime. Are we opting for the compile-time behavior at runtime too here?

wixoaGit (changes requested):

Content.Tests/DMProject/Broken Tests/Operators/valid_and_null.dm
amylizzle: Well with the `pick()` macro we're explicitly asking for runtime behaviour, so I guess it ought to do that, though it pains me that they're different.

amylizzle, line 60:
```suggestion
	//ASSERT(C(2) / C(null) == C(2)) //runtime: undefined operation
```

wixoaGit (approved):

Issue #1309, opened 2023-06-18
Constant evalation for native functions is not implemented


martinlyra:
https://github.com/OpenDreamProject/OpenDream/blob/1b80c46ef2ffcabcd4e6cea0cffe52ac65fbc0b1/DMCompiler/DM/DMExpression.cs#L51-L55

Should be implemented for `ProcCall`

Besides the compiler has a superficial handling of knowing and handling what global procs are native and which are not.

Associated test file: https://github.com/OpenDreamProject/OpenDream/blob/1b80c46ef2ffcabcd4e6cea0cffe52ac65fbc0b1/Content.Tests/DMProject/Broken%20Tests/Const/ConstProc.dm#L1-L6

PR #1313, opened 2023-06-19
Split RT and DM shared code

martinlyra:
When attempting to do a PR for #1309, I was using the OpenDreamRuntime and that was when I realised the solution was a hellscape and a half, so after one-two days of deliberation in the Discord. It was suggested to split `OpenDreamShared` into two shared-code libraries instead.

- `DMShared` intended to be used by compiler and the runtime. Although it is included in `OpenDreamShared` as dependency due to `Content.Tests`, `DMDisassembler`, and `OpenDreamClient` also needing the items from there.
- `OpenDreamShared` contains `RobustToolbox` code and dependency, and its use should be limited to the OpenDream game binaries. The compiler must avoid using it as possible.

On that note. This does not remove the RobustToolbox dependency off DMCompiler, however, due to it still using analyser and attributes from that library via OpenDreamShared. `Jetbrains.Annotations` too, for some reason.


martinlyra:
Merge conflict my fucking

wixoaGit (changes requested):
The DMShared project needs added to the `OpenDream.sln` file.

wixoaGit:
So, big problem with this. Making the client reference a new `DMShared` assembly makes the RT sandbox very upset. You can't launch the client at all.

`OpenDreamShared` is only allowed by the sandbox because it is specially whitelisted [here](https://github.com/space-wizards/RobustToolbox/blob/master/Robust.Shared/ContentPack/Sandbox.yml#L13)

martinlyra:
I will hold off on it for now until we have have a plan on what we should to do with this project, now that the sandbox security has reminded us that it does exist.

PR #1319, opened 2023-06-21
Make \ref[] format the same as byond

amylizzle:
type IDs are accurate for major types but does not fully adhere to [*the list*](https://gn32.uk/f/byond-typeids.txt) but otherwise make \ref return a string formatted like `[0x1000000]`

### DIFF
--- a/Content.Tests/DMProject/Tests/Text/ref.dm
+++ b/Content.Tests/DMProject/Tests/Text/ref.dm
@@ -2,6 +2,16 @@
 	var/obj/thing = new()
 	var/obj/thing2 = new()
 	var/obj/thing3 = new()
+	var/not_this_one = "\ref[thing]"
 	var/ref = "\ref[thing2]"
+	var/not_this_one_either = "\ref[thing]"
 	var/test_thing = locate(ref)
 	ASSERT(test_thing == thing2)
+
+	var/string = "farts"
+	var/string_ref = "\ref[string]"
+	ASSERT(locate(string_ref) == string)
+
+	var/proc_ref = "\ref[/proc/RunTest]"
+	ASSERT(length(proc_ref)==10)
+
--- a/Content.Tests/DMProject/Tests/Text/ref.dm
+++ b/Content.Tests/DMProject/Tests/Text/ref.dm
@@ -17,5 +17,5 @@
 	ASSERT(length(proc_ref)==12)
 
 	var/list/test = list(1,2,3)
-	ASSERT(lowertext(copytext("\ref[test]",4,5)) == "f")
+	ASSERT(copytext("\ref[test]",4,5) == "f")
 	ASSERT(locate("\ref[test]") == test)
--- a/Content.Tests/DMProject/Tests/Text/ref.dm
+++ b/Content.Tests/DMProject/Tests/Text/ref.dm
@@ -1,3 +1,4 @@
+#define ASSERT(expr) ((expr) ? null : CRASH("Assertion Failed: " + #expr))
 /proc/RunTest()
 	var/obj/thing = new()
 	var/obj/thing2 = new()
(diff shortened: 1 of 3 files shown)


wixoaGit:
Could you also wrap the ref in `[]` and give lists the ID `0xf`? SS13's view variables uses this to determine if a list is a special list.

OpenDreamRuntime/DreamManager.cs
```diff
@@ -171,8 +171,20 @@ internal partial class DreamManager : IDreamManager {
...
+                        case DreamObjectTurf: refType = RefType.DreamObjectTurf; break;
+                        case DreamObjectMob: refType = RefType.DreamObjectMob; break;
+                        case DreamObjectArea: refType = RefType.DreamObjectArea; break;
+                        case DreamObjectClient: refType = RefType.DreamObjectArea; break;
+                        case DreamObjectImage: refType = RefType.DreamObjectImage; break;
+                        case DreamList: refType = RefType.DreamObjectList; break;
+                        default: {
+                            refType = RefType.DreamObjectDatum;
+                            if(refObject.IsSubtypeOf(_objectTree.Obj))
+                                refType = RefType.DreamObject;
```
wixoaGit, line 184: Ref ID 0xF should be reserved for DreamList but not any of its descendants.
```suggestion
                        default: {
                            refType = RefType.DreamObjectDatum;
                            if(refObject.IsSubtypeOf(_objectTree.Obj))
                                refType = RefType.DreamObject;
                            else if (refObject.GetType() == typeof(DreamList))
                                refType = RefType.DreamObjectList;
```

OpenDreamRuntime/DreamManager.cs
```diff
@@ -204,12 +216,12 @@ internal partial class DreamManager : IDreamManager {
                 throw new NotImplementedException($"Ref for {value} is unimplemented");
             }
 
-            // The first digit is the type, i.e. 1 for objects and 2 for strings
-            return $"{(int) refType}{idx}";
+            // The first digit is the type
+            return $"[0x{((int) refType+idx):X}]";
```
wixoaGit, line 220: BYOND gives these in lowercase
```suggestion
            return $"[0x{((int) refType+idx):x}]";
```

OpenDreamRuntime/IDreamManager.cs
```diff
@@ -42,13 +42,20 @@ public interface IDreamManager {
...
+        DreamObject = 0x2000000,
+        DreamObjectMob = 0x3000000,
+        DreamObjectArea = 0x4000000,
+        DreamObjectClient = 0x5000000,
+        DreamObjectImage = 0xD000000,
+        DreamObjectList = 0xF000000,
+        DreamObjectDatum = 0x21000000,
+        String = 0x6000000,
+        DreamType = 0x9000000, //in byond type is from 0x8 to 0xb, but fuck that
+        DreamResource = 0xFF000000, //doesn't exist in byond
```
wixoaGit, line 57: This would be the equivalent to a file (0x27 in BYOND)

wixoaGit (changes requested):

OpenDreamRuntime/IDreamManager.cs
amylizzle, line 57:
```suggestion
        DreamResource = 0x27000000, //Equivalent to file
```

Content.Tests/DMProject/Tests/Text/ref.dm
```diff
@@ -0,0 +1,21 @@
...
+
+	var/string = "farts"
+	var/string_ref = "\ref[string]"
+	ASSERT(locate(string_ref) == string)
+
+	var/proc_ref = "\ref[/proc/RunTest]"
+	ASSERT(length(proc_ref)==12)
+
+	var/list/test = list(1,2,3)
+	ASSERT(lowertext(copytext("\ref[test]",4,5)) == "f")
```
amylizzle, line 20:
```suggestion
	ASSERT(copytext("\ref[test]",4,5) == "f")
```

Content.Tests/DMProject/Tests/Text/ref.dm
```diff
@@ -0,0 +1,21 @@
+#define ASSERT(expr) ((expr) ? null : CRASH("Assertion Failed: " + #expr))
```
wixoaGit, line 1: No need to re-define `ASSERT()`
```suggestion
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #1326, opened 2023-06-24
Include turf contents in area contents

### LINKED ISSUE #1004: Area contents don't fully work
An area's `contents` is supposed to hold every turf and movable inside their area. Right now they only contain their turfs.

### LINKED ISSUE #1068: Cache area contents
Similar to https://github.com/OpenDreamProject/OpenDream/issues/1067 except areas should cache the turfs within them and then use that + the `turf.contents` cache to build its actual contents list when accessed.

I haven't done any testing to see how much this would help, or if SS13 even uses `area.contents` enough for it to matter, but I figured I should open the issue anyways so we don't forget about it.

Related to https://github.com/OpenDreamProject/OpenDream/issues/1004

wixoaGit:
Loops like `for (var/i in area)` should contain turf contents in addition to turfs.

This fixes light switches, area power updates, and likely a bunch of other things.

Fixes #1004
Resolves #1068

screenshot

### DIFF
--- a/OpenDreamRuntime/Objects/Types/DreamList.cs
+++ b/OpenDreamRuntime/Objects/Types/DreamList.cs
@@ -900,7 +900,7 @@ namespace OpenDreamRuntime.Objects.Types {
     // area.contents list
     public sealed class AreaContentsList : DreamList {
         private readonly DreamObjectArea _area;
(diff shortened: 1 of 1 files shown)


ike709:
Fixes https://github.com/OpenDreamProject/OpenDream/issues/1004 ?

Issue #1344, opened 2023-06-30
skin elements with an empty name throw runtimes


ZeWaka:
`System.ArgumentException: An item with the same key has already been added. Key:`

`at OpenDreamClient.Interface.InterfaceMenu.AddChild(ElementDescriptor descriptor) in OpenDreamClient\Interface\InterfaceMenu.cs:line 49`
Example elements that get flagged as duplicates:
```
elem
	name = ""
	command = ""
	category = "&Audio"
```
These end up rendering as spacers/horizontal rules, which is why they're used for Goon.
![image](https://github.com/OpenDreamProject/OpenDream/assets/4741640/a0cef96a-4081-4e3e-9c23-be0d4c5be087)

amylizzle:
Fixed by #1402

PR #1345, opened 2023-07-01
Split ID and name attributes in skin controls

wixoaGit:
We've been treating ID and name as the same thing which isn't correct. These are now separate and uses of `Name` have been switched to `Id` where I think is appropriate.

This fixes attempts to reparent existing macros with names different from their id. That and a change to wincloning macro sets fixes non-hotkey mode in modern Paradise.

I also made every file in the `Interface/` folder use file-scoped namespaces.


ZeWaka:
Fixes #1344?

wixoaGit:
I don't think it would. Menu elements are still added to a dictionary using their id, and the id is set to the name if there isn't an explicit one.

PR #1357, opened 2023-07-02
Add support for non-string keys in DMM lists

wixoaGit:


wixoaGit:
I wanted to instead implement modified types which would make this unnecessary, but that's going to end up taking more time than I thought it would.

PR #1364, opened 2023-07-09
Fix nullability of a bunch of DreamIcon stuff

amylizzle:
Appearance was nullable but wasn't marked as such. It causes a lot of exceptions when appearances haven't loaded yet.


amylizzle:
Tried to make this use `[Dependency]` instead of the entity system thing, but it doesn't work. Official comment from RT dev:
wait do overlays get their dependencies injected automatically

let me see

wow they don't

overlay code sucks

wixoaGit:
To my knowledge you can only `[Dependency]` entity systems if the field is itself in an entity system.

PR #1365, opened 2023-07-10
Allow server to be started without using `--cvar`

amylizzle:
Lets you start a server by `OpenDreamServer /path/to/env.json` instead of needing to use `--cvar "opendream.json_path=/path/to/env.json"` which is much more intuitive. Note that if the cvar is defined, it takes precedence.


amylizzle:
it does not. TBF, RT's arg handling literally just does this with a bit of extra parsing.

wixoaGit (approved):

PR #1366, opened 2023-07-11
Move `parent_type` handling to the first step in DMObjectBuilder

wixoaGit:
Moves `parent_type` handling from step 6 to step 1. It's now treated as part of the object definition rather than a var override, so every step from there on out is aware of the object's correct parent.


wixoaGit:
Breaks Paradise somehow. Will look into later.

wixoaGit:
Tried it again. No longer causes any issues?

PR #1367, opened 2023-07-12
RT Update 136.0.1

amylizzle:
Pretty much just replacing RT's Vector2 with `System.Numeric`'s and adjusted syntax to match (ie, `vector2var + (1,2)` is no longer valid, you have to explicitly instantiate a new Vector2)

Browser elements can be debugged on Linux! I'm so happy!


wixoaGit:
`System.Numerics` is probably something we can put in the global usings file.

wixoaGit (approved):

amylizzle:
I did not know about global using, that is useful.

wixoaGit (approved):

PR #1371, opened 2023-07-13
Optimize `LocateCoord()` for invalid turfs

ike709:
Credit to moony for the idea, they're just too lazy to PR it.

Since valid turfs start at 1, DreamMapManager should just return null for any coords less than 1 anyways.

Entirely untested.


OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -1915,8 +1915,8 @@ private static void HandleSuffixPronoun(ref StringBuilder formattedString, ReadO
             var z = state.Pop();
             var y = state.Pop();
             var x = state.Pop();
-            if (x.TryGetValueAsInteger(out var xInt) && y.TryGetValueAsInteger(out var yInt) &&
-                z.TryGetValueAsInteger(out var zInt)) {
+            if (x.TryGetValueAsInteger(out var xInt) && xInt > 0 && y.TryGetValueAsInteger(out var yInt) &&
+                yInt > 0 && z.TryGetValueAsInteger(out var zInt) && zInt > 0) {
```
wixoaGit, line 1919: This isn't an optimization as-is.

I think what moony was trying to say is that `DreamMapManager.TryGetTurfAt()` will give you `null` for the location (0, 0, 0). Meaning you can use `(int)UnsafeGetValueAsFloat()` and get `null` for any given invalid value, which removes all the TryGets.

wixoaGit (changes requested):

wixoaGit (approved):

PR #1374, opened 2023-07-16
Simplifies regex replacement for \L and \l

redmoogle:
Should fully replicate byond. I wonder if we could have our own regex parser


wixoaGit:
This doesn't work if the regex contains anything before the `\l`

wixoaGit (approved):

Issue #1341, opened 2023-06-28
`/world/Topic()` is not supported


wixoaGit:
https://www.byond.com/docs/ref/#/world/proc/Topic

~Looks like it's an HTTP GET request handler on the server~ It's a custom TCP protocol!

As used by the BYOND.TopicSender library [here](https://github.com/Cyberboss/BYOND.TopicSender/blob/d9e273943c2f9cc91b205d0050b1190ee475de75/Byond.TopicSender/TopicClient.cs#L113)

ike709:
Reopened since #1382 was reverted by #1398

Cyberboss:
Should be fixed soon or OD isn't gonna be in the initial release of TGS 6

Cyberboss:
And by soon, I mean by about a month from now

PR #1382, opened 2023-07-25
Add Topic/Export capabilities

ZephyrTFA:
Note that this only implements raw functionality and does not implement the encrypted Topic protocol because I don't know if anyone has ever figured that one out
Closes https://github.com/OpenDreamProject/OpenDream/issues/1341

### DIFF
--- a/OpenDreamRuntime/DreamManager.Connections.cs
+++ b/OpenDreamRuntime/DreamManager.Connections.cs
@@ -51,7 +51,7 @@ namespace OpenDreamRuntime {
             _netManager.RegisterNetMessage<MsgAckLoadInterface>(RxAckLoadInterface);
             _netManager.RegisterNetMessage<MsgSound>();
 
-            var worldTopicAddress = new IPEndPoint(IPAddress.Loopback, _configManager.GetCVar(CVars.NetPort));
+            var worldTopicAddress = new IPEndPoint(IPAddress.Loopback, _netManager.Port);
             _sawmill.Debug($"Binding World Topic at {worldTopicAddress}");
             _worldTopicSocket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp) {
                 ReceiveTimeout = 5000,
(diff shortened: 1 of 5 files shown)


wixoaGit (changes requested):

OpenDreamRuntime/DreamManager.Connections.cs
```diff
@@ -32,6 +50,119 @@ public sealed partial class DreamManager {
...
+                return Encoding.ASCII.GetString(buffer[5..^1]);
+            }
+
+            if (_worldTopicSocket is null)
+                throw new InvalidOperationException("Attempted to start the World Topic Listener without a valid socket bind address.");
+
+            while (!cancellationToken.IsCancellationRequested) {
+                try {
+                    var pending = await _worldTopicSocket.AcceptAsync(cancellationToken);
+                    var topic = await ParseByondTopic(pending);
```
Cyberboss, line 103: TGS won't send simultaneous topics but it's still probably a case you want to handle. Especially since a sleep() in world/Topic should block the entire response. Everything after the Accept should be done asynchronously without being immediately awaited so further accepts can be processed simultaneously.

Cyberboss (changes requested):

OpenDreamRuntime/DreamManager.Connections.cs
```diff
@@ -32,6 +50,119 @@ public sealed partial class DreamManager {
             _netManager.RegisterNetMessage<MsgLoadInterface>();
             _netManager.RegisterNetMessage<MsgAckLoadInterface>(RxAckLoadInterface);
             _netManager.RegisterNetMessage<MsgSound>();
+
+            var worldTopicAddress = new IPEndPoint(IPAddress.Loopback, _netManager.Port);
```
Cyberboss, line 54: Won't this only listen on localhost? For parity, you'll eventually have to implement cross server comms

OpenDreamRuntime/DreamManager.Connections.cs
```diff
@@ -32,6 +50,119 @@ public sealed partial class DreamManager {
...
+            _sawmill.Debug($"Binding World Topic at {worldTopicAddress}");
+            _worldTopicSocket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp) {
+                ReceiveTimeout = 5000,
+                SendTimeout = 5000,
+                ExclusiveAddressUse = false,
+            };
+            _worldTopicSocket.Bind(worldTopicAddress);
+            _worldTopicSocket.Listen();
```
Cyberboss, line 62: Pretty sure this is blocking?

Cyberboss:
@wixoaGit I should have asked you to create the tag AFTER this since it's still required. Could you make a new one once it's merged please?

wixoaGit:
> @wixoaGit I should have asked you to create the tag AFTER this since it's still required. Could you make a new one once it's merged please?

Will do. Will TGS only support tagged versions?

Cyberboss:
> > @wixoaGit I should have asked you to create the tag AFTER this since it's still required. Could you make a new one once it's merged please?
> 
> Will do. Will TGS only support tagged versions?

It will support both versioned tags and any git SHA, but it prefers sticking to versions for consistency reasons

wixoaGit (approved):

PR #1391, opened 2023-07-31
Fix `locate()` with a tag surrounded by brackets

wixoaGit:
Doing `locate("\[abc]")` is no different from doing `locate("abc")`

### DIFF
--- a/OpenDreamRuntime/DreamManager.cs
+++ b/OpenDreamRuntime/DreamManager.cs
@@ -290,6 +290,7 @@ namespace OpenDreamRuntime {
 
             // Search for an object with this ref as its tag
             // Note that surrounding [] are stripped out at this point, this is intentional
+            // Doing locate("[abc]") is the same as locate("abc")
             if (Tags.TryGetValue(refString, out var tagList)) {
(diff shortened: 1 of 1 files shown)


OpenDreamRuntime/DreamManager.cs
```diff
@@ -233,57 +233,69 @@ public sealed partial class DreamManager {
...
-                    _appearanceSystem ??= _entitySystemManager.GetEntitySystem<ServerAppearanceSystem>();
-                    return _appearanceSystem.TryGetAppearance((uint)refId, out IconAppearance? appearance)
-                        ? new DreamValue(appearance)
-                        : DreamValue.Null;
-                case RefType.Proc:
-                    return new(_objectTree.Procs[refId]);
-                default:
-                    throw new Exception($"Invalid reference type for ref {refString}");
+            // Search for an object with this ref as its tag
+            // Note that surrounding [] are stripped out at this point, this is intentional
```
wixoaGit, line 292:
```suggestion
            // Note that surrounding [] are stripped out at this point, this is intentional
            // Doing locate("[abc]") is the same as locate("abc")
```

PR #1404, opened 2023-08-07
Minimal `winget()` implementation

### LINKED ISSUE #192: `winget()` not implemented
http://www.byond.com/docs/ref/#/proc/winget

NTstation

wixoaGit:
Implements `winget()`. Currently only supports the `size` and `is-disabled` properties.

I didn't use RT's serialization manager to read the property because I want to move UI away from using it. It's not great for setting singular properties and reading properties that are constantly changing like `size` and `splitter`.

Also implemented the CHILD control's `splitter` property. This PR is enough to get the "Fit Viewport" verb working.

Fixes #192

### DIFF
--- a/OpenDream.sln.DotSettings
+++ b/OpenDream.sln.DotSettings
@@ -22,7 +22,6 @@
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=fcopy/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=flist/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=hwmode/@EntryIndexedValue">True</s:Boolean>
-	
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=isarea/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=isfile/@EntryIndexedValue">True</s:Boolean>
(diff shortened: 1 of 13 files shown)


amylizzle:
Easy one to add to this would be `hwmode` since for us it should always return true. Also it's used in goon.

wixoaGit:
> Easy one to add to this would be `hwmode` since for us it should always return true. Also it's used in goon.

Done

OpenDream.sln.DotSettings
```diff
@@ -21,6 +21,8 @@
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=deciseconds/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=fcopy/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=flist/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=hwmode/@EntryIndexedValue">True</s:Boolean>
+
```
wixoaGit, line 25:
```suggestion
	<s:Boolean x:Key="/Default/UserDictionary/Words/=hwmode/@EntryIndexedValue">True</s:Boolean>
```

OpenDreamClient/Interface/InterfaceElement.cs
```diff
@@ -26,6 +26,17 @@ public class InterfaceElement {
         UpdateElementDescriptor();
     }
 
+    /// <summary>
+    /// Attempt to get a DMF property
+    /// </summary>
+    /// <param name="property"></param>
+    /// <param name="value"></param>
+    /// <returns></returns>
```
wixoaGit, line 34:
```suggestion
```

PR #1408, opened 2023-08-09
Several interface fixes

wixoaGit:
Several interface fixes that get modern tgui-chat/tgui-panel working
* Try/Catch any exceptions in BeforeBrowseHandler because exceptions there freeze CEF
* Using `id` in a global winset will target a specific element
* Having only 1 pane in a CHILD control will have that pane take up the whole space (though the splitter can still be grabbed, this needs to be disabled later)
* Only care about visible controls when determining the available space for a control with size `0x0`

New state of [auxmos-removed BeeStation](https://github.com/itsmeowForks/BeeStation-Hornet/tree/db2b4b439a98af9b56cb75c9b8cce8f8493f2a7f):
image


itsmeow:
Super based, thank you

PR #1413, opened 2023-08-12
Alpha applying to overlays properly

PowerfulBacon:
Previously, unless you were using keep together or RESET_ALPHA, the overlay's alpha value was completely ignored. Now it will multiply the parent's alpha value by the provided alpha value. This is the behaviour that byond exhibits, as when I place an object sandwhiched between another object and its overlay, decreasing the alpha of the behind object will decrease the alpha of the overlay.

I noticed this was an issue when sent a screenshot of kilostation, as our decals on Beestation use an alpha of 50 which is set at the overlay object level rather than the sprite level.

**BYOND:**
<details>

image

image

image

</details>

**OpenDream:**
Without the alpha variable set:
image
image

WIth the alpha variable set (Alpha of 50 was used):
image

Without the overlay alpha variable, but with the parent alpha:
image

With the overlay alpha and parent alpha:
image

### DIFF
--- a/TestGame/renderer_tests.dm
+++ b/TestGame/renderer_tests.dm
@@ -73,7 +73,9 @@
 	Crossed(var/atom/movable/AM)
 		src.loc = AM
 		AM << "You picked up [src]"
-		AM.overlays += image(src.icon, AM.loc, src.icon_state)	
+		var/mutable_appearance/ma = new(src.icon, src.icon_state)
+		ma.alpha = 50
+		AM.overlays += ma
 
 //simple underlay
 /obj/plaque/simple_underlay_test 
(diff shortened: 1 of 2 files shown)


TestGame/renderer_tests.dm
```diff
@@ -73,7 +73,9 @@
 	Crossed(var/atom/movable/AM)
 		src.loc = AM
 		AM << "You picked up [src]"
-		AM.overlays += image(src.icon, AM.loc, src.icon_state)	
+		var/mutable_appearance/ma = new(src.icon, src.icon_state)
+		ma.alpha = 100
+		AM.overlays += ma
```
wixoaGit, line 78: You left some debug code here
```suggestion
		AM.overlays += image(src.icon, AM.loc, src.icon_state)	
```

wixoaGit (changes requested):

TestGame/renderer_tests.dm
PowerfulBacon: Thought I'd keep it in since cit was the renderer test but I can remove it.

wixoaGit (approved):

PR #1416, opened 2023-08-12
Adds compiler version metadata to the compiled json

### LINKED ISSUE #1323: Feature: Add compiler version metadata to the compiled json
Opcodes are volatile, especially at this alpha stage. Running an old version of a json with the current runtime can have unpredictable effects. You should ideally be either prevented or warned if you're about to do that.

ZeWaka:
Title, hashes the DreamProcOpcodes and puts it in a metadata header in the json.

Fixes #1323

### DIFF
--- a/OpenDreamRuntime/DreamManager.cs
+++ b/OpenDreamRuntime/DreamManager.cs
@@ -9,6 +9,7 @@ using OpenDreamRuntime.Rendering;
 using OpenDreamRuntime.Resources;
 using OpenDreamShared;
 using OpenDreamShared.Dream;
+using OpenDreamShared.Dream.Procs;
 using OpenDreamShared.Json;
 using Robust.Server;
 using Robust.Server.Player;
@@ -106,6 +107,10 @@ namespace OpenDreamRuntime {
             DreamCompiledJson? json = JsonSerializer.Deserialize<DreamCompiledJson>(jsonSource);
             if (json == null)
                 return false;
+
+            if (!json.Metadata.Version.Equals(OpcodeVerifier.GetOpcodesHash())) {
+                _sawmill.Warning("Compiler opcode version does not match the runtime version! This may cause issues!");
+            }
 
             if (json.Maps == null || json.Maps.Count == 0) throw new ArgumentException("No maps were given");
             if (json.Maps.Count > 1) {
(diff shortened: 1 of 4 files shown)


OpenDreamRuntime/DreamManager.cs
```diff
@@ -107,6 +108,10 @@ public sealed partial class DreamManager {
             if (json == null)
                 return false;
 
+            if (!json.Metadata.Version.Equals(OpcodeVerifier.GetOpcodesHash())) {
+                _sawmill.Warning("Compiler opcode version does not match the runtime version! This may cause issues!");
```
ZeWaka, line 112: I chose a warning because it /could/ be intentional? This could be upgraded to an error if wanted.

OpenDreamShared/Json/DreamCompiledJson.cs
```diff
@@ -1,15 +1,23 @@
...
+    public string? Interface { get; set; }
+    public DreamTypeJson[]? Types { get; set; }
+    public ProcDefinitionJson[]? Procs { get; set; }
+}
+
+public sealed class DreamCompiledJsonMetadata {
+    /// <summary>
+    ///  Hash of all the <c>DreamProcOpcode</c>s
+    /// </summary>
+    public string Version { get; set; }
```
ZeWaka, line 22: Could also change this to something like `OpcodeHash` - I figured Version as the field name was more clear.

wixoaGit:
This should also take the enum names into account, so it'll know if you swap two opcodes or something

OpenDreamRuntime/DreamManager.cs
wixoaGit: In what cases could running a json from an old bytecode version be intentional?

OpenDreamShared/Json/DreamCompiledJson.cs
wixoaGit: `Version` works well

OpenDreamRuntime/DreamManager.cs
ZeWaka: Spent the day thinking about it, couldn't really think of anything. If someone is really determined to try this though, they can just comment this out.

wixoaGit: An error log by itself doesn't stop the program anyways

wixoaGit (approved):

PR #1417, opened 2023-08-13
Add `none` color support

amylizzle:
https://www.byond.com/docs/ref/#/{skin}/param/background-color

`none` is a valid value for background-color, so I add a custom deserialiser which allows that since the default one throws an error on "none".


wixoaGit:
I think `none` means "default color" rather than "no color", but that is gonna have to wait until we remove use of the serializer entirely.

amylizzle:
It could return null instead?

wixoaGit (approved):

PR #1418, opened 2023-08-14
Handle invalid dir/frame index in `/icon.Insert()`

wixoaGit:
Using an invalid dir/frame should just not insert an icon state


OpenDreamShared/Resources/DMIParser.cs
```diff
@@ -144,11 +144,12 @@ public sealed class ParsedDMIState {
             }
 
             /// <summary>
-            /// Get this state's frames
+            /// Get this state's frames<br/>
```
wixoaGit, line 147:
```suggestion
            /// Get this state's frames
```

PR #1428, opened 2023-08-19
Adds a pragma for unsafe client accesses

ike709:
I didn't have time to revisit static typing so I did this instead.

Lints for `client.foo` and suggests `client?.foo`.

I consider this a stylistic pragma and a disabled-by-default pragma because it'll throw a lot of false positives since SS13 will nullcheck clients in various ways like `if(client && client.foo)` (even though `if(client?.foo)` is faster).

But with how often I've been seeing null client bugs in SS13 lately, I still consider this a useful lint.

### DIFF
--- a/DMCompiler/DM/Visitors/DMVisitorExpression.cs
+++ b/DMCompiler/DM/Visitors/DMVisitorExpression.cs
@@ -598,7 +598,6 @@ namespace DMCompiler.DM.Visitors {
                         _ => throw new InvalidOperationException(),
                     };
                 }
-
                 switch (operation.Kind) {
                     case DMASTDereference.OperationKind.Field:
                     case DMASTDereference.OperationKind.FieldSafe: {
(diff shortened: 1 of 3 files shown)


DMCompiler/DM/Visitors/DMVisitorExpression.cs
```diff
@@ -618,6 +617,10 @@ sealed class DMVisitorExpression : DMASTVisitor {
                                 operation.Identifier = field;
                                 operation.GlobalId = null;
                                 operation.Path = property.Type;
+                                if (operation.Kind == DMASTDereference.OperationKind.Field &&
```
ike709, line 620: Checking `Kind` again here is kind of yucky but I'm not sure how else to do it without duplicating a bunch of this code in the other switch case.

sjh00028 (approved):

PR #1435, opened 2023-08-27
Bump RT

ike709:
Bumps RT to latest. Main thing for us is that tile IDs are now an `int` instead of `ushort`.

Paradise doesn't explode:

image

### DIFF
--- a/OpenDreamClient/Interface/Controls/ScalingViewport.cs
+++ b/OpenDreamClient/Interface/Controls/ScalingViewport.cs
@@ -16,6 +16,7 @@ namespace OpenDreamClient.Interface.Controls;
 public sealed class ScalingViewport : Control, IViewportControl {
     [Dependency] private readonly IClyde _clyde = default!;
     [Dependency] private readonly IInputManager _inputManager = default!;
+    [Dependency] private readonly IEntityManager _entityManager = default!;
 
     // Internal viewport creation is deferred.
     private IClydeViewport? _viewport;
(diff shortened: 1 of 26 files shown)


OpenDreamClient/Interface/Controls/ScalingViewport.cs
```diff
@@ -222,6 +223,22 @@ public sealed class ScalingViewport : Control, IViewportControl {
         return _viewport!.LocalToWorld(matrix.Transform(coords));
     }
 
+    public MapCoordinates PixelToMap(Vector2 coords)
+    {
```
wixoaGit, line 227:
```suggestion
    public MapCoordinates PixelToMap(Vector2 coords) {
```

wixoaGit (changes requested):

OpenDreamRuntime/LocalHostConGroup.cs
```diff
@@ -42,6 +50,13 @@ public sealed class LocalHostConGroup : IConGroupControllerImplementation, IPost
             return Equals(addr, IPAddress.Loopback) || Equals(addr, IPAddress.IPv6Loopback);
         }
 
+        public bool CheckInvokable(CommandSpec command, ICommonSession? user, out IConError? error)
+        {
```
wixoaGit, line 54:
```suggestion
        public bool CheckInvokable(CommandSpec command, ICommonSession? user, out IConError? error) {
```

OpenDreamRuntime/LocalHostConGroup.cs
```diff
@@ -42,6 +50,13 @@ public sealed class LocalHostConGroup : IConGroupControllerImplementation, IPost
...
+            error = null;
+            if (user is null) return false;
```
wixoaGit, line 56: Looking at [SS14's implementation](https://github.com/space-wizards/space-station-14/blob/4fb46c1cb69efa6a46a8edeab47689b9b4c7ba77/Content.Server/Administration/Managers/AdminManager.cs#L454) shows that `user` being null implies it's the server console. The server console should always be able to execute a command.
```suggestion
            if (user is null) return true; // Server console
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #1440, opened 2023-08-29
Remove allocations in `DMIParser.GetExportedDirectionCount()`

wixoaGit:
This was allocating huge amounts on tg

Before:
image
image

After:
image
image

Spending 20 seconds on this is still crazy, this will need changing later.


ZeWaka:
what the fuck lol

PR #1441, opened 2023-08-30
Fix global init order

wixoaGit:
Two important changes to global var init order:

1. All global vars (except lists) are initialized in the order they appear in the code, except globals inside procs are all lumped together and start at the appearance of the first one.
2. Proc global vars that reference other global vars are moved to the end of init. I achieved this by making global vars not available on the first pass, which makes them handled by the late-var-def system.

### DIFF
--- a/DMCompiler/DM/Visitors/DMVisitorExpression.cs
+++ b/DMCompiler/DM/Visitors/DMVisitorExpression.cs
@@ -14,7 +14,7 @@ namespace DMCompiler.DM.Visitors {
             // Only global vars and procs available
             Static,
 
-            // Only constant global vars and global procs available
+            // Only global procs available
             FirstPassStatic
         }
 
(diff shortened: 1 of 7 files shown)


DMCompiler/DM/Visitors/DMVisitorExpression.cs
```diff
@@ -6,14 +6,24 @@
...
-    sealed class DMVisitorExpression : DMASTVisitor {
+    internal sealed class DMVisitorExpression : DMASTVisitor {
+        public enum ScopeMode {
+            // All in-scope procs and vars available
+            Normal,
+
+            // Only global vars and procs available
+            Static,
+
+            // Only constant global vars and global procs available
```
wixoaGit, line 17:
```suggestion
            // Only global procs available
```

PR #1443, opened 2023-09-01
Implement `/tmp` and `/const` var decorators on object vars for `issaved()`

### LINKED ISSUE #832: Add support for `tmp` and `const` vars to `DMOpcodeHandlers.IsSaved()`
See the TODO in `DMOpcodeHandlers.IsSaved()`

Best to wait until https://github.com/wixoaGit/OpenDream/pull/814 is merged first

### LINKED ISSUE #1318: `issaved()` doesn't return the correct value for `type`
Causes runtime error on goon with 
```
for(var/V in src.vars)
	if (issaved(src.vars[V]))
		copy.vars[V] = src.vars[V]
```
as it tries to set `copy.vars["type"]`, which it shouldn't

amylizzle:
Fixes #832 
Fixes #1318


Content.Tests/DMProject/Tests/Stdlib/issaved.dm
wixoaGit, line 1: This test doesn't perform any `ASSERT()` so it isn't actually testing anything. `issaved_vars_index.dm` already does what this intends to, so this can just be removed.

OpenDreamShared/Json/DreamObjectJson.cs
```diff
@@ -20,6 +20,8 @@ public sealed class DreamTypeJson {
         public List<int> Verbs { get; set; }
         public Dictionary<string, object> Variables { get; set; }
         public Dictionary<string, int> GlobalVariables { get; set; }
+        public HashSet<string> ConstVariables { get; set; }
+        public HashSet<string> TmpVariables { get; set; }
```
wixoaGit, line 24: Since [the compiler doesn't always set these](https://github.com/OpenDreamProject/OpenDream/pull/1443/files#diff-76e845bf829013a0d2cf4684050050d5ccf1eb3e24b46200944b951b233f2ca3R174-R181) they can be null
```suggestion
        public HashSet<string>? ConstVariables { get; set; }
        public HashSet<string>? TmpVariables { get; set; }
```

OpenDreamRuntime/Objects/DreamObject.cs
```diff
@@ -171,6 +171,8 @@ public class DreamObject {
         }
 
         protected virtual void SetVar(string varName, DreamValue value) {
+            if(ObjectDefinition.ConstVariables.Contains(varName))
+                throw new Exception($"Cannot set const var \"{varName}\"");
```
wixoaGit, line 175: Possible performance impact?

OpenDreamRuntime/Objects/DreamObjectTree.cs
```diff
@@ -364,6 +364,18 @@ public sealed class DreamObjectTree {
                     objectDefinition.GlobalVariables.Add(jsonGlobalVariable.Key, jsonGlobalVariable.Value);
                 }
             }
+
+            if (jsonObject.ConstVariables != null) {
+                foreach (string jsonConstVariable in jsonObject.ConstVariables) {
+                    objectDefinition.ConstVariables.Add(jsonConstVariable);
+                }
```
wixoaGit, line 371: You can make `DreamObjectDefinition.ConstVariables` not readonly and just do `objectDefinition.ConstVariables = jsonObject.ConstVariables` instead of this loop. Maybe also make it nullable and start the value off as null since a lot of objects aren't going to have const vars.

OpenDreamRuntime/Objects/DreamObjectTree.cs
```diff
@@ -364,6 +364,18 @@ public sealed class DreamObjectTree {
...
+            }
+
+            if(jsonObject.TmpVariables != null) {
+                foreach (string jsonTmpVariable in jsonObject.TmpVariables) {
+                    objectDefinition.TmpVariables.Add(jsonTmpVariable);
+                }
```
wixoaGit, line 377: Same as `ConstVariables` here.

wixoaGit (changes requested):

wixoaGit:
I profiled Paradise init with this and the HashSet check in `DreamObject.SetVar()` only took ~68ms.

wixoaGit (approved):

PR #1450, opened 2023-09-02
Implement `on-show` and `on-hide` for map controls, plus `winset` parsing fixes

amylizzle:
Implements the `on-show` and `on-hide` winset calls for the map control.

Fixes nested attribute parsing in winset, ie `thing.stuff.var=value` (previously we could only do `thing.stuff=value` and deeper references would fail to parse silently).

Slight buff to the robustness of winset parsing also, as it's no longer fragile about whitespace. Previously it would break if there was no space between `.winset` and the command.

Also refactors the DreamCommandSystem so all verbs run through the DreamInterfaceManager now


OpenDreamClient/Input/DreamCommandSystem.cs
wixoaGit, line 1: If running commands is going to be through `DreamInterfaceManager` now, keeping this system seems unnecessary.

OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -434,6 +435,40 @@ internal sealed class DreamInterfaceManager : IDreamInterfaceManager {
...
+                    winsetParams = winsetParams.Trim('\"'); //clip quotes
+
+                    WinSet(null, winsetParams);
+                    break;
+
+                default: {
+                    // Send the entire command to the server.
+                    // It has more info about argument types so it can parse it better than we can.
+                    //RaiseNetworkEvent(new CommandEvent(command));
+                    EntitySystem.Get<DreamCommandSystem>().SendServerCommand(command);
```
wixoaGit, line 466: This kinda breaks the whole point of moving this to `DreamInterfaceManager`. Moving it here so you can use it before entity systems load won't work if you still attempt to get the entity system anyways.

You'll have to replace the network event with a custom `NetMessage`.

OpenDreamClient/Interface/Descriptors/ControlDescriptors.cs
```diff
@@ -138,8 +138,14 @@ public sealed partial class ControlDescriptorInfo : ControlDescriptor {
...
 
 public sealed partial class ControlDescriptorMap : ControlDescriptor {
+
+    [DataField("on-show")]
+    public string? OnShowCommand;
+    [DataField("on-hide")]
+    public string? OnHideCommand;
     [DataField("zoom-mode")]
     public string ZoomMode = "normal";
+
```
wixoaGit, line 148: Unnecessary spaces
```suggestion
public sealed partial class ControlDescriptorMap : ControlDescriptor {
    [DataField("on-show")]
    public string? OnShowCommand;
    [DataField("on-hide")]
    public string? OnHideCommand;
    [DataField("zoom-mode")]
    public string ZoomMode = "normal";
```

OpenDreamClient/Interface/Controls/ControlMap.cs
```diff
@@ -54,4 +65,19 @@ public sealed class ControlMap : InterfaceControl {
...
+        if (controlDescriptor.OnShowCommand != null) {
+            _dreamInterfaceManager.RunCommand(controlDescriptor.OnShowCommand);
+        }
+    }
+
+    public void OnHideEvent() {
+        ControlDescriptorMap controlDescriptor = (ControlDescriptorMap)ControlDescriptor;
+        if (controlDescriptor.OnHideCommand != null) {
+            _dreamInterfaceManager.RunCommand(controlDescriptor.OnHideCommand);
+            //EntitySystem.Get<DreamCommandSystem>().RunCommand(controlDescriptor.OnHideCommand);
```
wixoaGit, line 80:
```suggestion
```

OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -434,6 +435,40 @@ internal sealed class DreamInterfaceManager : IDreamInterfaceManager {
...
+                    winsetParams = winsetParams.Trim(); //clip space
+                    winsetParams = winsetParams.Trim('\"'); //clip quotes
+
+                    WinSet(null, winsetParams);
+                    break;
+
+                default: {
+                    // Send the entire command to the server.
+                    // It has more info about argument types so it can parse it better than we can.
+                    //RaiseNetworkEvent(new CommandEvent(command));
```
wixoaGit, line 465:
```suggestion
```

wixoaGit (changes requested):

OpenDreamClient/Interface/InterfaceMacro.cs
```diff
@@ -293,14 +292,14 @@ public sealed class InterfaceMacro : InterfaceElement {
             return;
         }
 
-        if (_entitySystemManager.TryGetEntitySystem(out DreamCommandSystem? commandSystem)) {
+
             string? keyName = ParsedKeybind.KeyToKeyName(args.Key);
             if (keyName == null)
                 return;
             string command = Command.Replace("[[*]]", keyName);
```
wixoaGit, line 299: Dedent these lines now that there's no `if` surrounding them
```suggestion
        string? keyName = ParsedKeybind.KeyToKeyName(args.Key);
        if (keyName == null)
            return;
        string command = Command.Replace("[[*]]", keyName);
```

OpenDreamClient/Interface/InterfaceMacro.cs
```diff
@@ -293,14 +292,14 @@ public sealed class InterfaceMacro : InterfaceElement {
...
-            commandSystem.RunCommand(command);
+            _interfaceManager.RunCommand(command);
             // args.Handle() omitted on purpose, in BYOND both the "specific" keybind and the ANY keybind are triggered
```
wixoaGit, line 301:
```suggestion
        _interfaceManager.RunCommand(command);
        // args.Handle() omitted on purpose, in BYOND both the "specific" keybind and the ANY keybind are triggered
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #1460, opened 2023-09-10
Improved `range` support in `view()` procs

### LINKED ISSUE #1456: `view(null, thing)` throws an invalid arugment error
According to the dmref, if range isn't specified the value should be `world.view` (which defaults to 5). I assume this applies to `hearers` `oview` etc. as well

ike709:
Closes #1456 

Implements null arg support and non-default `world.view` size support.

### DIFF
--- a/OpenDreamRuntime/Procs/Native/DreamProcNativeHelpers.cs
+++ b/OpenDreamRuntime/Procs/Native/DreamProcNativeHelpers.cs
@@ -154,16 +154,11 @@ internal static partial class DreamProcNativeHelpers {
             } else if (arg.TryGetValueAsString(out var distString)) {
                 range = new ViewRange(distString);
             } else if (!arg.IsNull) { // null range arg is handled by DefaultView above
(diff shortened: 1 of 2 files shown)


OpenDreamRuntime/Procs/Native/DreamProcNativeHelpers.cs
```diff
@@ -136,29 +137,35 @@ internal static partial class DreamProcNativeHelpers {
...
             }
         }
 
         return (center, range);
     }
 
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    private static void ThrowBadViewArg(DreamValue arg) {
+        throw new Exception($"Invalid argument: {arg}");
+    }
```
wixoaGit, line 167: This seems unnecessary. `ResolveViewArguments()` isn't very hot code and the `foreach` is expected to only run 2 times at most.

wixoaGit (changes requested):

wixoaGit (approved):

PR #1462, opened 2023-09-13
`is-checked` and `group` for menus

amylizzle:
Handles checking and group checking for menu items.

Note we need to fix a bug in RT upstream, because right now closing a submenu doesn't close it's parent, and so updates aren't reflected in the menu until the menu is fully closed. I'll PR something upstream in a bit but also RT is making some major changes right now so that might be a bit of a thing.

I added property getting also, although property setting is not implemented. The only notable thing to do there would be to call `CreateMenu()` after one of the element descriptors was updated.


OpenDreamClient/Interface/InterfaceMenu.cs
```diff
@@ -106,16 +116,53 @@ public sealed class MenuElement : InterfaceElement {
...
                 Text = text
             };
 
-            //result.IsCheckable = MenuElementDescriptor.CanCheck;
-            if (!String.IsNullOrEmpty(Command))
-                menuButton.OnPressed += () => { EntitySystem.Get<DreamCommandSystem>().RunCommand(Command); };
+
+            menuButton.OnPressed += () => {
+                    if(MenuElementDescriptor.CanCheck)
+                        if(MenuElementDescriptor.Group != null)
```
wixoaGit, line 130: According to documentation, a blank group is also treated as no group.
```suggestion
                        if(!String.IsNullOrEmpty(MenuElementDescriptor.Group))
```

wixoaGit (changes requested):

PR #1478, opened 2023-10-04
fix #warn choking on apostrophe

### LINKED ISSUE #1439: `#warn` chokes on apostrophe
In goon:
```
#ifndef I_KNOW_WHAT_IM_DOING_PROBABLY
#warn Only using one z-level. This will fuck everything up. You're gonna have a bad time.
#endif
INCLUDE_MAP("../utilities/devtest.dmm")
#define MAP_MODE "standard"
```
the `#warn` consumes the newline and `#endif`, resulting in an unmatched if block error.

Result: 
```
(base) amy@amy-ThinkPad:~/OpenDream$ DMCompiler/bin/Release/net7.0/DMCompiler ~/goonstation/goonstation.dme --suppress-unimplemented
Compiling goonstation.dme on 514.1584
Warning OD0000 at <internal>: Unimplemented proc & var warnings are currently suppressed
Warning OD1201 at maps/config/gottagofast.dm:2:6: Only using one z-level. This will fuck everything up. You#endif
Error OD0010 at goonstation.dme:2200:7: Missing 1 #endif directive
Compilation failed with 1 errors and 2 warnings
Total time: 00:08
```

amylizzle:
`#warn` directives shouldn't try to process the tokens. Fixes #1439 and adds test

### DIFF
--- a/Content.Tests/DMProject/Tests/Preprocessor/Pragma/warning_quote.dm
+++ b/Content.Tests/DMProject/Tests/Preprocessor/Pragma/warning_quote.dm
@@ -1,0 +1,10 @@
+#define CONDITION_TRUE
+
+#ifdef CONDITION_TRUE
+#warn Oh no you're gonna have a bad time
+#endif
+
+/proc/RunTest()
+	return
+
+#warn end of file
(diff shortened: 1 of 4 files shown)


wixoaGit:
Might be better to instead have the lexer read the rest of the line and save it as the warning/error token's value. Something like this in the lexer:

```cs
case "warning": {
    StringBuilder message = new StringBuilder();

    while (GetCurrent() is not '\0' and not '\n') {
        message.Append(GetCurrent());
        Advance();
    }

    token = CreateToken(TokenType.DM_Preproc_Warning, "#warn", message.ToString());
    break;
}
```

and use `(string)token.Value!` in `HandleErrorOrWarningDirective()`

PR #1479, opened 2023-10-07
Fix image constructor argument handling

out-of-phaze:
Before, it would always set `icon_state` to the corresponding argument even if `icon_state` was set by `loc`. Now, it properly skips it.

### DIFF
--- a/OpenDreamRuntime/Objects/Types/DreamObjectImage.cs
+++ b/OpenDreamRuntime/Objects/Types/DreamObjectImage.cs
@@ -44,17 +44,21 @@ public sealed class DreamObjectImage : DreamObject {
                     .Warning($"Attempted to create an /image from {icon}. This is invalid and a default image was created instead.");
         }
 
-        int argIndex = 1;
+        int startIndex = 0;
         DreamValue loc = args.GetArgument(1);
-        if (loc.TryGetValueAsDreamObject(out _loc)) { // If it's not a DreamObject, it's actually icon_state and not loc
-            argIndex = 2;
+        if (!loc.TryGetValueAsDreamObject(out _loc)) { // If it's not a DreamObject, it's actually icon_state and not loc
+            startIndex = 1;
+            if (!loc.IsNull) {
+                AtomManager.SetAppearanceVar(Appearance, "icon_state", loc);
+            }
         }
 
-        foreach (string argName in IconCreationArgs) {
-            var arg = args.GetArgument(argIndex++);
+        for (int i = startIndex; i < IconCreationArgs.Length; i++) {
+            var arg = args.GetArgument(i + 2);
             if (arg.IsNull)
                 continue;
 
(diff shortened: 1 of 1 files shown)


OpenDreamRuntime/Objects/Types/DreamObjectImage.cs
```diff
@@ -44,17 +44,21 @@ public sealed class DreamObjectImage : DreamObject {
...
         }
 
-        int argIndex = 1;
+        int startIndex = 0;
         DreamValue loc = args.GetArgument(1);
-        if (loc.TryGetValueAsDreamObject(out _loc)) { // If it's not a DreamObject, it's actually icon_state and not loc
-            argIndex = 2;
+        if (!loc.TryGetValueAsDreamObject(out _loc)) { // If it's not a DreamObject, it's actually icon_state and not loc
+            startIndex = 1;
+            if (!loc.IsNull) {
```
wixoaGit, line 51: `loc.IsNull` will always be false here because `TryGetValueAsDreamObject()` will return true for a null object.

OpenDreamRuntime/Objects/Types/DreamObjectImage.cs
```diff
@@ -44,17 +44,21 @@ public sealed class DreamObjectImage : DreamObject {
                     .Warning($"Attempted to create an /image from {icon}. This is invalid and a default image was created instead.");
         }
 
-        int argIndex = 1;
+        int startIndex = 0;
         DreamValue loc = args.GetArgument(1);
-        if (loc.TryGetValueAsDreamObject(out _loc)) { // If it's not a DreamObject, it's actually icon_state and not loc
-            argIndex = 2;
+        if (!loc.TryGetValueAsDreamObject(out _loc)) { // If it's not a DreamObject, it's actually icon_state and not loc
+            startIndex = 1;
```
wixoaGit, line 50: This ends up skipping the 3rd argument

wixoaGit (changes requested):
Playing around with argument skipping in /image's constructor won't work.

When named arguments are used, the VM places the named arguments in their positional order. Once you're in the constructor you have no way of knowing if positional arguments were used and you have to skip an arg, or if named arguments were used and skipping an arg will result in incorrect values.

`image('icon.dmi', "state", FLOAT_LAYER, EAST)`
gives you the args `0='icon.dmi' 1="state" 2=FLOAT_LAYER 3=EAST 4..6=null`

`image('icon.dmi', "state", layer=FLOAT_LAYER, dir=EAST)`
gives you `0='icon.dmi' 1="state" 2=null 3=FLOAT_LAYER 4=EAST 5..6=null`

Both of them have a string in the second spot. However, the first one skips an arg but the second does not.

This will probably have to be special-cased in the VM. That or extra type-guessing that I don't think BYOND does.

wixoaGit:
Closing in favor of #1627

PR #1480, opened 2023-10-07
Stop image overlays with null/blank icon state from inheriting icons

out-of-phaze:
As far as I know, a blank image `new /image` added as an overlay shouldn't inherit the parent's icon. It should only do that if you add a raw icon state as an overlay, like `"foo"`.

### DIFF
--- a/OpenDreamRuntime/Objects/Types/DreamList.cs
+++ b/OpenDreamRuntime/Objects/Types/DreamList.cs
@@ -678,13 +678,13 @@ public sealed class DreamOverlaysList : DreamList {
             overlay = new IconAppearance() {
                 IconState = iconState
             };
+            overlay.Icon ??= defaultIcon;
         } else if (atomManager.TryCreateAppearanceFrom(value, out var overlayAppearance)) {
             overlay = overlayAppearance;
         } else {
             return null; // Not a valid overlay
         }
 
-        overlay.Icon ??= defaultIcon;
         return overlay;
     }
 }


wixoaGit:
I tried `usr.overlays |= new /image(null, "arrow")` in BYOND, and it did create an overlay with the `arrow` state. A null or empty icon_state did not use the default, however.

out-of-phaze:
> I tried `usr.overlays |= new /image(null, "arrow")` in BYOND, and it did create an overlay with the `arrow` state and usr's icon. A null or empty icon_state did not use the default, however.

Huh, I held off on touching this PR because I assumed mirroring this functionality would require changes, but... I think that's what it does in this PR as is. Good to know.

wixoaGit:
> > I tried `usr.overlays |= new /image(null, "arrow")` in BYOND, and it did create an overlay with the `arrow` state and usr's icon. A null or empty icon_state did not use the default, however.
> 
> Huh, I held off on touching this PR because I assumed mirroring this functionality would require changes, but... I think that's what it does in this PR as is. Good to know.

That is not what this PR does. Using a line like what I gave does not add a visible overlay.

wixoaGit:
I came back to this and was unable to replicate my claim that `usr.overlays |= new /image(null, "arrow")` inherits usr's icon. I don't know what I'm doing different or what's changed, but this PR's behavior appears to be correct. Sorry!

wixoaGit (approved):

PR #1483, opened 2023-10-11
More comprehensive addition operator testing

wixoaGit:
I added a little framework for testing an operation against many types of values. Currently only used by an addition test.

Fixed several issues that popped up in the addition test, including the fact that `/datum/proc` and `/datum/verb` compile as strings in BYOND for some reason.


Content.Tests/DMProject/Tests/Operators/Shared/operator_testing.dm
```diff
@@ -0,0 +1,37 @@
...
+			var/expected_result = expected[i++]
+			var/result
+			
+			try
+				result = call(operator_proc)(a, b)
+			catch
+				result = "Error"
+			
+			if (!(result ~= expected_result))
+				CRASH("Expected [json_encode(expected_result)] for [json_encode(a)] + [json_encode(b)], instead got [json_encode(result)]")
```
wixoaGit, line 37:
```suggestion
			if (result ~! expected_result)
				CRASH("Expected [json_encode(expected_result)] for [json_encode(a)] and [json_encode(b)], instead got [json_encode(result)]")
```

PR #1486, opened 2023-10-12
Allow []= list typecasting in proc arg definition

### LINKED ISSUE #1415: Error declaring a list-typed proc argument using []
When compiling /vg/station, the following:
```
/datum/n_Interpreter/proc/CallProc(name, params[]=null)`
```
results in the errors:
```
Error OD0000 at code\modules\scripting\Interpreter\Interaction.dm:133:55: error: params: missing comma ',' or right-paren ')'
Error OD0000 at code\modules\scripting\Interpreter\Interaction.dm:133:55: error: bad argument definition '[]='
```

If I understand correctly
`params[]=null` should be the same as `list/params=null`.

Hinaichigo:
This is a draft to try to fix #1415 so that you can do things like

`/mob/proc/have_fun(ways_to_have_fun[] = list("dancing", "jumping around", "sightseeing"))`

Also, if anyone could please let me know how to run unit tests locally, it'd be greatly appreciated.

Fixes #1415


Content.Tests/DMProject/Tests/Procs/doublebracketlistarg
wixoaGit, line 1: This needs a `.dm` file extension. It will not be recognized as a unit test otherwise.

wixoaGit (changes requested):

Hinaichigo:
Thanks. It seems like the unit test passed.

wixoaGit (approved):

PR #1491, opened 2023-10-15
Don't throw `WarningCode.EmptyBlock` on semicolon bodies

ZeWaka:
Essentially, throw `WarningCode.EmptyBlock` on
```cs
if(xyz)
	// :(
else
	...
```
but not 
```cs
if(xyz)
	; // <--
else
	...
```
This way, developers can insert a semicolon to denote that they intentionally want to write a statement a certain way.

----
this is pretty cursed code, but it's the only really viable way i could see how to do this

### DIFF
--- a/DMCompiler/Compiler/DM/DMAST.cs
+++ b/DMCompiler/Compiler/DM/DMAST.cs
@@ -618,6 +618,12 @@ namespace DMCompiler.Compiler.DM {
         /// </remarks>
         public readonly DMASTProcStatement[] SetStatements;
 
+        /// <summary>
+        /// This is used to determine if we should flag for <c>WarningCode.EmptyBlock</c>,
+        /// which we don't do if there was a manually entered semicolon within.
+        /// </summary>
+        public bool NotCompletelyEmpty;
+
         /// <summary> Initializes an empty block. </summary>
         public DMASTProcBlockInner(Location location) : base(location) {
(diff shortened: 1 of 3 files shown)


DMCompiler/Compiler/DM/DMAST.cs
```diff
@@ -620,6 +620,12 @@ public sealed class DMASTProcBlockInner : DMASTNode {
         /// </remarks>
         public readonly DMASTProcStatement[] SetStatements;
 
+        /// <summary>
+        /// This is used to determine if we should flag for <c>WarningCode.EmptyBlock</c>,
+        /// which we don't do if there was a manually entered semicolon within.
+        /// </summary>
+        public bool NotCompletelyEmpty;
```
ZeWaka, line 627: definitely open to a wording change on this

DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -2616,5 +2639,19 @@ public partial class DMParser : Parser<Token> {
 
             return hasDelimiter;
         }
+
+        private bool DelimiterHasSemicolon(ref bool hasSemicolon) {
```
ZeWaka, line 2643: also this

ZeWaka:
> wixoa: What if we instead had something in our AST to represent an "empty statement"?
note

wixoaGit:
Closing in favor of #1670

PR #1493, opened 2023-10-15
Handle lone-carriage-return line endings

### LINKED ISSUE #1492: Mixed line endings error in OD
READ UPDATE IN BELOW COMMENT

Consider the following .dme in CRLF
```dm
// Hand crafted DME, will not work if saved with DreamMaker

// BEGIN_INTERNALS
// END_INTERNALS

// BEGIN_FILE_DIR
#define FILE_DIR .
// END_FILE_DIR

// BEGIN_PREFERENCES
// END_PREFERENCES

// BEGIN_INCLUDE
#include "Config.dm"
#include "../test_prelude.dm"
#include "Test.dm"
// END_INCLUDE

```

Adding the following two lines (13 and 17) breaks OD (Notice no `\r`)

image

This works fine in BYOND.

wixoaGit:
Lines ending with just `\r` are treated as newlines.
Multiple consecutive `\r` characters are also handled now.

Also some general cleanup in the preprocessor, especially in the lexer which no longer inherits `TextLexer`

Fixes #1492

### DIFF
--- a/DMCompiler/Compiler/DM/DMParser.cs
+++ b/DMCompiler/Compiler/DM/DMParser.cs
@@ -325,7 +325,7 @@ namespace DMCompiler.Compiler.DM {
                     statement = new DMASTObjectDefinition(loc, _currentPath, null);
                 }
 
-                if (requireDelimiter && !PeekDelimiter() && Current().Type != TokenType.DM_Dedent && Current().Type != TokenType.DM_RightCurlyBracket) {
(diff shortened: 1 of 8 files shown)


Cyberboss:
Note the problem was lines actually being
```
ghjkjfdslg\r\n
asdf\r\r\n
fdsa\r\r\n
ghjkjfdslg\r\n
```

Not sure if this is BYOND-like behaviour

Cyberboss:
Related #1492

PR #1501, opened 2023-10-15
Fix OS selected topic port reading as 0 in DM code

Cyberboss:


Cyberboss:
This requires https://github.com/space-wizards/RobustToolbox/pull/4493

wixoaGit (approved):

PR #1515, opened 2023-10-20
Opcode Metadata

ike709:
Adds a custom attribute for specifying metadata on individual opcodes.

Used this to pretty much nuke `ShrinkStack()` and `GrowStack()`.

Please pay close attention to the attributes because this is a great time to audit MaxStack.


ike709:
It may be trivial to handle local non-static vars properly depending on how wixoa answers a question on Discord.

ike709:
Locals should now be handled.

ike709:
Nuked proc purity from this PR to do it differently in a future PR. I still think the metadata system is useful to keep.

DMCompiler/Bytecode/DreamProcOpcode.cs
```diff
@@ -402,3 +402,50 @@ public static class OpcodeVerifier {
...
+{
+    private static readonly OpcodeMetadata[] MetadataCache = new OpcodeMetadata[256];
+
+    static OpcodeMetadataCache()
+    {
+        foreach (DreamProcOpcode opcode in Enum.GetValues(typeof(DreamProcOpcode)))
+        {
+            var field = typeof(DreamProcOpcode).GetField(opcode.ToString());
+            var attribute = Attribute.GetCustomAttribute(field!, typeof(OpcodeMetadataAttribute));
+            var metadataAttribute = Unsafe.As<OpcodeMetadataAttribute>(attribute);
```
wixoaGit, line 442: Why the `Unsafe.As<>()`? a normal cast should work fine

DMCompiler/DM/DMProc.cs
```diff
@@ -111,6 +111,7 @@ private class DMProcScope {
         private readonly List<SourceInfoJson> _sourceInfo = new();
         private string? _lastSourceFile;
 
+
```
wixoaGit, line 114: Extra line
```suggestion
```

wixoaGit (changes requested):

DMCompiler/Bytecode/DreamProcOpcode.cs
ike709: Why not?

wixoaGit: It's generally good to stay away from unsafe things unless necessary.

While `attribute` will probably always be an `OpcodeMetadataAttribute` here, you're still foregoing any safety checks which could lead to really odd issues down the line.

ike709:
```suggestion
            var metadataAttribute = (OpcodeMetadataAttribute)attribute;
```

wixoaGit (approved):

PR #1523, opened 2023-10-23
Attempt to reconnect after a disconnect

### LINKED ISSUE #1520: Client reconnection should fully reload state like BYOND
The server dying and instantly coming back does not reset the client.

There should be a restart protocol if the server detects a new client connected to an old session.

wixoaGit:
Makes the SS14 launcher redial the server after a disconnect.

Closes #1520

### DIFF
--- a/OpenDreamClient/States/DreamUserInterfaceStateManager.cs
+++ b/OpenDreamClient/States/DreamUserInterfaceStateManager.cs
@@ -31,6 +31,8 @@ public sealed class DreamUserInterfaceStateManager {
                 // When we disconnect from the server:
                 case ClientRunLevel.Error:
                 case ClientRunLevel.Initialize when args.OldLevel >= ClientRunLevel.Connected:
+                    // TODO: Reconnect without returning to the launcher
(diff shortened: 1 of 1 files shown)


OpenDreamClient/States/DreamUserInterfaceStateManager.cs
```diff
@@ -31,8 +31,12 @@ public sealed class DreamUserInterfaceStateManager {
                 // When we disconnect from the server:
                 case ClientRunLevel.Error:
                 case ClientRunLevel.Initialize when args.OldLevel >= ClientRunLevel.Connected:
-                    if (_gameController.LaunchState.FromLauncher) {
-                        _stateManager.RequestStateChange<ConnectingState>();
+                    if (_gameController.LaunchState is {
```
wixoaGit, line 34:
```suggestion
                    // TODO: Reconnect without returning to the launcher
                    // The client currently believes its still connected at this point and will refuse
                    if (_gameController.LaunchState is {
```

PR #1535, opened 2023-11-18
Add support for tmp savefiles, fix multiple handles for savefiles, clean up

### LINKED ISSUE #1449: `new /datum/savefile(null)` throws an error
According to the DM ref: If no name is specified, a temporary file will be created, which will be destroyed when the savefile is no longer in use. If a resource cache is specified, a temporary file will be created and the contents of the cached file will be copied into it. Changes will therefore only be temporary. 

TL;DR it should create a temporary file which is deleted when the datum is.

amylizzle:
Adds support for temporary savefiles (fixes #1449)
Allows multiple savefile datums pointing at the same file to remain synchronised
Reduces memory usage by reducing duplication of savefiles and closing resources when they're no longer in use
Moves savefile flushing to be between ticks, except when `flush()` is called manually.

### DIFF
--- a/OpenDreamRuntime/EntryPoint.cs
+++ b/OpenDreamRuntime/EntryPoint.cs
@@ -72,7 +72,7 @@ namespace OpenDreamRuntime {
 
         protected override void Dispose(bool disposing) {
             // Write every savefile to disk
-            foreach (var savefile in DreamObjectSavefile.Savefiles) {
+            foreach (var savefile in DreamObjectSavefile.Savefiles.ToArray()) { //ToArray() to avoid modifying the collection while iterating over it
                 savefile.Close();
             }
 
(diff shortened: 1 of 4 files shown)


wixoaGit (changes requested):

OpenDreamRuntime/EntryPoint.cs
```diff
@@ -72,8 +72,8 @@ public sealed class EntryPoint : GameServer {
 
         protected override void Dispose(bool disposing) {
             // Write every savefile to disk
-            foreach (var savefile in DreamObjectSavefile.Savefiles) {
-                savefile.Flush();
+            foreach (var savefile in DreamObjectSavefile.Savefiles.ToArray()) { //ToArray() to avoid modifying the collection while iterating over it
```
wixoaGit, line 75: Copying the list seems unnecessary, but it's probably fine if it doesn't grow too big (especially once softdels are added)

wixoaGit (approved):

PR #1539, opened 2023-11-26
Fixes yield proc scheduling. Turns sleeps into an opcode

### LINKED ISSUE #1262: `set background = TRUE` procs are extremely slow
Using `set background = TRUE` in `setup_atmos_machinery()` on Nebula adds 151seconds to Ministation init.
```
/datum/controller/subsystem/machines/proc/setup_atmos_machinery(list/machines)
	set background=1

	report_progress("Initializing atmos machinery")
	for(var/obj/machinery/atmospherics/A in machines)
		A.atmos_init()
		CHECK_TICK

	report_progress("Initializing pipe networks")
	for(var/obj/machinery/atmospherics/machine in machines)
		machine.build_network()
		CHECK_TICK
```
169 second Ministation init with the above, *even if you replace the contents of the loops with a no-op.* Removing `set background = TRUE` from this one proc makes all of init take 28 seconds.

Cyberboss:
The behavior is now the same for either `spawn` or `sleep`:

- `delay < 0`: Proc will continue immediately if nothing is scheduled in the same tick. Otherwise, same as `delay == 0`.
- `delay == 0`: Proc scheduled on end of same tick.
- `delay > 0`: Proc scheduled after `n` ticks.

This also fixes the ProcScheduler being able to still have queued procs when running multiple tests in serial.

Then, to fix perf issues I had to redo sleeping into an opcode:

- Remove `sleep` from DMStandard.
- Two new opcodes `Sleep` and `BackgroundSleep` (which is just `sleep -1`, much faster than popping a `float` off for background sleeps).
- Create new lightweight async proc state for sleeping.
- Compiler/runtime adjustments to handle new opcodes.
- Added `IOpenDreamGameTiming` as a wrapper around `IGameTiming` that we can control in unit tests.

Closes #1262


Content.Tests/DMProject/Tests/Sleeping/YieldOrder.dm
```diff
@@ -0,0 +1,58 @@
...
+	world.log << "Inline:"
+	TestSequence(MODE_INLINE)
+
+	world.log << "Background:"
+	TestSequence(MODE_BACKGROUND)
+
+	rand_seed(22475)
+	for(var/i in 1 to 10000)
+		world.log << "Rand-[i]:"
+		TestSequence(MODE_RAND)
```
Cyberboss, line 58: Had to up the test timeout because this overran 500ms

Cyberboss:
Apparently this is still unperformant for the issue in question

DMCompiler/DM/DMProc.cs
```diff
@@ -459,20 +459,19 @@ private class DMProcScope {
...
-                    throw new CompileErrorException(Location, "Cannot do a background sleep without a sleep proc");
-                }
-
-                PushFloat(-1); // argument given to sleep()
-                Call(DMReference.CreateGlobalProc(sleepProc.Id), DMCallArgumentsType.FromStack, 1);
-                Pop(); // Pop the result of the sleep call
+        public void SleepDelayPushed() => WriteOpcode(DreamProcOpcode.Sleep);
+
+        public void Sleep(float delay) {
+            if(delay == -1.0f) // yielding
```
Cyberboss, line 465: Should be `<= 0`

Cyberboss:
Somethings still fucky, which is why the TGS tests are failing

wixoaGit:
This has been abandoned for a long while and a lot of it would probably need redone anyways, so I'm closing this.

PR #1544, opened 2023-12-04
Add compiler warning when enumerating the contents of an area

### LINKED ISSUE #1003: `in area` lint
Would be cool to lint over this for tg:

```dm
for (var/obj/thing in my_cool_area)
```

...as this performs a hidden `in world` loop.

For OpenDream this might not be relevant since I would hope it Does Not Do That but would be nice

distributivgesetz:
Fixes #1003

When enumerating over the contents of an area, BYOND  also enumerates over world contents apparently. That's not a thing here in OD land, but giving people a warning is nice I guess

### DIFF
--- a/DMCompiler/DM/Visitors/DMProcBuilder.cs
+++ b/DMCompiler/DM/Visitors/DMProcBuilder.cs
@@ -466,7 +466,7 @@ namespace DMCompiler.DM.Visitors {
                             var outputVar = DMExpression.Create(_dmObject, _proc, outputExpr);
                             var list = DMExpression.Create(_dmObject, _proc, exprIn.List);
 
-                            if (list is LValue {Path: { } path} && path.IsDescendantOf(DreamPath.Area)) {
+                            if (list is {Path: { } path} && path.IsDescendantOf(DreamPath.Area)) {
                                 DMCompiler.Emit(WarningCode.SuspiciousAreaContentsEnumeration, list.Location,
                                     "Enumerating over the contents of an area implicitly performs a world enumeration in BYOND");
                             }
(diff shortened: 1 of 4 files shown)


DMCompiler/DM/Visitors/DMProcBuilder.cs
```diff
@@ -466,6 +466,11 @@ sealed class DMProcBuilder {
                             var outputVar = DMExpression.Create(_dmObject, _proc, outputExpr);
                             var list = DMExpression.Create(_dmObject, _proc, exprIn.List);
 
+                            if (list is {Path: { } path} && path.IsDescendantOf(DreamPath.Area)) {
```
wixoaGit, line 469: Checking the path itself isn't quite correct, since an object's path and its inheritance aren't the same thing. For example:
```
/space
    parent_type = /area
```
The area's path is `/space` but it's still a type of area. You want to instead get the object from `DMObjectTree` and use `IsSubtypeOf()`
```suggestion
                            if (list is {Path: { } path} && DMObjectTree.GetDMObject(path, createIfNonexistent: false)?.IsSubtypeOf(DreamPath.Area) == true) {
```

OpenDreamShared/Compiler/CompilerError.cs
```diff
@@ -45,6 +45,7 @@ public enum WarningCode {
         PointlessBuiltinCall = 2206, // For pointless calls to issaved() or initial()
         SuspiciousMatrixCall = 2207, // Calling matrix() with seemingly the wrong arguments
         FallbackBuiltinArgument = 2208, // A builtin (sin(), cos(), etc) with an invalid/fallback argument
+        SuspiciousAreaContentsEnumeration = 2209, // See https://github.com/OpenDreamProject/OpenDream/issues/1003
```
wixoaGit, line 48: 3000-3999 may be a better place to put this, since this affects no compile-time behavior.

wixoaGit (changes requested):

distributivgesetz:
Now that I think about it, I don't think that a warning for a bug that we don't have is something that opendream really needs, and I wouldn't be upset if this PR got closed. The only reason I made this PR was to get comfortable with learning how this works anyways

wixoaGit:
I have to say, I agree. I don't see a reason for us to have a warning that doesn't apply to OD. This seems like a better fit for something like DreamChecker whose target is BYOND.

PR #1545, opened 2023-12-05
implement savefile << operator

### LINKED ISSUE #1306: `savefile << object` is not implemented
https://www.byond.com/docs/ref/#/savefile/operator/%3c%3c

The file of interest is https://github.com/OpenDreamProject/OpenDream/blob/1b80c46ef2ffcabcd4e6cea0cffe52ac65fbc0b1/OpenDreamRuntime/Objects/Types/DreamObjectSavefile.cs#L8-L9

Which would need this implemented:
https://github.com/OpenDreamProject/OpenDream/blob/1b80c46ef2ffcabcd4e6cea0cffe52ac65fbc0b1/OpenDreamRuntime/Objects/DreamObject.cs#L383-L387

LetterN:
resolves #1306 , required by #1546

### DIFF
--- a/OpenDreamRuntime/Objects/Types/DreamObjectSavefile.cs
+++ b/OpenDreamRuntime/Objects/Types/DreamObjectSavefile.cs
@@ -103,9 +103,8 @@ public sealed class DreamObjectSavefile : DreamObject {
 
         if (CurrentDir.TryGetValue(entryName, out DreamValue entry)) {
             return entry;
-        } else {
-            return DreamValue.Null;
         }
+        return DreamValue.Null;
     }
 
     public override void OperatorIndexAssign(DreamValue index, DreamValue value) {
@@ -121,5 +120,10 @@ public sealed class DreamObjectSavefile : DreamObject {
         if (!Directories.ContainsKey(_currentDirPath)) {
             Directories.Add(_currentDirPath, new SavefileDirectory());
         }
+    }
+
+    // << statement
+    public override void OperatorOutput(DreamValue value) {
+        CurrentDir[_currentDirPath] = value;
     }
 }


amylizzle:
This is already handled https://github.com/OpenDreamProject/OpenDream/blob/81a2a0470e07c90687f1dc1c81ea78b01a29d218/OpenDreamRuntime/Procs/DMOpcodeHandlers.cs#L2059

OpenDreamRuntime/Objects/Types/DreamObjectSavefile.cs
```diff
@@ -122,4 +121,9 @@ public sealed class SavefileDirectory : Dictionary<string, DreamValue> { }
             Directories.Add(_currentDirPath, new SavefileDirectory());
         }
     }
+
+    // << statement
+    public override void OperatorOutput(DreamValue value) {
+        CurrentDir[_currentDirPath] = value;
```
wixoaGit, line 127: In BYOND this appears to do more than just set the value. It looks like it turns it into a list of some sort if there's already a value there.

For example
```
var/savefile/S = new()
S << "a"
S << "b"
world.log << S.ExportText()
```
Gives `. = "a","b"`

wixoaGit (changes requested):

LetterN:
Merging this with #1546 instead

OpenDreamRuntime/Objects/Types/DreamObjectSavefile.cs
LetterN: Redirecting this to  OperatorIndexAssign or some fancy intermediate that sets the index to (A) value so that it's a list

PR #1546, opened 2023-12-05
[DNM] implement datum saving

### LINKED ISSUE #1303: `/datum.Write(savefile)` is not implemented
```dm
/datum/foo
	var/bar = 15

/datum/foo/Write(savefile/S)
	ASSERT(bar == 15)

/proc/RunTest()
	var/datum/foo/F = new()
	var/savefile/S = new("foo_bar.sav")
	S << F
```

Emits
```
 Error OD0405 at write_override.dm:5:7: Definition for proc Write on type /datum/foo not found
 Compilation failed with 1 errors and 0 warnings
 Total time: 00:00
```

https://www.byond.com/docs/ref/#/datum/proc/Write

### LINKED ISSUE #1305: `/datum.Read(savefile)` is not implemented
https://www.byond.com/docs/ref/#/datum/proc/Read

It is defined, but not implemented

### LINKED ISSUE #1306: `savefile << object` is not implemented

LetterN:
Title. implements how the Read() and Write() work

resolves #1303
resolves #1305
resolves #1306

requires #1535

currently nothing is being written on savefile(?) as asserts accessing fails `S["mapdata/.0/type"]`


OpenDreamRuntime/Procs/Native/DreamProcNativeDatum.cs
```diff
@@ -0,0 +1,45 @@
+using OpenDreamRuntime.Objects;
+using OpenDreamRuntime.Objects.Types;
+
+namespace OpenDreamRuntime.Procs.Native;
+
+internal static class DreamProcNativeDatum {
+    [DreamProc("Write")]
+    [DreamProcParameter("F", Type = DreamValue.DreamValueTypeFlag.DreamObject)]
+    public static DreamValue NativeProc_Write(NativeProc.Bundle bundle, DreamObject? src, DreamObject? usr) {
+        var probablySavefile = bundle.GetArgument(0, "F");
```
amylizzle, line 10: Insane though it seems, I think this should actually be done in the savefile's `OperatorIndexAssign` and the native `Write()` and `Read()` should be stubs. The behaviour is that the writing/reading happens regardless of whether the native `Write`/`Read` are called, and you cannot cancel it by, for example, returning early.

To be clear, you could do this:
```
/obj/test
  Write(savefile/F)
    if(some_condition)
      ..()
    else
      return null
```

and it would still do the write even if `some_condition` was true

OpenDreamRuntime/Procs/Native/DreamProcNativeDatum.cs
```diff
@@ -0,0 +1,43 @@
...
+        if (!probablySavefile.TryGetValueAsDreamObject(out DreamObjectSavefile? savefile) && savefile == null)
+            return DreamValue.Null; // error out bad path or something
+
+        foreach (var key in src!.GetVariableNames()) {
+            if (!src.IsSaved(key)) continue;
+            var result = src.GetVariable(key);
+
+            // skip if initial var is same
+            if (src.ObjectDefinition.TryGetVariable(key, out var val) && val == result) continue;
+            savefile.AddOrAssignValue(key, result);
```
amylizzle, line 20: Potential infinite loop here if you have something like `object.myvar = object` which byond handles with a special reference `object("..")`

LetterN: it pulls from the object definition there shouldn't be a loop there

PR #1547, opened 2023-12-06
Add --help for DMCompiler

### LINKED ISSUE #1229: Add a `--help` arg
Since OD isn't quite production-ready yet, it should probably just print something advising the user to ask in OD discord for assistance or to check the github wiki.

LetterN:
possibly resolves #1229 (?) Only adds --help on DMCompiler

### DIFF
--- a/DMCompiler/Program.cs
+++ b/DMCompiler/Program.cs
@@ -62,7 +62,7 @@ internal static class Program {
 
     private static void PrintHelp() {
         Console.WriteLine("DM Compiler for OpenDream");
-        Console.WriteLine("For more information please visit the (link here to be modified in the PR)");
+        Console.WriteLine("For more information please visit https://github.com/OpenDreamProject/OpenDream/wiki");
         Console.WriteLine("Usage: ./DMCompiler [options] [file].dme\n");
         Console.WriteLine("Options and arguments:");
         Console.WriteLine("--help                    : Show this help");
--- a/DMCompiler/Program.cs
+++ b/DMCompiler/Program.cs
@@ -66,11 +66,11 @@ internal static class Program {
         Console.WriteLine("Usage: ./DMCompiler [options] [file].dme\n");
         Console.WriteLine("Options and arguments:");
         Console.WriteLine("--help                    : Show this help");
-        Console.WriteLine("--version                 : Print the compiler version");
-        Console.WriteLine("--skip-bad-args           : ");
-        Console.WriteLine("--suppress-unimplemented  : Do not log about unimplemented functions in OpenDream");
-        Console.WriteLine("--dump-preprocessor       : ");
-        Console.WriteLine("--no-standard             : ");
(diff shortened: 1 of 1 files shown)


DMCompiler/Program.cs
```diff
@@ -4,163 +4,183 @@
...
-                            Console.WriteLine("Compiler arg 'version' requires a full BYOND build (e.g. --version=514.1584)");
-                            return false;
-                        }
+    private static void PrintHelp() {
+        Console.WriteLine("DM Compiler for OpenDream");
+        Console.WriteLine("For more information please visit https://github.com/OpenDreamProject/OpenDream/wiki");
+        Console.WriteLine("Usage: ./DMCompiler [options] [file].dme\n");
+        Console.WriteLine("Options and arguments:");
+        Console.WriteLine("--help                    : Show this help");
+        Console.WriteLine("--version                 : Print the compiler version");
```
wixoaGit, line 69: This does not print the compiler version. This arg is used to set the `DM_VERSION` and `DM_BUILD` macros. For example: `--version=514.1584`

DMCompiler/Program.cs
```diff
@@ -4,163 +4,183 @@
...
+        Console.WriteLine("--skip-bad-args           : ");
```
wixoaGit, line 70: By default the compiler will exit early if there's an argument it doesn't recognize. This disables that behavior.

DMCompiler/Program.cs
```diff
@@ -4,163 +4,183 @@
...
+        Console.WriteLine("--suppress-unimplemented  : Do not log about unimplemented functions in OpenDream");
```
wixoaGit, line 71: This includes unimplemented vars, and there's no need to specify it's in OpenDream.
```suggestion
        Console.WriteLine("--suppress-unimplemented  : Do not warn about unimplemented proc and var uses");
```

DMCompiler/Program.cs
```diff
@@ -4,163 +4,183 @@
...
+        Console.WriteLine("--dump-preprocessor       : ");
```
wixoaGit, line 72: This saves the result of preprocessing (`#include`, `#if`, defines, etc) in a file called `preprocessor_dump.dm` beside the given DME file.

DMCompiler/Program.cs
```diff
@@ -4,163 +4,183 @@
...
+        Console.WriteLine("--no-standard             : ");
```
wixoaGit, line 73: This disables objects and procs that are usually built-into every DM program by not including `DMStandard.dm`. This is only really useful if you're writing a new DM runtime outside of OpenDream, such as trying to [run DM in Roblox](https://github.com/OpenDreamProject/DMInRoblox).

wixoaGit (changes requested):

wixoaGit (approved):

PR #1550, opened 2023-12-12
Implement `::` operator

### LINKED ISSUE #854: `::` not implemented
Once this is done in the compiler, it might need to be worked on in increments, since it can do 4-5 different things.
> * ::varname and ::procname(...) can be used as a shorthand for global.varname and global.procname(...), respectively.
> * /thing::staticvar can be used to disambiguate var names when a static and local var have the same name.
> * Most commonly, A::B is a shorthand for initial(A.B), but can be used where A is a constant type path, e.g. /foo::bar.
> * In an object's var declarations, and static var declarations, type::varname and parent_type::varname are valid, so an expression like HP = parent_type::HP + 20 is allowed.
> * /mytype::procname() can provide a reference to a specific proc for use with call().

[https://www.byond.com/docs/ref/#/operator/::](https://www.byond.com/docs/ref/#/operator/::)

### LINKED ISSUE #1414: :: proc reference scope operator not being parsed correctly
Recently, /vg/station moved to being compatible with BYOND 515. Some changes had to be made to get it to compile, and as a result many statements were changed from this form:
		`owned_by.mob.unregister_event(/event/destroyed, src, .proc/qdel_self)`
 to
		`owned_by.mob.unregister_event(/event/destroyed, src, nameof(src::qdel_self()))`

These now result in the error:
`code\datums\chat_message.dm:64:68: Expected ')'`
when trying to compile in OD. 

Sub-issue of #854

distributivgesetz:
Closes #854 
Fixes #1414

Title says all.

I am still pretty new to this codebase, I'm learning as I'm going along so there's going to be a few mistakes. Please watch out for those, I will definitely get something wrong.

Features to implement:

- [x] Global shorthand
- [x] Static var disambiguation
- [x] Initial shorthand
- [x] type::variable and parent_type::variable initial shorthands
- [x] Proc reference


Hinaichigo:
fqscD3

Hinaichigo:
full send
do you know how to do the operator overloads?

distributivgesetz:
ye but I'm taking a break for personal life stuff

Hinaichigo:
This fixes a lot of errors in [fstation](https://gitgud.io/Hinaichigo/fstation13) and /vg/, however I'm not sure everything is working as intended:

```
/obj/item/clothing/suit/bomber_vest/proc/activate_vest()
	var/mob/living/carbon/human/H = loc
	if(!H)
		return
	if(!ishuman(H))
		return
	if(!(H.wear_suit == src))
		return
	active = 1
	canremove = 0
	H.register_event(/event/touched, src, nameof(src::on_touched()))
	H.register_event(/event/attacked_by, src, nameof(src::on_attacked_by()))
	H.register_event(/event/hitby, src, nameof(src::on_hitby()))
	H.register_event(/event/unarmed_attack, src, nameof(src::on_unarmed_attack()))
	H.register_event(/event/to_bump, src, nameof(src::on_to_bump()))
	H.register_event(/event/bumped, src, nameof(src::on_bumped()))

```
Errors:
`Error OD0000 at code/modules/clothing/suits/miscellaneous.dm:1029:5: nameof: requires a var, proc reference, or type path
`

wixoaGit:
> Unless I am way off this seems to re-introduce some errors around the order of const variable initialization:
> 
> ...
> 
> Errors: `Error OD0404 at code/modules/assembly/igniter.dm:10:9: var "wires" is not declared` Which seems related to #1011

This is because steps 5 and 6 were swapped around in `DMObjectBuilder.BuildObjectTree()`. Doing the steps in this order is incorrect, since now it tries to apply var overrides before all the vars have been defined.

Hinaichigo:
What was the reason why those steps had to be reordered to get :: to work in this case?

distributivgesetz:
> What was the reason why those steps had to be reordered to get :: to work in this case?

Variable definitions as well as overrides have to be parsed in order for `type::` and `parent_type::` to return the correct values

wixoaGit (approved):

PR #1552, opened 2023-12-14
Fix Topic calls dropping partial packets

Cyberboss:
This should fix the TGS test issues

### DIFF
--- a/OpenDreamRuntime/DreamManager.Connections.cs
+++ b/OpenDreamRuntime/DreamManager.Connections.cs
@@ -106,11 +106,18 @@ namespace OpenDreamRuntime {
                             var length = BitConverter.ToUInt16(buffer);
 
                             buffer = new byte[length];
-                            var read = await from.ReceiveAsync(buffer, cancellationToken);
-                            if (read != buffer.Length) {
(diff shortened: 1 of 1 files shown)


OpenDreamRuntime/DreamManager.Connections.cs
```diff
@@ -106,11 +106,16 @@ public sealed partial class DreamManager {
...
 
                             buffer = new byte[length];
-                            var read = await from.ReceiveAsync(buffer, cancellationToken);
-                            if (read != buffer.Length) {
-                                _sawmill.Warning("failed to parse byond topic due to insufficient data read");
-                                return null;
-                            }
+                            var totalRead = 0;
+                            do {
+                                var read = await from.ReceiveAsync(buffer, cancellationToken);
```
wixoaGit, line 111: Won't this write other parts of the packet to the beginning of `buffer`?

wixoaGit (changes requested):

OpenDreamRuntime/DreamManager.Connections.cs
Cyberboss: doy

wixoaGit (approved):

PR #1562, opened 2023-12-22
Update to RT v195.0.1 and .NET 8

wixoaGit:


global.json
```diff
@@ -1,6 +0,0 @@
-{
-  "sdk": {
-    "version": "7.0.0",
```
Cyberboss, line 3: Probably should keep this and just change it to 8

Cyberboss (changes requested):

ike709:
Audio will be broken by this PR until https://github.com/space-wizards/RobustToolbox/pull/4761

PR #1571, opened 2023-12-27
Update main menu

LetterN:
Update main menu + reshuffled Entrypoint

main menu (when launching on its own; not from launcher)
image

connecting menu
image

also soon :tm: other byond games that work on OD as background images (different on startup)


wixoaGit:
Gotta be honest I don't like the circularly moving background much

wixoaGit:
I don't like these changes to the main menu. It just moves things around, bolds some of the text, and adds a gray background that makes some things harder to read. I also think the background should be up to the codebase to set, and not be something included in OD's resources.

Any potential changes to the main menu I think should be put through some design with public input. The localization would be nice to have too, if we can get that in another PR.

PR #1580, opened 2023-12-31
Fix division type coalescing

ike709:
image


ike709:
nevermind I didn't finish reading chat history

PR #1592, opened 2023-12-31
Fix empty `for()` loop body error location

ike709:
This changes the error message location of an empty body to be the location of `for()`.

Daedalus has an empty for loop at the very end of the file, so the error message previously reported an entirely different file.

Also converts some stuff to use warningcodes. Let me know if this should be its own unique code, but I didn't think it was worth it.


DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -1192,25 +1192,28 @@ public partial class DMParser : Parser<Token> {
...
                 Consume(TokenType.DM_RightParenthesis, "Expected ')' in for after expression 3");
-                return new DMASTProcStatementFor(loc, expr1, expr2, expr3, dmTypes, GetForBody());
+                return new DMASTProcStatementFor(loc, expr1, expr2, expr3, dmTypes, GetForBody(loc));
             }
 
             return null;
 
-            DMASTProcBlockInner GetForBody() {
+            DMASTProcBlockInner GetForBody(Location forLocation) {
+
```
wixoaGit, line 1201:
```suggestion
```

DMCompiler/Compiler/DM/DMParserHelper.cs
```diff
@@ -21,6 +21,13 @@ public partial class DMParser {
             return level == ErrorLevel.Error;
         }
 
+        protected bool Error(WarningCode code, Location location, string message) {
+            ErrorLevel level = DMCompiler.CodeToLevel(code);
+            if (Emissions.Count < MAX_EMISSIONS_RECORDED)
+                Emissions.Add(new CompilerEmission(level, code, location, message));
+            return level == ErrorLevel.Error;
+        }
+
```
wixoaGit, line 30: Why use this over `DMCompiler.Emit()`?

The `Emissions` list is something we're trying to move away from.

wixoaGit (changes requested):

wixoaGit (approved):

PR #1595, opened 2023-12-31
Implement color inputting

Absolucy:
See: https://github.com/OpenDreamProject/OpenDream/issues/43

I have _no clue_ if this fully works or not. Database entries seem right, and the public OD paradise test server renders find, but when I test locally, in-game hair just refuses to render. Unsure if I'm stupid or if OD just hates my guts.

The color picker already mostly existed in Robust Toolbox... except some of the actual assets were in the SS14 codebase instead for some reason, _despite the fact that base RT references them._

this code prolly sucks ass, I haven't touched C# forever, it's like 6 AM, and I've been tweaking out trying to get this working since like midnight. feel free to uproot it or whatever, I just want to set Shion's hair color smh

## Preview

image


OpenDreamRuntime/DreamConnection.cs
```diff
@@ -209,7 +210,8 @@ public sealed class DreamConnection {
                 }
 
                 return DreamValue.Null;
-            } finally {
+            }
+            finally {
```
wixoaGit, line 214:
```suggestion
            } finally {
```

OpenDreamClient/Interface/Prompts/ColorPrompt.cs
```diff
@@ -0,0 +1,184 @@
...
+internal sealed class ColorPrompt : InputWindow {
+    private readonly BoxContainer _baseControl;
+    private readonly ColorSelectorSliders _colorSelector;
+    private readonly LineEdit _hexColor;
+    private readonly Button _preview;
+    private readonly bool _nullable;
+    private readonly Color _originalColor;
+
+    public ColorPrompt(string title, string message, string defaultValue, bool canCancel,
+        Action<DMValueType, object?>? onClose, bool alpha = false) : base(title, message, true, onClose) {
```
wixoaGit, line 21: Why is this passing `true` instead of `canCancel` to the base? Do color prompts always have a cancel button in byond?

wixoaGit (changes requested):

OpenDreamRuntime/DreamConnection.cs
Absolucy: blame vscode autoformatter

OpenDreamClient/Interface/Prompts/ColorPrompt.cs
Absolucy: Unsure. It was early morning, I might've not been thinking the best haha, but my logic was "if you cancel, it just returns the default color instead", bc I was scared of some edge-case I couldn't remember

wixoaGit: Color prompts do not always have a cancel button.

`input(usr) as color`:
image

`input(usr) as color|null`:
image

Absolucy: Alright, thanks.

wixoaGit (approved):

Issue #576, opened 2022-01-18
Add a warning for doing assignment in a conditional


ike709:
Example from coderbus:
```
var/thing="Thing"
if(thing="Thong")
    world.log<<thing
```
Output: `Thong`

This is a pretty common typo that the compiler could easily catch, but we might need to come up with a sane way to disable a warning in DM like C# pragmas.

PR #1596, opened 2023-12-31
Adds a warning when assigning a value in a conditional

Zonespace27:
When assigning a value in an `if()` or `switch()` statement, the compiler will emit an `AssignmentInConditional` warning.

Closes https://github.com/OpenDreamProject/OpenDream/issues/576

### DIFF
--- a/DMCompiler/Compiler/DM/DMParser.cs
+++ b/DMCompiler/Compiler/DM/DMParser.cs
@@ -1069,7 +1069,7 @@ namespace DMCompiler.Compiler.DM {
                     Error("Expected a condition");
                 }
                 if (condition is DMASTAssign) {
-                    DMCompiler.Emit(WarningCode.AssignmentInConditional, Current().Location, "Assignment in conditional");
+                    DMCompiler.Emit(WarningCode.AssignmentInConditional, condition.Location, "Assignment in conditional");
                 }
                 BracketWhitespace();
                 ConsumeRightParenthesis();
@@ -1381,9 +1381,6 @@ namespace DMCompiler.Compiler.DM {
                             DMCompiler.Emit(WarningCode.BadExpression, Current().Location, "Expected an expression");
 
                         break;
-                    }
-                    if (expression is DMASTAssign) {
-                        DMCompiler.Emit(WarningCode.AssignmentInConditional, Current().Location, "Assignment in conditional");
                     }
 
                     if (Check(TokenType.DM_To)) {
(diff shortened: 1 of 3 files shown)


DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -1377,6 +1382,9 @@ public partial class DMParser : Parser<Token> {
 
                         break;
                     }
+                    if (expression is DMASTAssign) {
+                        DMCompiler.Emit(WarningCode.AssignmentInConditional, Current().Location, "Assignment in conditional");
```
Zonespace27, line 1386: Not sure if this is necessary, OD500 already errors due to the `if()` in the switch being non-constant
image

ike709:
Wixoa may disagree but I think this should go with the 2000s warning codes since this emission can point out actual logic errors and isn't purely a stylistic choice.

wixoaGit:
> Wixoa may disagree but I think this should go with the 2000s warning codes since this emission can point out actual logic errors and isn't purely a stylistic choice.

So is `SuspiciousSwitchCase` but that's also a 3000.

Though I am struggling to come up with a distinction between these two ranges. The 2000s are "for compiler configuration of actual behaviour" but this behavior seems to just be error/not error? Not much different from any of the other configurable warning codes.

DMCompiler/Compiler/DM/DMParser.cs
wixoaGit: It's not necessary. That error is the more important thing in that case I'd say.

DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -1065,7 +1065,12 @@ public partial class DMParser : Parser<Token> {
                 Consume(TokenType.DM_LeftParenthesis, "Expected '('");
                 BracketWhitespace();
                 DMASTExpression? condition = Expression();
-                if (condition == null) Error("Expected a condition");
+                if (condition == null) {
+                    Error("Expected a condition");
+                }
+                if (condition is DMASTAssign) {
+                    DMCompiler.Emit(WarningCode.AssignmentInConditional, Current().Location, "Assignment in conditional");
```
wixoaGit, line 1072: Place the warning at the beginning of the expression rather than after.
```suggestion
                    DMCompiler.Emit(WarningCode.AssignmentInConditional, condition.Location, "Assignment in conditional");
```

wixoaGit (changes requested):

DMCompiler/Compiler/DM/DMParser.cs
Zonespace27: Removed

DMCompiler/Compiler/DM/DMParser.cs
Zonespace27: Resolved

wixoaGit (approved):

PR #1603, opened 2024-01-02
Refactor dereference operations into tagged unions

distributivgesetz:
It's just like rust enums no wayyyy

This PR tackles mainly nullability by refactoring the existing Dereference operation to be safer with the help of the type system. 

The previous implementation used a struct called `Operation`, which held an enum value for what kind of operation the struct described as well as metadata about the operation. This resulted in a struct where pretty much every additional metadata was optional and had to be asserted at the time of checking `Kind`, which wasn't very elegant.

I refactored dereference operations by binding the metadata into types with one shared superclass called `Operation`. 
The subclasses are there to describe all kinds of different metadata that they need. To get the metadata, you can use type patterns to extract the data. This completely eliminates the need to assert no nullability with `!`, making the code much clearer and more elegant. 

Since `Operation` had to inherit from other things now, I had to turn it into a class. This has zero impact on performance since all operations were located on the heap anyways.

Additionally I cleaned up some of the code involving dereferences. One less area with warnings.

### DIFF
--- a/DMCompiler/DM/Visitors/DMExpressionBuilder.cs
+++ b/DMCompiler/DM/Visitors/DMExpressionBuilder.cs
@@ -503,203 +503,172 @@ internal static class DMExpressionBuilder {
 
         // Special behaviour for `global.x`, `global.vars`, and `global.f()`
         if (expr is Global) {
-            ref DMASTDereference.Operation firstOperation = ref astOperations[0];
+            DMASTDereference.Operation firstOperation = astOperations[0];
 
-            if (firstOperation is { Kind: DMASTDereference.OperationKind.Field, Identifier.Identifier: "vars" }) {
-                // `global.vars`
-                expr = new GlobalVars(expr.Location);
+            switch (firstOperation)
+            {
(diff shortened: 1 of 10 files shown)


DMCompiler/DM/Visitors/DMExpressionBuilder.cs
```diff
@@ -554,156 +532,138 @@ public enum ScopeMode {
...
-                    pathIsFuzzy = true;
+                    prevPath = property?.Type;
+                    pathIsFuzzy = property == null;
                     break;
+                }
 
-                case DMASTDereference.OperationKind.Index:
-                case DMASTDereference.OperationKind.IndexSafe:
+                case DMASTDereference.IndexOperation indexOperation:
                     // Passing the path here is cursed, but one of the tests seems to suggest we want that?
```
wixoaGit, line 615: This comment was on the `DMExpression.Create()` call, move it above the `Index` assignment for clarity

wixoaGit (changes requested):

wixoaGit (approved):

PR #1607, opened 2024-01-04
Optimise ProcScheduler.Delays

amylizzle:
Was trying to figure out why goon was slow, came across non-optimal code. Was compelled.

Switches the `spawn()` and `sleep()` task scheduler to use a sorted list of tasks, and do binary insertion instead of unsorted adds.
This should change the time complexity as: best/avg/worst
`spawn()`/`sleep()` becomes O(1)/O(log2(n))/O(log2(n)) from O(1)/O(1)/O(1)
tick processing becomes O(1)/O(n_T)/O(n_T) from O(n)/O(n)/O(n) where n_T is the number of tasks that would resume this tick, and n is just the number of tasks queued. 
Best case is O(1) because we always gotta check the first task unless the task list is empty, but we gotta check that too so O(1)


amylizzle:
reopen with less fucked up git history

PR #1610, opened 2024-01-04
Optimise ProcScheduler.Delays

amylizzle:
Was trying to figure out why goon was slow, came across non-optimal code. Was compelled.

Switches the `spawn()` and `sleep()` task scheduler to use a priorityqueue instead of unsorted adds to a list.
This should change the time complexity as: best/avg/worst
`spawn()`/`sleep()` becomes `O(log2(n)`)/`O(log2(n))`/`O(log2(n))` from `O(1)`/`O(1)`/`O(1)`
tick processing becomes `O(1)`/`O((n_T)log2(n))`/`O(nlog2(n))` from `O(n)`/`O(n)`/`O(n)` where `n_T` is the number of tasks that would resume this tick, and `n` is just the number of tasks queued. 
Best case is `O(1)` because we always gotta check the first task unless the task list is empty, but we gotta check that too so O(1), and PQ.Peek() is `O(1)`. Dequeue is `O(log2(n))`

### DIFF
--- a/OpenDreamRuntime/Procs/ProcScheduler.Delays.cs
+++ b/OpenDreamRuntime/Procs/ProcScheduler.Delays.cs
@@ -1,5 +1,4 @@
 ﻿using System.Threading.Tasks;
-using Robust.Shared.Collections;
 using Robust.Shared.Timing;
 
 namespace OpenDreamRuntime.Procs;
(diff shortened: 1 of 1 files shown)


OpenDreamRuntime/Procs/ProcScheduler.Delays.cs
```diff
@@ -53,28 +52,33 @@ public sealed partial class ProcScheduler {
...
-            _tickers.RemoveSwap(i);
-            i -= 1;
+            _tickers.Dequeue();
         }
     }
 
     private sealed class DelayTicker {
         public readonly TaskCompletionSource TaskCompletionSource;
-        public int TicksLeft;
+        public uint TicksAt;
```
distributivgesetz, line 81:
```suggestion
        public required uint TicksAt;
```

PR #1611, opened 2024-01-06
Add `--debug` option to the packaging tool

wixoaGit:


Cyberboss (approved):

wixoaGit:
RobustToolbox does not compile with these compiler flags

wixoaGit:
> RobustToolbox does not compile with these compiler flags

No longer an issue after #1726

PR #1613, opened 2024-01-07
Enable server garbage collection

ike709:
According to https://stackoverflow.com/questions/44035282/how-can-i-enable-server-gc-in-net-core you may need to run `dotnet clean` first for it to take effect.


ike709:
Apparently already handled by opendreamserver.

PR #1618, opened 2024-01-10
Tweak `Nameof` slightly, add `GetNameof()` support to more expressions

distributivgesetz:
Throwing is bad. Unwinding isn't healthy for tracking code errors. I refactored `GetNameof()` to return an optional string, which gets checked in the actual `Nameof` expression. There we can emit the error. 

The arguments for `GetNameof` felt unnecessary since `dmObject` was unused, `proc` was used only once in `Local`, so I cut the arguments from the method. Let me know if I should keep it.

`nameof()` is also usable in constant contexts (eg. switch cases), so I added an override for `TryAsConstant()`.

Also a few l-values didn't have any nameof support, so I added it in this PR.

### DIFF
--- a/DMCompiler/DM/Expressions/Builtins.cs
+++ b/DMCompiler/DM/Expressions/Builtins.cs
@@ -603,8 +603,7 @@ namespace DMCompiler.DM.Expressions {
             if (_expr.GetNameof() is { } nameof) {
                 proc.PushString(nameof);
             } else {
-                DMCompiler.Emit(WarningCode.BadArgument, Location,
-                    "nameof() requires a var, proc reference, or type path");
+                DMCompiler.Emit(WarningCode.BadArgument, Location, "nameof() requires a var, proc reference, or type path");
                 proc.PushString("");
             }
         }
@@ -614,8 +613,7 @@ namespace DMCompiler.DM.Expressions {
                 constant = new String(Location, nameof);
                 return true;
             }
-            DMCompiler.Emit(WarningCode.BadArgument, Location,
-                "nameof() requires a var, proc reference, or type path");
+            DMCompiler.Emit(WarningCode.BadArgument, Location, "nameof() requires a var, proc reference, or type path");
             constant = null;
             return false;
         }
--- a/DMCompiler/DM/Expressions/Builtins.cs
+++ b/DMCompiler/DM/Expressions/Builtins.cs
@@ -1,10 +1,10 @@
-using OpenDreamShared.Compiler;
+using DMCompiler.Bytecode;
 using DMCompiler.Compiler.DM;
+using OpenDreamShared.Compiler;
 using OpenDreamShared.Dream;
 using OpenDreamShared.Json;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
-using DMCompiler.Bytecode;
 
 namespace DMCompiler.DM.Expressions {
     // "abc[d]"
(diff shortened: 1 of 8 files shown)


Hinaichigo:
When attempting to test this I get the following error:
`C:...OpenDream\DMCompiler\DM\Expressions\Dereference.cs(295,45): error CS0246: The type or namespace name 'NotNullWhen' could not be found (are you missing a using directive or an assembly reference?)`
I usually build OD using the `dotnet build` command, so maybe it's something that is normally included when you use an IDE like visual studio? Is there a workaround or some way to make it just werk with some kind of project setting?

DMCompiler/DM/Expressions/Builtins.cs
```diff
@@ -600,7 +600,22 @@ sealed class Nameof : DMExpression {
...
+            if (_expr.GetNameof() is { } nameof) {
+                proc.PushString(nameof);
+            } else {
+                DMCompiler.Emit(WarningCode.BadArgument, Location, "nameof() requires a var, proc reference, or type path");
+                proc.PushString("");
+            }
+        }
+
+        public override bool TryAsConstant([NotNullWhen(true)] out Constant? constant) {
+            if (_expr.GetNameof() is { } nameof) {
```
wixoaGit, line 612: Instead of copy/pasting this code, it would be better to perform this once in the constructor and save the result in a `private string _nameof`.

DMCompiler/DM/Expressions/Builtins.cs
```diff
@@ -644,6 +659,11 @@ public ProcOwnerType(Location location)
                 proc.PushType(dmObject.Id);
             }
         }
+
+        public override string GetNameof() {
+            DMCompiler.UnimplementedWarning(Location, "nameof(__TYPE__) is unimplemented");
+            return "";
+        }
```
wixoaGit, line 666: This would be possible if it still had the `dmObject` argument

DMCompiler/DM/Expressions/Builtins.cs
```diff
@@ -655,6 +675,12 @@ public ProcType(Location location)
         public override void EmitPushValue(DMObject dmObject, DMProc proc) {
             proc.PushProc(proc.Id);
         }
+
+        public override string GetNameof() {
+            // TODO: refactor proc references
+            DMCompiler.UnimplementedWarning(Location, "nameof(__PROC__) is unimplemented");
+            return "";
+        }
```
wixoaGit, line 683: This would be possible if it still had the `proc` argument

DMCompiler/DM/Expressions/Dereference.cs
```diff
@@ -241,7 +240,12 @@ public Dereference(Location location, DreamPath? path, DMExpression expression,
             proc.AddLabel(endLabel);
         }
 
-        public override bool TryAsConstant([NotNullWhen(true)] out Constant? constant) {
+        // BYOND says the nameof is invalid if the chain is not purely field operations
+        public override string? GetNameof() => _operations.All(op => op is FieldOperation)
+            ? ((FieldOperation)_operations[^1]).Identifier
+            : null;
+
+        public override bool TryAsConstant(out Constant constant) {
```
wixoaGit, line 248:
```suggestion
        public override bool TryAsConstant([NotNullWhen(true)] out Constant? constant) {
```

DMCompiler/DM/Expressions/LValue.cs
```diff
@@ -104,9 +105,7 @@ public Local(Location location, DMProc.LocalVariable localVar)
             EmitPushValue(dmObject, proc);
         }
 
-        public override string GetNameof(DMObject dmObject, DMProc proc) {
-            return LocalVar.IsParameter ? proc.Parameters[LocalVar.Id] : proc.GetLocalVarName(LocalVar.Id);
```
wixoaGit, line 108: This was the only use of `GetLocalVarName()`, that method should be removed

wixoaGit (changes requested):

DMCompiler/DM/Expressions/Builtins.cs
wixoaGit: Though feel free to ignore this if you're going to re-add the dmObject and proc args, since that wouldn't be possible.

distributivgesetz:
I'm going to move the eval of an expression name from the expression emission to the expression building phase, and if nothing depends on the existence of `Nameof` then I'm down to removing the class to return a constant string expression. `Nameof` a pretty bare class anyways and I don't feel like its existence justifies having to include context parameters into `TryAsConstant` (which is why I removed the parameters from `GetNameof`)

distributivgesetz:
> When attempting to test this I get the following error: `C:...OpenDream\DMCompiler\DM\Expressions\Dereference.cs(295,45): error CS0246: The type or namespace name 'NotNullWhen' could not be found (are you missing a using directive or an assembly reference?)`

Could be a problem with your RT submodule because the annotations package is referenced there. no idea otherwise.

wixoaGit (approved):

PR #1623, opened 2024-01-17
Fix compatibility mode

wixoaGit:
Fix the RT renderer's compatibility mode.

The shaders were updated to work in an older GLSL version. This meant giving values explicit high-precision and removing instances of implicit float/int conversion.


wixoaGit:
The client at least runs but clicking doesn't work because `GL.GetTextureImage()` is not supported in compatibility mode.

Issue #1077, opened 2023-03-09
Error using new with procs/verbs


Hinaichigo:
Attempting to compile /vg/station, the following errors occur:
```
Error OD0404 at code\modules\clothing\accessories\holster.dm:111:28: Type /obj/item/clothing/accessory/holster/verb/holster_verb does not exist
Error OD0000 at code\modules\clothing\accessories\holster.dm:107:5: Negative stack size
Error OD0404 at code\modules\clothing\accessories\holster.dm:116:28: Type /obj/item/clothing/accessory/holster/verb/holster_verb does not exist
Error OD0000 at code\modules\clothing\accessories\holster.dm:113:5: Negative stack size
Error OD0404 at code\modules\hydroponics\hydro_tray.dm:572:6: Type /atom/proc/remove_label does not exist
Error OD0404 at code\modules\paperwork\handlabeler.dm:143:6: Type /atom/proc/remove_label_verb does not exist
Error OD0000 at code\modules\paperwork\handlabeler.dm:138:6: Negative stack size
```
Relevant lines:

```
code\modules\clothing\accessories\holster.dm
107	/obj/item/clothing/accessory/holster/on_attached(obj/item/clothing/under/S)
108		..()
109		//We're making a new verb, see http://www.byond.com/forum/?post=238593
110		if(attached_to)
111			attached_to.verbs += new/obj/item/clothing/accessory/holster/verb/holster_verb(attached_to,holster_verb_name)
112
113	/obj/item/clothing/accessory/holster/on_removed(mob/user as mob)
114		//Yes, we're calling "new" when removing a verb. I blame verbs entirely for this shit. See: http://www.byond.com/forum/?post=80230
115		if(attached_to)
116			attached_to.verbs -= new/obj/item/clothing/accessory/holster/verb/holster_verb(attached_to,holster_verb_name)
117		..()
```

```
code\modules\hydroponics\hydro_tray.dm
570		labeled = copytext(n_label, 1, 32) //technically replaces any traditional hand labeler labels, but will anyone really complain?
571		update_name()
572		new/atom/proc/remove_label(src)
```

```
code\modules\paperwork\handlabeler.dm
138	/atom/proc/set_labeled(var/label, var/start_text = " (", var/end_text = ")")
139		if(labeled)
140			remove_label()
141		labeled = "[start_text][label][end_text]"
142		name = "[name][labeled]"
143		new/atom/proc/remove_label_verb(src)
```
[...]

PR #1633, opened 2024-01-23
Complete rework of verbs

wixoaGit:
A lot of preliminary work for more complete verb support (and some postliminary work as well).

A list of all verbs are now sent to the client, and it is up to the client to decide what verbs should be executable and displayed to the user. Using this I also added support for executing verbs by right-clicking a target and selecting a verb that applies to it. Right now "verb that applies to it" only means something in its `verbs` list because I haven't added support for `set src = ...`.

Things still left to do:

- `set src = ...`
- `in list()` arguments
- `default_verb_category`
- Other things probably

image
_Wow I can rotate a chair without pressing alt_

image
image
_Wow I can flip a table_

image
_Wow I can pick up a specific shard from this huge pile of shards_

_Wow I can close #245_


Hinaichigo:
Would allowing #1077 be in the scope of this?

wixoaGit:
> Would allowing #1077 be in the scope of this?

I'm not going to do anything further in this PR, it's fairly large as it is. I can do it in my next verb-related PR though which will be soon after.

OpenDreamClient/ClientVerbSystem.cs
```diff
@@ -0,0 +1,178 @@
...
+
+        _playerManager.LocalPlayerAttached += OnLocalPlayerAttached;
+
+        SubscribeNetworkEvent<AllVerbsEvent>(OnAllVerbsEvent);
+        SubscribeNetworkEvent<RegisterVerbEvent>(OnRegisterVerbEvent);
+        SubscribeNetworkEvent<UpdateClientVerbsEvent>(OnUpdateClientVerbsEvent);
+    }
+
+    public override void Shutdown() {
+        _verbs.Clear();
```
wixoaGit, line 37:
```suggestion
        _verbs.Clear();
        _clientVerbs = null;
```

PR #1634, opened 2024-01-23
Update an old ForcedError

ike709:
Made it a pragma.

### DIFF
--- a/DMCompiler/DM/DMProc.cs
+++ b/DMCompiler/DM/DMProc.cs
@@ -387,7 +387,7 @@ namespace DMCompiler.DM {
 
         public void CreateFilteredListEnumerator(DreamPath filterType) {
             if (!DMObjectTree.TryGetTypeId(filterType, out var filterTypeId)) {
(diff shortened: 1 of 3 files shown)


DMCompiler/DM/DMProc.cs
```diff
@@ -387,7 +387,7 @@ private class DMProcScope {
 
         public void CreateFilteredListEnumerator(DreamPath filterType) {
             if (!DMObjectTree.TryGetTypeId(filterType, out var filterTypeId)) {
-                DMCompiler.ForcedError($"Cannot filter enumeration by type {filterType}");
+                DMCompiler.Emit(WarningCode.ItemDoesntExist, Location, $"Cannot filter enumeration by type {filterType}");
```
wixoaGit, line 390: `Location` here refers to the location of the proc itself, and not the expression that called `CreateFilteredListEnumerator()`

wixoaGit (changes requested):

wixoaGit (approved):

PR #1641, opened 2024-01-28
Recursive constant list

### LINKED ISSUE #1445: Nested lists don't count as constant for global init order
As discussed on discord, global init for lists is not quite right still. 

Two issues:
- If the list is defined as 
```
var/global/list/thing = list(
    "key" = list(),
    "key2" = list(),
)
``` 
it does not get initialised because `list()` counts as non-constant.

- Minor order mismatch, noticed on goon. `var/global/list/material_cache = buildMaterialCache()` was firing out of order, with `/datum/mapPreLoad/New()` triggering before it. This was actually partially a bug in goon, but the out-of-order execution is still wrong for OD.

Update: It turns out the order mismatch wasn't real. Inferred new on lists is different from just `list()` init, and BYOND was failing silently. OD was failing loudly. Fixed in gooncode.

amylizzle:
Fixes #1445

### DIFF
--- a/DMCompiler/Compiler/DM/DMAST.cs
+++ b/DMCompiler/Compiler/DM/DMAST.cs
@@ -1337,8 +1337,8 @@ namespace DMCompiler.Compiler.DM {
                 ||
                 (value is {
                     Key: DMASTExpressionConstant,
-                    Value: DMASTList
-                } && ((DMASTList) value.Value).AllValuesConstant())
+                    Value: DMASTList valueList
+                } && valueList.AllValuesConstant())
             );
         }
     }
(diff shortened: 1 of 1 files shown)


DMCompiler/Compiler/DM/DMAST.cs
```diff
@@ -1329,10 +1329,17 @@ public sealed class DMASTList : DMASTExpression {
...
+            return Values.All(
+                value => (value is {
+                    Key: DMASTExpressionConstant,
+                    Value: DMASTExpressionConstant
+                })
+                ||
+                (value is {
+                    Key: DMASTExpressionConstant,
+                    Value: DMASTList
+                } && ((DMASTList) value.Value).AllValuesConstant())
```
wixoaGit, line 1341: Could be cleaned up a bit by removing the cast
```suggestion
                    Value: DMASTList valueList
                } && valueList.AllValuesConstant())
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #1642, opened 2024-01-28
Fix isnull

### LINKED ISSUE #1638: `isnull(deleted object)` returns `0`
It should return `1`

```
var/obj/O = new()
del(O)
world.log << "[json_encode(O)] [json_encode(isnull(O))]"
```

BYOND: `null 1`
OD: `null 0`

amylizzle:
Fixes #1638

### DIFF
--- a/OpenDreamRuntime/DreamValue.cs
+++ b/OpenDreamRuntime/DreamValue.cs
@@ -108,7 +108,7 @@ namespace OpenDreamRuntime {
 
         public bool IsNull {
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
-            get => Type == DreamValueType.DreamObject && (_refValue == null || ((DreamObject) _refValue).Deleted);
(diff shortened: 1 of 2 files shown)


OpenDreamRuntime/DreamValue.cs
```diff
@@ -108,7 +108,7 @@ public enum DreamValueTypeFlag {
 
         public bool IsNull {
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
-            get => Type == DreamValueType.DreamObject && _refValue == null;
+            get => Type == DreamValueType.DreamObject && (_refValue == null || ((DreamObject) _refValue).Deleted);
```
ike709, line 111: Since we're already checking Type you can probably use `Unsafe.As` to avoid a typecast on every `IsNull` call

wixoaGit (approved):

PR #1646, opened 2024-02-02
Fix min/max also test

amylizzle:


Content.Tests/DMProject/Tests/Math/minmax.dm
```diff
@@ -0,0 +1,16 @@
+/proc/RunTest()
+	ASSERT(min(1,2,3) == 1)
+	ASSERT(min(null, 1) == null)
+	ASSERT(min(null, -1) == -1)
+	ASSERT(min(1, null) == null)
+	ASSERT(min(0,null)==null)
```
wixoaGit, line 6: This test does not test `min(null, 0)`. In BYOND this gives `0` but this PR is giving `null`.
```suggestion
	ASSERT(min(0,null)==null)
	ASSERT(min(null, 0) == 0)
```

Content.Tests/DMProject/Tests/Math/minmax.dm
```diff
@@ -0,0 +1,16 @@
...
+	ASSERT(min("a","b","c")=="a")
+	ASSERT(min("b","a","c")=="a")
+
+	ASSERT(max(1,2,3) == 3)
+	ASSERT(max(null, 1) == 1)
+	ASSERT(max(null, -1) == null)
+	ASSERT(max(1, null) == 1)
+	ASSERT(max(0,null)==null)
```
wixoaGit, line 14: This does not test `max(null, 0)`. This PR gives `0` which is the same as BYOND.

wixoaGit (changes requested):

wixoaGit (approved):

PR #1648, opened 2024-02-02
Fix ascii2text encoding

amylizzle:
Allows use of UTF8 chars


Content.Tests/DMProject/Tests/Builtins/text2ascii_char.dm
```diff
@@ -1,2 +1,3 @@
 ﻿/proc/RunTest()
 	ASSERT(text2ascii_char("abcdéfg",6) == 102) //102 is f
+	ASSERT(ascii2text(text2ascii("a")) == "a")
```
wixoaGit, line 3: This file is for testing `text2ascii_char()`. Would be reasonable to rename this file to `text2ascii.dm` and use both forms in it though.

Would also be good to test some non-ASCII values.

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -136,7 +136,7 @@ internal static class DreamProcNativeRoot {
             if (!ascii.TryGetValueAsInteger(out int asciiValue))
                 throw new Exception($"{ascii} is not a number");
 
-            return new DreamValue(Convert.ToChar(asciiValue).ToString());
+            return new DreamValue(char.ConvertFromUtf32(asciiValue).ToString());
```
wixoaGit, line 139: The `ToString()` is unnecessary
```suggestion
            return new DreamValue(char.ConvertFromUtf32(asciiValue));
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #1650, opened 2024-02-02
Add security clearance for TGS test workflow on PRs

Cyberboss:
Adds a security check which requires the label `CI Cleared` to be added to guard against secret exposure.

TGS uses this same trick


wixoaGit:
I believe using the github token var has fixed this. We can revisit this PR if the problem occurs again.

PR #1652, opened 2024-02-03
Fix min icon size in GenerateDMI()

amylizzle:
Default value of 0,0 when the icon passed to `new /image()` is null throws an error.

### DIFF
--- a/OpenDreamRuntime/Objects/DreamIcon.cs
+++ b/OpenDreamRuntime/Objects/DreamIcon.cs
@@ -83,6 +83,9 @@ public sealed class DreamIcon {
         if (_cachedDMI != null)
             return _cachedDMI;
 
+        if(Width == 0 && Height == 0)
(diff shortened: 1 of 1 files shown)


OpenDreamRuntime/Objects/DreamIcon.cs
```diff
@@ -83,6 +83,9 @@ public sealed class IconFrame {
         if (_cachedDMI != null)
             return _cachedDMI;
 
+        if(Width == 0 && Height == 0)
+           Width = Height = 32; //TODO should be world.icon_size
```
wixoaGit, line 87: Should this instead set the size of `Width` or `Height` if either of them is 0, instead of only when both are?

wixoaGit (changes requested):

wixoaGit (approved):

PR #1659, opened 2024-02-05
Fix loading defaults from image object def

### LINKED ISSUE #1658: `/image` subclasses don't have vars set
```
/image/foreground_parallax_occlusion_overlay
	icon = 'icons/mob.dmi'
	icon_state = "mob"
	plane = PLANE_FOREGROUND_PARALLAX_OCCLUSION
```

`var/image/test = new /image/foreground_parallax_occlusion_overlay` should result in an image with `icon == 'icons/mob.dmi'` and `icon_state == "mob"` etc, but does not. Instead you get basically a `new /image` with no vars set

breaks parallax occlusion on goon, and probably a bunch of overlay bugs across SS13

here's a failing test about it:
```
/image/subclass
	plane = 123
	icon_state = "subclass"

/proc/RunTest()
	var/image/test = new /image/subclass
	ASSERT(test.plane == 123)
	ASSERT(test.icon_state == "subclass")
```

amylizzle:
Fixes #1658


OpenDreamRuntime/Objects/Types/DreamObjectImage.cs
```diff
@@ -37,14 +37,14 @@ public sealed class DreamObjectImage : DreamObject {
 
     public override void Initialize(DreamProcArguments args) {
         base.Initialize(args);
+        Appearance = AtomManager.GetAppearanceFromDefinition(ObjectDefinition);
```
wixoaGit, line 40: Would be good to avoid the `AtomManager.GetAppearanceFromDefinition()` if possible. Wouldn't we get the same behavior if the below `if` was kept the same but `Appearance = new IconAppearance()` was replaced with this?

wixoaGit (changes requested):

OpenDreamRuntime/Objects/Types/DreamObjectImage.cs
amylizzle: Any particular reason to avoid `AtomManager.GetAppearanceFromDefinition()`? It seems like the only sensible thing to do would be to duplicate its code anyway.

wixoaGit: I'm not saying to avoid it entirely, but only use it if there was no argument given to copy the appearance from.

Appearance creation is slow/heavy so it's good to avoid it if it's never used. Even if it's cached that's an extra unnecessary dictionary hit in what I imagine to be somewhat hot code.

DMCompiler/DMStandard/Types/Image.dm
```diff
@@ -7,8 +7,11 @@
 	var/appearance_flags = 0
 	var/blend_mode = 0
 	var/color = null
+	var/contents as opendream_unimplemented
```
wixoaGit, line 10:
```suggestion
	var/list/contents as opendream_unimplemented
```

DMCompiler/DMStandard/Types/Image.dm
```diff
@@ -39,6 +43,8 @@
 	var/text = "i" as opendream_unimplemented
 	var/matrix/transform
 	var/list/underlays = null
+	var/verbs as opendream_unimplemented
```
wixoaGit, line 46:
```suggestion
	var/list/verbs as opendream_unimplemented
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #1663, opened 2024-02-07
Add size labelling to PRs

amylizzle:
I always like this feature on repos, so we should have it too.
Gotta create the labels tho:
## Create the needed labels

Export both `GITHUB_TOKEN` and `REPO` (e.g. `my-username/my-repository`) and run the script below:

```bash
for size in XL XXL XS S M L; do
  curl -sf -H "Authorization: Bearer $GITHUB_TOKEN" "https://api.github.com/repos/kubernetes/kubernetes/labels/size/$size" |
    jq '. | { "name": .name, "color": .color, "description": .description }' |
    curl -sfXPOST -d @- -H "Authorization: Bearer $GITHUB_TOKEN" https://api.github.com/repos/$REPO/labels
done
```


wixoaGit:
Could we put this, #1664, and [conflict-labeler.yml](https://github.com/OpenDreamProject/OpenDream/blob/master/.github/workflows/conflict-labeler.yml) all in one workflow?

ZeWaka:
> Could we put this, #1664, and [conflict-labeler.yml](https://github.com/OpenDreamProject/OpenDream/blob/master/.github/workflows/conflict-labeler.yml) all in one workflow?

No, because #1664 isn't a workflow, it's a GitHub App

wixoaGit:
> > Could we put this, #1664, and [conflict-labeler.yml](https://github.com/OpenDreamProject/OpenDream/blob/master/.github/workflows/conflict-labeler.yml) all in one workflow?
> 
> No, because #1664 isn't a workflow, it's a GitHub App

What I was wondering is if we could put them in the same file.

If not (or it's unusual), I'd like conflict-labeler to have its name changed to something more specific than "Add Extra PR Labels"

amylizzle:
I don't actually think I can verify this in any way, but I've tried to combine size and merge conflict labelling. Just the what-files-were-touched one works differently.

.github/workflows/pr-labeler.yml
```diff
@@ -16,3 +16,8 @@ jobs:
           dirtyLabel: "Merge Conflict"
           repoToken: "${{ secrets.GITHUB_TOKEN }}"
           commentOnDirty: "This pull request has conflicts, please resolve those before we can evaluate the pull request."
+      - name: size-label
```
wixoaGit, line 19:
```suggestion
      - name: Apply Size Label
```

wixoaGit (approved):

PR #1666, opened 2024-02-09
Conditional labelling

amylizzle:
Should stop the size labelling job running on master I think

### DIFF
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -17,7 +17,7 @@ jobs:
           repoToken: "${{ secrets.GITHUB_TOKEN }}"
           commentOnDirty: "This pull request has conflicts, please resolve those before we can evaluate the pull request."
       - name: Apply Size Label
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-labeler.yml
```diff
@@ -17,6 +17,7 @@ jobs:
           repoToken: "${{ secrets.GITHUB_TOKEN }}"
           commentOnDirty: "This pull request has conflicts, please resolve those before we can evaluate the pull request."
       - name: Apply Size Label
+        if: ${{ ! github.ref == 'refs/heads/master' }}
```
wixoaGit, line 20: I'm unsure if the `!` applies before or after the `==`. Probably best to use `!=` instead.
```suggestion
        if: ${{ github.ref != 'refs/heads/master' }}
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #1670, opened 2024-02-12
Add "null statements" to the compiler

wixoaGit:
Just a lone semicolon that represents a statement with nothing in it. Allows for suppressing EmptyBlock emissions.

```dm
/proc/main()
    if (TRUE) // Has no body, considered an empty block.
```
```dm
/proc/main()
    if (TRUE)
        ; // A null statement. Not considered an empty block.
```


ZeWaka:
Thanks! Sorry for not getting to this.

PR #1684, opened 2024-02-24
Implement `JSON_PRETTY_PRINT`

### LINKED ISSUE #1662: `JSON_PRETTY_PRINT` is not supported
As of v515 `json_encode()` takes a `flags` argument. It currently seems to only take one flag, `JSON_PRETTY_PRINT`

https://www.byond.com/docs/ref/#/proc/json_encode
> The JSON_PRETTY_PRINT flag introduces spacing into the output for improved readibility. Spaces will be added after colons and commas. Non-empty arrays and object literals will have a line break and tabs before each item, and a line break with one fewer tab before the closing bracket or brace. (The special formatting for numbers like Infinity, or the list form for matrices, will not be given tabs and line breaks.)

ike709:
Closes #1662 

Checks for the exact define value since BYOND seems to ignore nonsense values.


ike709:
Unsure why tests are failing when they work fine on my machine.

Content.Tests/DMProject/Tests/Builtins/JsonEncode.dm
```diff
@@ -13,3 +13,5 @@
 	ASSERT(json_encode(list("A" = 3, "B" = 5)) == @'{"A":3,"B":5}')
 	ASSERT(json_encode(matrix(1,2,3,4,5,6)) == @'[1,2,3,4,5,6]')
 	ASSERT(json_encode(/datum/proc/foo) == "\"/datum/proc/foo\"")
+	ASSERT(length(json_encode(list("name"="fridge", "power"=12))) == 28)
```
ZeWaka, line 16: 28???????????

ZeWaka, line 16: crlf vs lf?

ZeWaka (commented):
X

Content.Tests/DMProject/Tests/Builtins/JsonEncode.dm
wixoaGit: This is the issue. `Utf8JsonWriter` uses `Environment.NewLine` which is `\n` on Linux and `\r\n` on Windows. Unfortunately I don't see any way to change this behavior, so we either have to do an inefficient text replace or decide not to care about the final length of `json_encode()`.

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -1175,11 +1175,24 @@ internal static class DreamProcNativeRoot {
...
+            if (bundle.GetArgument(1, "flags").TryGetValueAsInteger(out var prettyPrint) && prettyPrint == 1) {
+                options = new JsonWriterOptions {
+                    Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping, // "\"" instead of "\u0022"
+                    Indented = true
+                };
+            } else {
+                options = new JsonWriterOptions {
+                    Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping, // "\"" instead of "\u0022"
+                };
+            }
```
wixoaGit, line 1193: A little nicer, in my opinion
```suggestion
            // 515 JSON_PRETTY_PRINT flag
            bundle.GetArgument(1, "flags").TryGetValueAsInteger(out var prettyPrint);
            
            JsonWriterOptions options = new JsonWriterOptions {
                Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping, // "\"" instead of "\u0022"
                Indented = (prettyPrint == 1)
            };
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #1686, opened 2024-02-28
Fix out-of-order overrides

wixoaGit:
#1550 introduced a bug that broke overriding variables that depended on out-of-order vars.


Content.Tests/DMProject/Tests/Const/Const_LateDefinition.dm
```diff
@@ -8,6 +8,9 @@ var/const/late_defined_const = 1
 	var/a = 1 + const_var
 	var/const/const_var = 2
 
+/obj/TestObj/SubType
+	a = 4
```
ike709, line 12:
```suggestion
	a = 4 // Catches a regression introduced in #1550
```

PR #1695, opened 2024-03-02
Update the compiler test workflow's repositories

wixoaGit:
I removed modified /tg/station (which hasn't been relevant for a while now) and replaced it with a fixed commit of /tg/station and Goonstation. This should help ensure we don't break anything in those codebases that currently works.


wixoaGit:
TG broke in the 24 hours since I fixed it. Will update once that's fixed.

PR #1696, opened 2024-03-02
Parse proc return types

ike709:
Doesn't do anything with the info currently.

### DIFF
--- a/DMCompiler/Compiler/DM/DMParser.cs
+++ b/DMCompiler/Compiler/DM/DMParser.cs
@@ -237,7 +237,7 @@ namespace DMCompiler.Compiler.DM {
 
                     // Proc return type
                     // TODO: Currently we parse it but don't do anything with this information
-                    AsTypes(out _);
+                    AsTypes(out _, true);
 
                     DMASTProcBlockInner? procBlock = ProcBlock();
                     if (procBlock is null) {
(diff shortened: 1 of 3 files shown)


DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -2614,27 +2619,39 @@ public partial class DMParser : Parser<Token> {
...
-                        case "sound": type |= DMValueType.Sound; break;
-                        case "icon": type |= DMValueType.Icon; break;
-                        case "opendream_unimplemented": type |= DMValueType.Unimplemented; break;
-                        case "opendream_compiletimereadonly": type |= DMValueType.CompiletimeReadonly; break;
-                        default: Error("Invalid value type '" + typeToken.Text + "'"); break;
-                    }
 
+                    if (!Check(new TokenType[] { TokenType.DM_Identifier, TokenType.DM_Null })) {
+                        // Proc return types
+                        if (allowPath) {
```
wixoaGit, line 2625: It would be better for error recovery if this always tried to parse a path. You can accept or deny the path based on `allowPath` afterwards.

Right now `var/a = input("A") as /list` gives multiple errors:
```
Error OD0001 at code.dm:183:30: Expected value type
Error OD0000 at code.dm:184:2: Unknown identifier "list"
```

wixoaGit (changes requested):

DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -2614,27 +2619,39 @@ public partial class DMParser : Parser<Token> {
...
+                            path = Path();
```
ike709, line 2626:
```suggestion
                        path = Path();
                        if (allowPath) {
```

wixoaGit (approved):

PR #1702, opened 2024-03-03
winset ternaries and embedded wingets

amylizzle:
Implements the `thing.attribute=value?thing.attribute=newvalue:thing.attribute=othervalue` syntax. 
Implements `thing.attribute=[[whatever.attribute as escaped]]`

oh also I made it so you don't need to explicitly implement `TryGetProperty` for every property, they are now backed by the data field annotations.

### DIFF
--- a/OpenDreamClient/Interface/DMF/DMFParser.cs
+++ b/OpenDreamClient/Interface/DMF/DMFParser.cs
@@ -228,7 +228,6 @@ public sealed class DMFParser(DMFLexer lexer, ISerializationManager serializatio
                 else
                     Error($"Invalid attribute value ({valueText})");
             else if (Check(TokenType.Ternary)) {
-                condition = new DMFWinSet(element, attributeToken.Text, valueText);
                 List<DMFWinSet> ifValues = new();
                 List<DMFWinSet> elseValues = new();
                 while(TryGetAttribute(out var ifValue)){
--- a/OpenDreamClient/Interface/DMF/DMFParser.cs
+++ b/OpenDreamClient/Interface/DMF/DMFParser.cs
@@ -27,8 +27,8 @@ public sealed class DMFParser(DMFLexer lexer, ISerializationManager serializatio
     public List<DMFWinSet> GlobalWinSet() {
         List<DMFWinSet> winSets = new();
 
-        while (TryGetAttribute(out var element, out var key, out var attribute)) {
-            winSets.Add(new(element, key, attribute));
+        while (TryGetAttribute(out var winset)) {
+            winSets.Add(winset);
         }
 
         return winSets;
@@ -185,10 +185,13 @@ public sealed class DMFParser(DMFLexer lexer, ISerializationManager serializatio
         return false;
     }
 
-    private bool TryGetAttribute(out string? element, [NotNullWhen(true)] out string? key, [NotNullWhen(true)] out string? token) {
-        element = null;
-        key = null;
-        token = null;
+    private bool TryGetAttribute([NotNullWhen(true)] out DMFWinSet? winSet) {
+        string? element = null;
+        string? key = null;
+        string? token = null;
+        winSet = null;
+        DMFWinSet? condition = null;
+        DMFWinSet? elseValue = null;
         Token attributeToken = Current();
 
         if (Check(_attributeTokenTypes)) {
@@ -225,10 +228,19 @@ public sealed class DMFParser(DMFLexer lexer, ISerializationManager serializatio
                     valueText = "";
                 else
                     Error($"Invalid attribute value ({valueText})");
+            else if (Check(TokenType.Ternary)) {
+                condition = new DMFWinSet(element, attributeToken.Text, valueText);
+                if (!(TryGetAttribute(out winSet) && Check(TokenType.Colon)) || !TryGetAttribute(out elseValue)) {
+                    Error("Badly formatted ternary!");
+                    return false;
+                }
+                winSet.Condition = condition;
+                winSet.ElseValue = elseValue;
+                return true;
+            }
 
             Newline();
-            key = attributeToken.Text;
-            token = valueText;
+            winSet = new DMFWinSet(element, attributeToken.Text, valueText);
             return true;
         }
 
@@ -238,16 +250,17 @@ public sealed class DMFParser(DMFLexer lexer, ISerializationManager serializatio
     public MappingDataNode Attributes() {
         var node = new MappingDataNode();
 
-        while (TryGetAttribute(out var element, out var key, out var value)) {
-            if (element != null) {
-                Error($"Element id \"{element}\" is not valid here");
+        while (TryGetAttribute(out var winset)) {
+            if (winset.Element != null) {
+                Error($"Element id \"{winset.Element}\" is not valid here");
                 continue;
             }
 
-            if (value == "none")
+            //TODO implement the conditional check
+            if (winset.Value == "none")
                 continue;
 
-            node.Add(key, value);
+            node.Add(winset.Attribute, winset.Value);
         }
 
         return node;
--- a/OpenDreamClient/Interface/DMF/DMFParser.cs
+++ b/OpenDreamClient/Interface/DMF/DMFParser.cs
@@ -225,17 +225,17 @@ public sealed class DMFParser(DMFLexer lexer, ISerializationManager serializatio
                 else
                     Error($"Invalid attribute value ({valueText})");
             else if (Check(TokenType.Ternary)) {
-                List<DMFWinSet> ifValues = new();
-                List<DMFWinSet> elseValues = new();
-                while(TryGetAttribute(out var ifValue)){
-                    ifValues.Add(ifValue);
+                List<DMFWinSet> trueStatements = new();
+                List<DMFWinSet> falseStatements = new();
+                while(TryGetAttribute(out var statement)){
+                    trueStatements.Add(statement);
                 }
                 if(Check(TokenType.Colon)){ //not all ternarys have an else
-                    while(TryGetAttribute(out var elseValue)) {
-                        elseValues.Add(elseValue);
+                    while(TryGetAttribute(out var statement)){
+                        falseStatements.Add(statement);
                     }
                 }
-                winSet = new DMFWinSet(element, attributeToken.Text, valueText, ifValues, elseValues);
+                winSet = new DMFWinSet(element, attributeToken.Text, valueText, trueStatements, falseStatements);
                 return true;
             }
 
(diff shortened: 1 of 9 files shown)


OpenDreamClient/Interface/DMF/DMFParser.cs
```diff
@@ -3,6 +3,7 @@
 using OpenDreamClient.Interface.Descriptors;
 using Robust.Shared.Serialization.Manager;
 using Robust.Shared.Serialization.Markdown.Mapping;
+using YamlDotNet.Core.Tokens;
```
amylizzle, line 6:
```suggestion
```

OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -23,6 +23,7 @@
 using Robust.Shared.Utility;
 using SixLabors.ImageSharp;
 using System.Linq;
+using System.Text.RegularExpressions;
```
amylizzle, line 26:
```suggestion
```

OpenDreamClient/Interface/DMF/DMFParser.cs
```diff
@@ -185,10 +185,11 @@ public sealed class DMFParser(DMFLexer lexer, ISerializationManager serializatio
...
     }
 
-    private bool TryGetAttribute(out string? element, [NotNullWhen(true)] out string? key, [NotNullWhen(true)] out string? token) {
-        element = null;
-        key = null;
-        token = null;
+    private bool TryGetAttribute([NotNullWhen(true)] out DMFWinSet? winSet) {
+        string? element = null;
+        winSet = null;
+        DMFWinSet? condition;
```
wixoaGit, line 191: This isn't used

OpenDreamClient/Interface/DMF/DMFParser.cs
```diff
@@ -185,10 +185,11 @@ public sealed class DMFParser(DMFLexer lexer, ISerializationManager serializatio
...
+        DMFWinSet? elseValue;
```
wixoaGit, line 192: This can be inlined as `var out elseValue` on line 236

OpenDreamClient/Interface/DMF/DMFParser.cs
```diff
@@ -221,14 +222,27 @@ public sealed class DMFParser(DMFLexer lexer, ISerializationManager serializatio
...
                     Error($"Invalid attribute value ({valueText})");
             } else if (!Check(TokenType.Value))
-                if(Check(TokenType.Semicolon)) //thing.attribute=; means thing.attribute=empty string
+                if(Check(TokenType.Semicolon) || Check(TokenType.EndOfFile)) //thing.attribute=; means thing.attribute=empty string
                     valueText = "";
                 else
                     Error($"Invalid attribute value ({valueText})");
+            else if (Check(TokenType.Ternary)) {
+                List<DMFWinSet> ifValues = new();
+                List<DMFWinSet> elseValues = new();
```
wixoaGit, line 231: In BYOND are there allowed to be multiple ifValues or more than 2 elseValues?

OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -484,8 +484,15 @@ internal sealed class DreamInterfaceManager : IDreamInterfaceManager {
     }
 
     public void WinSet(string? controlId, string winsetParams) {
-        DMFLexer lexer = new DMFLexer(winsetParams);
-        DMFParser parser = new DMFParser(lexer, _serializationManager);
+        DMFLexer lexer;
+        DMFParser parser;
+        try{
+            lexer = new DMFLexer(winsetParams);
```
wixoaGit, line 490:
```suggestion
        DMFParser parser;
        try{
            var lexer = new DMFLexer(winsetParams);
```

OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -498,34 +505,92 @@ internal sealed class DreamInterfaceManager : IDreamInterfaceManager {
             return true;
         }
 
+        string HandleEmbeddedWinget(string? controlId, string value) {
+            string result = value;
+            int startPos = result.IndexOf("[[");
```
wixoaGit, line 510: IndexOf is culture sensitive. Probably not a big deal with `[[` and `]]` but still good practice to specify how to compare the strings.
```suggestion
            int startPos = result.IndexOf("[[", StringComparison.Ordinal);
```

OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -498,34 +505,92 @@ internal sealed class DreamInterfaceManager : IDreamInterfaceManager {
...
+            while(startPos > -1){
+                int endPos = result.IndexOf("]]", startPos);
```
wixoaGit, line 512:
```suggestion
                int endPos = result.IndexOf("]]", startPos, StringComparison.Ordinal);
```

OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -498,34 +505,92 @@ internal sealed class DreamInterfaceManager : IDreamInterfaceManager {
...
+                string inner = result.Substring(startPos+2, endPos-startPos-2);
+                string[] elementSplit = inner.Split('.');
+                string innerControlId = controlId ?? "";
+                if(elementSplit.Length > 1){
+                    innerControlId = (string.IsNullOrEmpty(innerControlId) ? "" : innerControlId+".")+string.Join(".", elementSplit[..^1]);
+                    inner = elementSplit[^1];
+                }
+                string innerResult = WinGet(innerControlId, inner);
+                result = result.Substring(0, startPos) + innerResult + result.Substring(endPos+2);
+                startPos = result.IndexOf("[[");
```
wixoaGit, line 524:
```suggestion
                startPos = result.IndexOf("[[", StringComparison.Ordinal);
```

OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -498,34 +505,92 @@ internal sealed class DreamInterfaceManager : IDreamInterfaceManager {
...
 
             if (CheckParserErrors())
                 return;
 
             // id=abc overrides the elements of other winsets without an element
-            string? elementOverride = winSets.FirstOrNull(winSet => winSet.Element == null && winSet.Attribute == "id")?.Value;
+            string? elementOverride = null;
+            foreach(var winSet in winSets)
+                if(winSet.Attribute == "id" && winSet.Element == null)
+                    elementOverride = HandleEmbeddedWinget(controlId, winSet.Value);
```
wixoaGit, line 539: Why this change? Does having multiple `id=...` do anything?

OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -498,34 +505,92 @@ internal sealed class DreamInterfaceManager : IDreamInterfaceManager {
...
-                        element.PopulateElementDescriptor(node, _serializationManager);
+                    if(winSet.IfValues is not null) {
+                        InterfaceElement? conditionalElement = FindElementWithId(elementId);
+                        if(conditionalElement is null)
+                            _sawmill.Error($"Invalid element on ternary condition \"{elementId}\"");
+                        else
+                            if(conditionalElement.TryGetProperty(winSet.Attribute, out var conditionalCheckValue) && conditionalCheckValue.Equals(winSet.Value, StringComparison.InvariantCultureIgnoreCase)) {
+                                foreach(DMFWinSet IfValue in winSet.IfValues) {
+                                    string? IfElementId = IfValue.Element ?? elementId;
+                                    InterfaceElement? IfElement = FindElementWithId(IfElementId);
```
wixoaGit, line 559: These should be renamed to `ifValue`/`ifElementId`/`ifElement`

OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -498,34 +505,92 @@ internal sealed class DreamInterfaceManager : IDreamInterfaceManager {
...
+                                        };
+                                        IfElement.PopulateElementDescriptor(node, _serializationManager);
+                                    } else {
+                                        _sawmill.Error($"Invalid element on ternary \"{IfElementId}\"");
+                                    }
+                                }
+                            } else if (winSet.ElseValues is not null){
+                                foreach(DMFWinSet ElseValue in winSet.ElseValues) {
+                                    string? ElseElementId = ElseValue.Element ?? elementId;
+                                    InterfaceElement? ElseElement = FindElementWithId(ElseElementId);
```
wixoaGit, line 572: These should be renamed to `elseValue`/`elseElementId`/`elseElement`

wixoaGit (changes requested):

OpenDreamClient/Interface/DMF/DMFParser.cs
amylizzle: Yup, you can do `condition=value?thing1=1;thing2=2:thing1=0;thing2=0`

OpenDreamClient/Interface/DreamInterfaceManager.cs
amylizzle: Id doesn't need to be the first element for it to work, and naturally goon has several instances where it isn't

wixoaGit: `FirstOrNull()` doesn't return the first value, it returns the earliest value that matches the predicate

OpenDreamClient/Interface/DMF/DMFParser.cs
wixoaGit: That's still only 1 condition and 2 branches

amylizzle: Oh I see what you mean, sorry this var is poorly named. The ifValues are the statements evaluated when the condition is true.

OpenDreamClient/Interface/DreamInterfaceManager.cs
amylizzle: Oh there was also an issue with using FirstOrNull() when DMFWinset was no longer a struct? I don't super remember, I'll check it

amylizzle:
```
The type 'DMFWinSet' must be a non-nullable value type in order to use it as parameter 'TSource' in the generic type or method 'Extensions.FirstOrNull<TSource>(IEnumerable<TSource>, Func<TSource, bool>)'[CS0453](https://msdn.microsoft.com/query/roslyn.query?appId%3Droslyn%26k%3Dk%28CS0453%29)
```
Seems like you can't do `FirstOrNull()` on classes, only structs. I do not know why.

OpenDreamClient/Interface/DMF/DMFParser.cs
amylizzle: renamed to TrueStatements and FalseStatements, plus code doc for extra clarity

OpenDreamClient/Interface/DreamInterfaceManager.cs
wixoaGit: `FirstOrNull()` is an RT extension for use with value types. Looks like `FirstOrDefault()` would be what you want here.

wixoaGit (approved):

PR #1703, opened 2024-03-04
Remove description from rightclick menu

amylizzle:
Is not BYOND parity. Is annoying on goon.

### DIFF
--- a/OpenDreamClient/Input/ContextMenu/VerbMenuPopup.xaml.cs
+++ b/OpenDreamClient/Input/ContextMenu/VerbMenuPopup.xaml.cs
@@ -23,13 +23,7 @@ internal sealed partial class VerbMenuPopup : Popup {
 
         _verbSystem = verbSystem;
         _target = target;
-
-        if (entityMetaData != null && !string.IsNullOrEmpty(entityMetaData.EntityDescription)) {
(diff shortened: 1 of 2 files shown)


wixoaGit:
Would it be better to instead limit the width of the description text?

amylizzle:
I suppose you could, but I don't really know why it's still there since the things that need it have `examine` verbs now.

OpenDreamClient/Input/ContextMenu/VerbMenuPopup.xaml.cs
```diff
@@ -23,13 +23,7 @@ internal sealed partial class VerbMenuPopup : Popup {
...
         _verbSystem = verbSystem;
         _target = target;
-
-        if (entityMetaData != null && !string.IsNullOrEmpty(entityMetaData.EntityDescription)) {
-            DescLabel.Margin = new Thickness(4, 0, 4, 0);
-            DescLabel.Text = entityMetaData.EntityDescription;
-        } else {
-            Desc.Visible = false;
-        }
+        Desc.Visible = false;
```
wixoaGit, line 26: The Desc control should be removed from `VerbMenuPopup.xaml` rather than just hidden.

wixoaGit (changes requested):

wixoaGit (approved):

Issue #1708, opened 2024-03-06
Adding an empty string to a save file lacks parity with DM


Zonespace27:
```
var/savefile/S = new()
S["test"] = ""
world.log << S.ExportText("/")
```

Will print `test = null` in DM and `test = ""` in OD.

Reproduced [here](https://discord.com/channels/847318834952405042/887074737879138315/1214725069834821715).

PR #1709, opened 2024-03-06
Converts empty strings to `null` in savefile serialization

Zonespace27:
Fixes https://github.com/OpenDreamProject/OpenDream/issues/1708

### DIFF
--- a/OpenDreamRuntime/Objects/Types/DreamObjectSavefile.cs
+++ b/OpenDreamRuntime/Objects/Types/DreamObjectSavefile.cs
@@ -398,11 +398,8 @@ public sealed class DreamObjectSavefile : DreamObject {
     public SFDreamJsonValue SerializeDreamValue(DreamValue val, int objectCount = 0) {
         switch (val.Type) {
             case DreamValue.DreamValueType.String:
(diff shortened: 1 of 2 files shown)


OpenDreamRuntime/Objects/Types/DreamObjectSavefile.cs
```diff
@@ -398,6 +398,12 @@ public sealed class DreamObjectSavefile : DreamObject {
     public SFDreamJsonValue SerializeDreamValue(DreamValue val, int objectCount = 0) {
         switch (val.Type) {
             case DreamValue.DreamValueType.String:
+                if(val.TryGetValueAsString(out var valString)) {
+                    if(valString == "") {
+                        val = DreamValue.Null;
+                    }
+                }
```
amylizzle, line 405:
```suggestion
                if(val.TryGetValueAsString(out var valString) && string.IsNullOrEmpty(valString)) 
                        val = DreamValue.Null;
```

Zonespace27:
> Maybe add it to one of the existing tests as well?

Done

wixoaGit (approved):

PR #1716, opened 2024-03-10
`exception.name` can be a datum

### LINKED ISSUE #1715: `exception.name` should support being not-a-string
per the DM ref (emphasis mine): `name: A text string (such as an error message) **or other value** `

essentially you can do `throw EXCEPTION(new /datum)` and have the resulting exception's `name` prop be that datum.

Naturally goon uses this.

amylizzle:
Fixes #1715

### DIFF
--- a/OpenDreamRuntime/DreamManager.cs
+++ b/OpenDreamRuntime/DreamManager.cs
@@ -330,10 +330,13 @@ namespace OpenDreamRuntime {
 
             // Invoke world.Error()
             var obj =_objectTree.CreateObject<DreamObjectException>(_objectTree.Exception);
-            obj.SetVariable("name", new DreamValue(e.Message));
-            obj.SetVariable("desc", new DreamValue(msg));
-            obj.SetVariable("file", new DreamValue(file));
-            obj.SetVariable("line", new DreamValue(line));
+            if(e is DMThrowException throwException)
+                obj.Name = throwException.Value;
+            else
+                obj.Name = new DreamValue(e.Message);
+            obj.Desc =  new DreamValue(msg);
+            obj.Line = new DreamValue(line);
+            obj.File = new DreamValue(file);
 
             WorldInstance.SpawnProc("Error", usr: null, new DreamValue(obj));
         }
--- a/OpenDreamRuntime/DreamManager.cs
+++ b/OpenDreamRuntime/DreamManager.cs
@@ -330,10 +330,10 @@ namespace OpenDreamRuntime {
 
             // Invoke world.Error()
             var obj =_objectTree.CreateObject<DreamObjectException>(_objectTree.Exception);
-            obj.Name = e.Message;
-            obj.Description = msg;
-            obj.Line = line;
-            obj.File = file;
+            obj.SetVariable("name", new DreamValue(e.Message));
+            obj.SetVariable("desc", new DreamValue(msg));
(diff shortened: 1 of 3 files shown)


OpenDreamRuntime/DreamManager.cs
```diff
@@ -330,10 +330,10 @@ public sealed partial class DreamManager {
...
             // Invoke world.Error()
             var obj =_objectTree.CreateObject<DreamObjectException>(_objectTree.Exception);
-            obj.Name = e.Message;
-            obj.Description = msg;
-            obj.Line = line;
-            obj.File = file;
+            obj.SetVariable("name", new DreamValue(e.Message));
+            obj.SetVariable("desc", new DreamValue(msg));
+            obj.SetVariable("file", new DreamValue(file));
+            obj.SetVariable("line", new DreamValue(line));
```
wixoaGit, line 336: Why go through the indirection of `SetVariable()` when we can set it directly?

OpenDreamRuntime/Objects/Types/DreamObjectException.cs
```diff
@@ -1,17 +1,38 @@
...
+            case "file":
+                if(value.TryGetValueAsString(out var stringFile))
+                    File = stringFile;
+                break;
+            case "line":
+                if(value.TryGetValueAsInteger(out var intLine))
+                    Line = intLine;
+                break;
+        }
+        base.SetVar(varName, value);
```
wixoaGit, line 29: `base.SetVar()` should only be called in the default case of the above switch. This is because it will add any variables it doesn't recognize to the object's list of modified variables. This is unnecessary when it will never be used.

OpenDreamRuntime/Objects/Types/DreamObjectException.cs
```diff
@@ -1,17 +1,38 @@
...
     //TODO: Match the format of BYOND exceptions since SS13 does splittext and other things to extract data from exceptions
 
+    protected override void SetVar(string varName, DreamValue value) {
+        switch (varName) { //set the internal vars of the C# object, then pass through to base.setvar to handle DM side
+            case "name":
+                Name = value;
+                break;
+            case "desc":
+                if(value.TryGetValueAsString(out var stringDescription))
+                    Description = stringDescription;
```
wixoaGit, line 18: This doesn't have to be a string in BYOND. It will take any value, with DreamObject types (that aren't lists) being turned into their display name, `DreamObject.GetDisplayName()`.

Same with `file` and `line`

wixoaGit (changes requested):

OpenDreamRuntime/Objects/Types/DreamObjectException.cs
```diff
@@ -1,26 +1,47 @@
...
+                return;
+            case "desc":
+                Desc = value;
+                return;
+            case "file":
+                File = value;
+                return;
+            case "line":
+                Line = value;
+                return;
```
wixoaGit, line 24: This misses the part where DreamObjects that aren't lists will instead use that object's `GetDisplayName()`

wixoaGit (changes requested):

OpenDreamRuntime/Objects/Types/DreamObjectException.cs
wixoaGit: Ignore me, misunderstood the output of `json_encode()`

PR #1717, opened 2024-03-14
Implement `::` for procs

ike709:
Wixoa's testing indicates that, oddly enough, this only seems to work for `name`

Added a `PointlessScopeOperator` pragma for other uses. I'll update the wiki when this is merged.

### DIFF
--- a/Content.Tests/DMProject/Tests/Operators/scope_proc_name.dm
+++ b/Content.Tests/DMProject/Tests/Operators/scope_proc_name.dm
@@ -1,4 +1,5 @@
 /datum/proc/foo()
+	set name = "abc"
 	return
 
 /proc/RunTest()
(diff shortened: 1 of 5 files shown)


Content.Tests/DMProject/Tests/Operators/scope_proc_pointless.dm
```diff
@@ -0,0 +1,8 @@
+// COMPILE ERROR
+#pragma PointlessScopeOperator error
+
+/datum/proc/foo()
+	return
```
wixoaGit, line 5: To drive home the point that `::desc` doesn't work
```suggestion
/datum/proc/foo()
	set desc = "abc"
	return
```

Content.Tests/DMProject/Tests/Operators/scope_proc_name.dm
```diff
@@ -0,0 +1,5 @@
+/datum/proc/foo()
+	return
```
wixoaGit, line 2: To show that it's not getting the verb name
```suggestion
/datum/proc/foo()
	set name = "abc"
	return
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #1722, opened 2024-03-20
Improve rgb() argument handling, add HSV support

wixoaGit:
I converted `rgb()` to an opcode rather than a proc. Named arguments now only look at the first character rather than the whole thing, mimicking BYOND behavior. I also added support for h/s/v arguments as well as use of the `space` argument.


OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -1860,6 +1860,117 @@ private static void HandleSuffixPronoun(ref StringBuilder formattedString, ReadO
...
+                    byte r = (byte)Math.Clamp(color1Value, 0, 255);
+                    byte g = (byte)Math.Clamp(color2Value, 0, 255);
+                    byte b = (byte)Math.Clamp(color3Value, 0, 255);
+
+                    color = new Color(r, g, b, aValue);
+                    break;
+                }
+                case ColorHelpers.ColorSpace.HSV: {
+                    // TODO: Going beyond the max defined in the docs returns a different value. Don't know why.
+                    float h = Math.Clamp(color1Value, 0, 360) / 360f;
```
amylizzle, line 1953: is it maybe just wrapping? HSV is basically polar co-ordinates. Try mod?

wixoaGit: That's not it. s=101 and s=1 give two different results:
```dm
world.log << rgb(h=180, s=101, v=50)
world.log << rgb(h=180, s=1, v=50)
```
```
#008080
#7e8080
```

PR #1753, opened 2024-04-16
Use master SS13 branches in CI

ike709:
Now that these codebases use OD in their CI workflows, we should test against their latest master on our end to minimize breaking changes (or to find them so we can adequately communicate them).

### DIFF
--- a/.github/workflows/compiler-test.yml
+++ b/.github/workflows/compiler-test.yml
@@ -51,7 +51,7 @@ jobs:
     - name: Compile Goonstation Master
       run: |
         New-Item goon\+secret\__secret.dme -type file
-        main\bin\DMCompiler\DMCompiler.exe goon\goonstation.dme --version=514.1584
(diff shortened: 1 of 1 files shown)


ZeWaka:
goon stay winning

.github/workflows/compiler-test.yml
```diff
@@ -34,29 +34,29 @@ jobs:
...
       with:
         repository: goonstation/goonstation
-        ref: 8c8b5276f231ae42b3b2390d837cab68af138fa6
+        ref: master
         path: goon
-    - name: Compile Goonstation 8c8b527
+    - name: Compile Goonstation Master
       run: |
         New-Item goon\+secret\__secret.dme -type file
         main\bin\DMCompiler\DMCompiler.exe goon\goonstation.dme --version=514.1584
```
ike709, line 54:
```suggestion
        main\bin\DMCompiler\DMCompiler.exe goon\goonstation.dme
```

wixoaGit (approved):

PR #1758, opened 2024-04-21
Remove CompileErrorException

wixoaGit:
CompileErrorException has been removed, opting to use alternatives over throwing an exception for DM code errors. Since it was marked obsolete this reduces the C# warning count from 179 to 117.

There's a new DMExpression called BadExpression that should be used anytime there was an error while building a DMExpression. The Null DMExpression was often used for this, but now we have something that can differentiate `null` from an error.

DMExpression code was also cleaned up. It now uses file-scoped namespaces, primary constructors, and are all marked `internal`.


DMCompiler/DM/Expressions/Builtins.cs
```diff
@@ -5,1052 +5,853 @@
...
+        }
+
+        DMCallArgumentsType argumentsType;
+        int stackSize;
+
+        switch (pathInfo.Value.Type) {
+            case ConstantPath.PathType.TypeReference:
+                var newProc = DMObjectTree.GetNewProc(pathInfo.Value.Id);
+
+                (argumentsType, stackSize) = arguments.EmitArguments(dmObject, proc, newProc);proc.PushType(pathInfo.Value.Id);
```
wixoaGit, line 78:
```suggestion
                (argumentsType, stackSize) = arguments.EmitArguments(dmObject, proc, newProc);
                proc.PushType(pathInfo.Value.Id);
```

PR #1771, opened 2024-05-06
Fix transforms not rendering

### LINKED ISSUE #1584: `/atom.transform` doesn't render
You click until horizontal but they don't go horizontal? 
It says X falls down, however their sprite remains upright

On SS14 paradise test

amylizzle:
Fixes #1584


wixoaGit:
Resting on Paradise places me a tile off, up or right depending on my direction
image
image

If possible, I'd prefer a fix in RT itself. Not sure how complicated the matrix passing-around is in Clyde though.

amylizzle:
That's odd, it works perfectly on goon. I'll poke at it. Also I think this is the proper fix, we need to translate to the 0 coordinate before rotation regardless of what we do.

Oh I bet something is doubling the width of the render target for paradise characters!

amylizzle:
That was exactly what it is. I'm now using the diff between the pos and render target size to calculate the origin, this seems much more reliable.

Tested on goon, paradise, and testgame

wixoaGit:
> I think this is the proper fix, we need to translate to the 0 coordinate before rotation regardless of what we do.

Maybe so with our current approach with `CreateRenderTargetFlipMatrix()`.

OpenDreamClient/Rendering/DreamViewOverlay.cs
```diff
@@ -415,12 +415,15 @@ internal sealed class DreamViewOverlay : Overlay {
...
-            iconMetaData.TextureOverride = ProcessKeepTogether(handle, iconMetaData, iconMetaData.Texture?.Size ?? (64,64));
+            Vector2i ktSize = iconMetaData.Texture?.Size ?? (64,64);
+            iconMetaData.TextureOverride = ProcessKeepTogether(handle, iconMetaData, ktSize);
+            positionOffset -= ((ktSize/EyeManager.PixelsPerMeter) - Vector2.One) * new Vector2(0.5f); //correct for KT group texture offset
         }
 
         var pixelPosition = (iconMetaData.Position + positionOffset) * EyeManager.PixelsPerMeter;
         var frame = iconMetaData.Texture;
 
+
```
wixoaGit, line 426:
```suggestion
```

wixoaGit (approved):

PR #1772, opened 2024-05-07
Adds vscode tasks and removes them from git ignore

TheLordME:
Adds VSC Tasks for compiling a repo, and running the server and client, via the ctrl. + shift + B shortcut for build tasks in VSC.
Compiling and running the server use an input, whose default has been left blank, but for a repository contained with in the Opendream repo, the input could look like this `./Citadel-Station-13-RP/citadel` the citadel at the end being the name of the .dme and later the .json file, with the ending/extension being added by the command

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -264,6 +264,7 @@ __pycache__/
 # Visual Studio Code workspace settings.
 .vscode/*
 !.vscode/extensions.json
+!.vscode/tasks.json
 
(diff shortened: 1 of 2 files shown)


TheLordME:
This does make for a much nicer workflow, than what I made
https://github.com/OpenDreamProject/OpenDream/wiki/Debugger

PR #1779, opened 2024-05-12
Enable hot reloading of interface files

amylizzle:
[Screencast from 2024-05-12 14-03-37.webm](https://github.com/OpenDreamProject/OpenDream/assets/3855802/d61cefc3-241a-4871-b904-463e9d656fd0)

Adds a new proc to `/world` `ODHotReloadInterface` which forces a reload of the cached interface file and notifies all connected clients to reload it. 

Note this also cleans up some stuff that would probably causes issues with reconnecting too, so that's a nice bonus.


OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -23,6 +23,7 @@
 using Robust.Shared.Utility;
 using SixLabors.ImageSharp;
 using System.Linq;
+using OpenToolkit.GraphicsLibraryFramework;
```
amylizzle, line 26:
```suggestion
```

TestGame/code.dm
```diff
@@ -193,6 +193,12 @@
 			src << "showing main window"
 			winset(src,"mainwindow","is-visible=true")
 
+	verb/test_hot_reload()
+		set category = "Test"
+		src << "tyring hot reload of interface..."
```
amylizzle, line 198:
```suggestion
		src << "trying hot reload of interface..."
```

amylizzle (commented):
typo + using

OpenDreamRuntime/Procs/Native/DreamProcNativeWorld.cs
```diff
@@ -185,6 +185,14 @@ internal static class DreamProcNativeWorld {
             return DreamValue.Null;
         }
 
+
```
amylizzle, line 188:
```suggestion
```

OpenDreamRuntime/Resources/DreamResourceManager.cs
```diff
@@ -78,9 +75,20 @@ public sealed class DreamResourceManager {
                         resource = new DreamResource(resourceId, resourcePath, resourcePath);
                         break;
                 }
+                return resource;
```
amylizzle, line 78:
```suggestion

                return resource;
```

wixoaGit:
I think it would be better to make this a server-side RT command

OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -792,11 +794,20 @@ internal sealed class DreamInterfaceManager : IDreamInterfaceManager {
 
     private void Reset() {
         _userInterfaceManager.MainViewport.Visible = false;
-
+        //close windows if they're open, and clear all child uielements
+        foreach (var window in Windows.Values){
+            window.CloseChildWindow();
+            window.UIElement.RemoveAllChildren();
+        }
```
wixoaGit, line 801:
```suggestion
        }
        
```

OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -792,11 +794,20 @@ internal sealed class DreamInterfaceManager : IDreamInterfaceManager {
...
         Windows.Clear();
         Menus.Clear();
         MacroSets.Clear();
+        //close popups if they're open
```
wixoaGit, line 805:
```suggestion
        
        //close popups if they're open
```

OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -792,11 +794,20 @@ internal sealed class DreamInterfaceManager : IDreamInterfaceManager {
...
+        foreach (var popup in _popupWindows.Values) {
+            popup.Close();
+        }
```
wixoaGit, line 808:
```suggestion
        }
        
```

OpenDreamRuntime/DreamManager.Connections.cs
```diff
@@ -278,5 +279,18 @@ public sealed partial class DreamManager {
         public DreamConnection GetConnectionBySession(ICommonSession session) {
             return _connections[session.UserId];
         }
+
+        public void HotReloadInterface() {
+            string? interfaceText = null;
+            if (_compiledJson.Interface != null)
+                interfaceText = _dreamResourceManager.LoadResource(_compiledJson.Interface, forceReload:true).ReadAsString();
+            var msgLoadInterface = new MsgLoadInterface() {
```
wixoaGit, line 287:
```suggestion

            var msgLoadInterface = new MsgLoadInterface() {
```

OpenDreamRuntime/DreamManager.Connections.cs
```diff
@@ -278,5 +279,18 @@ public sealed partial class DreamManager {
...
+                InterfaceText = interfaceText
+            };
+            foreach (var connection in _connections.Values) {
```
wixoaGit, line 290:
```suggestion

            foreach (var connection in _connections.Values) {
```

OpenDreamRuntime/DreamManager.Connections.cs
```diff
@@ -278,5 +279,18 @@ public sealed partial class DreamManager {
...
+                connection.Session?.Channel.SendMessage(msgLoadInterface);
+            }
+        }
     }
+
```
wixoaGit, line 295:
```suggestion
```

wixoaGit (changes requested):
Linter seems to have missed these whitespace issues.

amylizzle:
> I think it would be better to make this a server-side RT command

I think it's actually better to leave this as a DM side action, so it can be easily integrated into existing admin and debug tooling which is pretty much all DM already.

The eventual plan here is to integrate it with hot reloading actions of IDEs, but until that's in place I think this is the best approach.

wixoaGit:
> > I think it would be better to make this a server-side RT command
> 
> I think it's actually better to leave this as a DM side action, so it can be easily integrated into existing admin and debug tooling which is pretty much all DM already.
> 
> The eventual plan here is to integrate it with hot reloading actions of IDEs, but until that's in place I think this is the best approach.

Can we get an RT command regardless so this can be used on codebases that haven’t added it?

wixoaGit (approved):

PR #1789, opened 2024-05-15
Update RT to 222.4.0

amylizzle:


wixoaGit:
Can you bump this to v222.3.0? It fixes the build errors in latest .NET.

wixoaGit (approved):

PR #1795, opened 2024-05-16
rgb2num accepts null

### LINKED ISSUE #1794: rgb2num() throws a `bad color` exception when given null
Byond behavior is to return white in whatever colorspace it's given.

Kapu1178:
Closes #1794

Im not sure this is technically the correct way to fix this, but it seems fine?


wixoaGit (approved):

wixoaGit:
> Im not sure this is technically the correct way to fix this, but it seems fine?

`DefaultValue` is generally used when the DM reference explicitly mentions it having a default value, but this works.

Issue #1778, opened 2024-05-12
Build warnings/errors are displayed up to 3x in PR annotations


wixoaGit:
![image](https://github.com/OpenDreamProject/OpenDream/assets/30789242/cbea1a22-404f-4b5f-96d6-117553951d5b)

Would be nice to have these 2 warnings only display once.

PR #1796, opened 2024-05-18
Deduplicate github action logs

redmoogle:
Deduplicate the logs

Fixes: #1778

only prints off the warnings from the ubuntu-latest run. 

This is very slightly hacky but only matters for 1 warning (CA1416)


redmoogle:
will do some testing on my fork later

.github/workflows/build-test.yml
```diff
@@ -32,7 +32,7 @@ jobs:
     - name: Install dependencies
       run: dotnet restore
     - name: Build
-      run: dotnet build --configuration Release --no-restore /m
+      run: dotnet build --property WarningLevel=${{ matrix.os == 'ubuntu-latest' && '1' || '0' }} --configuration Release --no-restore /m
```
wixoaGit, line 35: According to the [documentation](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/errors-warnings#warninglevel), the default value here is 4
```suggestion
      run: dotnet build --property WarningLevel=${{ matrix.os == 'ubuntu-latest' && '4' || '0' }} --configuration Release --no-restore /m
```

wixoaGit (changes requested):

wixoaGit:
Another possible approach could be to use the `::remove-matcher` workflow command according to [this documentation](https://github.com/actions/toolkit/blob/main/docs/commands.md#problem-matchers). I think how it would be used here would be to do `echo "::remove-matcher owner=csc::"` on non-ubuntu runners.

wixoaGit (approved):

PR #1797, opened 2024-05-19
lint using resharper

harryob:
implements linting using rider's annotation tool, resharper. these lints aren't annotated using standard github pattern as there is a Ton of them, and resharper outputs them in SARIF. the best fitting tool for this seems to be the codeql security analysis tool, which is odd for a linter, but i think it still provides handy annotation

oh no! new bad code!! (only shows new inspections on the pr, not existing ones)
CleanShot 2024-05-22 at 23 21 59@2x

new commit addresses the issue, the issue is flagged as fixed and minimised
CleanShot 2024-05-22 at 23 27 08@2x

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Setup Resharper
         run: dotnet tool install -g JetBrains.ReSharper.GlobalTools
       - name: Run Linter
-        run: jb inspectcode OpenDream.sln -o="output.json" --project="OpenDream*;DM*" --no-swea --no-build
+        run: jb inspectcode OpenDream.sln -o="output.json" --project="OpenDream*;DM*" --exclude="*.xaml" --no-swea --no-build
       - uses: github/codeql-action/upload-sarif@v3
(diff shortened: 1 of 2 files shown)


wixoaGit:
This gives a ton of errors for our XAML UI. Is there something that can be done about that?
image

harryob:
> This gives a ton of errors for our XAML UI (~6 pages). Is there something that can be done about that? image

that's odd, i never got anything reporting as an error on my fork. i'll add an exclusion for reporting anything on .xaml pages, given it seems to be misreporting for them anyway

wixoaGit:
Turns out to also be giving lots of errors for things defined in RobustToolbox. This is probably due to the RT projects being excluded.
image

harryob:
> Turns out to also be giving lots of errors for things defined in RobustToolbox. This is probably due to the RT projects being excluded. image

added a flag to skip the build to cut linting times, but this ended up also skipping the restore that happened - manually forcing it the restore seems to ensure the analysis works correctly

wixoaGit (approved):

PR #1804, opened 2024-05-22
Animate Defines Cleanup

LemonInTheDark:
Actually defines ANIMATE_CONTINUE and ANIMATE_SLICE

Moves easing flags up next to the animation flags

Moves EASE_IN and EASE_OUT to easing flags instead of blockin em in with the animation flags.

Broke my unit tests REEEEEEEEEEEE


ike709 (approved):
New 515 defines

wixoaGit (approved):

PR #1807, opened 2024-05-24
Adds error message to missing file arg

### LINKED ISSUE #1803: Extremely unclear message when `.dme` is not found
```
Unhandled exception. System.Collections.Generic.KeyNotFoundException: The given key 'MissingIncludedFile' was not present in the dictionary.
```
is not the clearest error message for what should just be `Error: DME file not found`

Laboredih123:
Fixes #1803

If a file passed through compiler arguments doesn't exist, it will now fail compilation and post an error instead of causing an exception.


Laboredih123:
first pr ever wowzers omg awesomes

wixoaGit (approved):

PR #1808, opened 2024-05-24
Implement a few easy operator overloads

amylizzle:
I swore I wouldn't do this again, but here we are.

Implements +, -, *,/, and | operator overloads because they were very easy.

I'm gonna do it in small parts this time. Next PR I'll clean up some of DMOpcodeHandler

Cleans up the emitted warning so that it tells you what operator is not implemented ie 
```
Warning OD0000 at Camera.dm:132:18: operator"" overloads are not implemented. They will be defined but never called.
```

### DIFF
--- a/OpenDreamRuntime/Objects/DreamObject.cs
+++ b/OpenDreamRuntime/Objects/DreamObject.cs
@@ -368,11 +368,11 @@ namespace OpenDreamRuntime.Objects {
 
         #region Operators
         // +
-        public virtual ProcStatus OperatorAdd(DreamValue b, DMProcState state, out DreamValue? result) {
+        public virtual ProcStatus OperatorAdd(DreamValue b, DMProcState state, out DreamValue result) {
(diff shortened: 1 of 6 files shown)


OpenDreamRuntime/Objects/DreamObject.cs
```diff
@@ -368,22 +368,61 @@ public class DreamObject {
 
         #region Operators
         // +
-        public virtual DreamValue OperatorAdd(DreamValue b) {
+        public virtual ProcStatus OperatorAdd(DreamValue b, DMProcState state, out DreamValue? result) {
```
wixoaGit, line 371: The `result` argument in this and all the other operators should be made not nullable since that can have a performance impact on struct values. `result = null` could be replaced with `default` or `DreamValue.Null`

wixoaGit (changes requested):

wixoaGit (approved):

PR #1810, opened 2024-05-25
Update RT to 223.1.2

amylizzle:
Fixes the crash with the mousemap on dodgy graphics drivers

### DIFF
--- a/OpenDreamClient/EntryPoint.cs
+++ b/OpenDreamClient/EntryPoint.cs
@@ -77,12 +77,6 @@ namespace OpenDreamClient {
 
             _dreamInterface.Initialize();
             IoCManager.Resolve<IDreamSoundEngine>().Initialize();
-
-            if (_configurationManager.GetCVar(CVars.DisplayCompat))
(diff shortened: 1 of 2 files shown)


OpenDreamClient/EntryPoint.cs
```diff
@@ -77,12 +77,6 @@ public sealed class EntryPoint : GameClient {
 
             _dreamInterface.Initialize();
             IoCManager.Resolve<IDreamSoundEngine>().Initialize();
-
-            if (_configurationManager.GetCVar(CVars.DisplayCompat))
-                _dreamInterface.OpenAlert(
-                    "Compatibility Mode Warning",
-                    "You are using compatibility mode. Clicking in-game objects is not supported in this mode.",
-                    "Ok", null, null, null);
```
wixoaGit, line 85: Clicking still doesn't work in compatibility mode. This alert needs brought back, along with avoiding calling `GetPixel()` now that it's a crash instead of skipped over.

wixoaGit (approved):

PR #1814, opened 2024-05-27
Make `browse_rsc()` respect cache

amylizzle:
For BYOND, calling `browse_rsc()` should be basically free if the resource already exists on the client. For OD, we were sending the whole file every time. This was very bad for goonstation's performance when a client joined, because it would basically send 40mb of crap every time and hang the client while it was doing so.

Note that `browse()` is supposed to block until `browse_rsc()` is complete, while `output()` is not (according to searches of the BYOND discord).

It's hard to test how much impact this will have locally, but from my testing it should probably shave off like 60 seconds from subsequent joins. Now we just need to sort out the shedload of appearances that get sent.

### DIFF
--- a/OpenDreamClient/Interface/Controls/ControlBrowser.cs
+++ b/OpenDreamClient/Interface/Controls/ControlBrowser.cs
@@ -113,7 +113,10 @@ internal sealed class ControlBrowser : InterfaceControl {
             Stream stream;
             HttpStatusCode status;
             var path = new ResPath(newUri.AbsolutePath);
-            try {
+            if(!_dreamResource.EnsureCacheFile(newUri.AbsolutePath)) {
+                stream = Stream.Null;
+                status = HttpStatusCode.NotFound;
+            } else try {
                 stream = _resourceManager.UserData.OpenRead(_dreamResource.GetCacheFilePath(newUri.AbsolutePath));
                 status = HttpStatusCode.OK;
             } catch (FileNotFoundException) {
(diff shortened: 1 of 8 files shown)


OpenDreamShared/Network/Messages/MsgBrowseResource.cs
```diff
@@ -1,26 +1,22 @@
...
+    }
 
-        public override void WriteToBuffer(NetOutgoingMessage buffer, IRobustSerializer serializer) {
-            buffer.Write(Filename);
-            buffer.WriteVariableInt32(Data.Length);
-            buffer.Write(Data);
-        }
+    public override void WriteToBuffer(NetOutgoingMessage buffer, IRobustSerializer serializer) {
+        buffer.Write(Filename);
+        buffer.WriteVariableInt32(DataHash);
```
wixoaGit, line 20: A data hash feels unlikely to benefit from being encoded as a variable int, this would probably end up taking more space.

OpenDreamClient/Interface/Controls/ControlBrowser.cs
```diff
@@ -113,7 +113,10 @@ public ControlBrowser(ControlDescriptor controlDescriptor, ControlWindow window)
             Stream stream;
             HttpStatusCode status;
             var path = new ResPath(newUri.AbsolutePath);
-            try {
+            if(!_dreamResource.EnsureCacheFile(newUri.AbsolutePath)) {
+                stream = Stream.Null;
+                status = HttpStatusCode.NotFound;
+            } else try {
```
wixoaGit, line 119: Put this `try` in a braced block to more clearly separate the catches from the if/else

wixoaGit (changes requested):

OpenDreamShared/Network/Messages/MsgBrowseResource.cs
wixoaGit: Once it's changed to be something other than file size at least.

amylizzle: might need to wait on RT to make hashing available before we can do this

wixoaGit (approved):

PR #1815, opened 2024-05-28
Appearance Weakrefs

amylizzle:
Fairly substantial rewrite of appearance handling.

There are now two appearance classes: `MutableIconAppearance` and `ImmutableIconAppearance`

`MutableIconAppearance` is the class you use for mutable appearance stuff. They are not sent to the client. To send them to the client, you must either use `AtomManager.SetAppearance()` or *if you have a really really good reason* use `ServerAppearanceSystem.AddAppearance()`.

`ImmutableIconAppearance` is the class that is sent to the client. It cannot be edited. If you want to edit it, you need to create a mutable copy (which is a `MutableIconAppearance`). These are held only as *weakrefs* in the `ServerAppearanceSystem`, and are removed from connected clients when they are garbage collected. This means that if you want to keep it alive you *must have a hard ref to it somewhere*. Any appearance (mutable or no) which has an overlay/underlay counts as a hard ref for that overlay/underlay.

The client has copies of the `ImmutableIconAppearance` and can use `ToMutable()` to edit them locally for animations and such.

Appearance ID is now stored on the `ImmutableIconAppearance` and generated by the `ServerAppearanceSystem`

KNOWN BUG: /turf animations don't apply, it just sets the final state of the turf appearance. Turf animations can go in another PR.


OpenDreamRuntime/Rendering/ServerAppearanceSystem.cs
```diff
@@ -11,8 +11,11 @@ namespace OpenDreamRuntime.Rendering;
 public sealed class ServerAppearanceSystem : SharedAppearanceSystem {
     private readonly Dictionary<IconAppearance, int> _appearanceToId = new();
     private readonly Dictionary<int, IconAppearance> _idToAppearance = new();
+    private readonly Dictionary<IconAppearance, int> _appearanceRefCounts = new();
     private int _appearanceIdCounter;
 
+    private ISawmill? _sawmill;
```
amylizzle, line 17:
```suggestion
```

amylizzle:
needs https://github.com/space-wizards/RobustToolbox/pull/5506

wixoaGit (approved):

PR #1816, opened 2024-05-28
Handle adding/setting null on a filters list

wixoaGit:
Doing `atom.filters += null` is ignored, and `atom.filters[i] = null` removes the filter.


amylizzle:
Fixes #1818

Issue #1821, opened 2024-06-01
`fexists()` is a constant expression within preprocessor statements


ZeWaka:
`Error OD0010 at goonstation.dme:30:14: Unexpected identifier fexists in preprocessor expression`

PR #1823, opened 2024-06-02
Pass ` --suppress-unimplemented` to the compiler workflow

ike709:
Makes it easier to see actual issues.


ike709:
Goon CI failure is due to https://github.com/OpenDreamProject/OpenDream/issues/1821

wixoaGit (approved):

Issue #1842, opened 2024-06-16
Peephole optimizer skips viable optimizations


ike709:
Per wixoa, if an optimization is followed by a label or the end of the proc, it is skipped.
![image](https://github.com/OpenDreamProject/OpenDream/assets/5714543/09e51619-da05-4b1d-8115-308e2753ec9e)

https://github.com/OpenDreamProject/OpenDream/blob/master/DMCompiler/Optimizer/PeepholeOptimizer.cs#L77

![image](https://github.com/OpenDreamProject/OpenDream/assets/5714543/e1d550da-615f-428e-9ec0-425b86fdd5dc)

https://github.com/OpenDreamProject/OpenDream/blob/master/DMCompiler/Optimizer/PeepholeOptimizer.cs#L79

PR #1843, opened 2024-06-16
Peephole optimize proc returns

ike709:
`(PushReferenceValue, Return) -> ReturnReferenceValue`

Per my frequency analysis this pattern occurs ~16k times in Paradise and ~25k times in goon and TG. It is a top-ten offender on all codebases tested.

However, this optimization is heavily kneecapped by #1842 which causes it to only be applied 664 times in Paradise bytecode. Once that issue is fixed this will become much more significant, as a large number of procs will no longer modify their stack when they return.


OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -24,6 +24,13 @@ internal static class DMOpcodeHandlers {
             return ProcStatus.Continue;
         }
 
+        public static ProcStatus ReturnReferenceValue(DMProcState state) {
```
wixoaGit, line 27: This would be better put inside `#region Flow`

wixoaGit (changes requested):

PR #1861, opened 2024-06-26
Fix `goto` inside multiple enumerator loops

### LINKED ISSUE #154: `goto` causes incorrect behavior when used inside loops
The current implementation of `goto` has no knowledge of the code it is jumping into/out of. This can cause problems in code like the following:
```dm
var/list/L = list(list(1, 2), list(3, 4))
for(var/item1 in L)
	world.log << "[item1]"
	if (!istype(item1, /list)) break

	for (var/item2 in item1)
		goto label
	label:
```
What this should do is print "/list" twice, but it instead prints "/list" and "2". This is because when a loop starts enumerating a list it pushes an enumerator onto a special stack, which should be popped when the loop is over. The goto skips over that, and ends up enumerating item1 in the first loop instead of L.

wixoaGit:
Fixes #154

Each DMProcState now has 16 slots for enumerators instead of them being a stack. Opcodes using enumerators must refer to which slot they're using.

### DIFF
--- a/DMCompiler/Bytecode/DreamProcOpcode.cs
+++ b/DMCompiler/Bytecode/DreamProcOpcode.cs
@@ -60,7 +60,7 @@ public enum DreamProcOpcode : byte {
     Modulus = 0x19,
     [OpcodeMetadata(0, OpcodeArgType.Reference)]
     Append = 0x1A,
-    [OpcodeMetadata(-3)]
+    [OpcodeMetadata(-3, OpcodeArgType.EnumeratorId)]
     CreateRangeEnumerator = 0x1B,
     [OpcodeMetadata(0, OpcodeArgType.Reference, OpcodeArgType.Reference)]
     Input = 0x1C,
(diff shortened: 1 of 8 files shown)


DMCompiler/Bytecode/DreamProcOpcode.cs
```diff
@@ -119,11 +119,11 @@ public enum DreamProcOpcode : byte {
     PushFloat = 0x38,
     [OpcodeMetadata(0, OpcodeArgType.Reference)]
     ModulusReference = 0x39,
-    [OpcodeMetadata(-1)]
+    [OpcodeMetadata(-1, OpcodeArgType.EnumeratorId)]
     CreateListEnumerator = 0x3A,
-    [OpcodeMetadata(0, OpcodeArgType.Reference, OpcodeArgType.Label)]
+    [OpcodeMetadata(0, Bytecode.OpcodeArgType.EnumeratorId, OpcodeArgType.Reference, OpcodeArgType.Label)]
```
wixoaGit, line 124:
```suggestion
    [OpcodeMetadata(0, OpcodeArgType.EnumeratorId, OpcodeArgType.Reference, OpcodeArgType.Label)]
```

DMCompiler/Bytecode/DreamProcOpcode.cs
```diff
@@ -223,7 +223,7 @@ public enum DreamProcOpcode : byte {
     TryNoValue = 0x70,
     [OpcodeMetadata]
     EndTry = 0x71,
-    [OpcodeMetadata(0, OpcodeArgType.Label)]
+    [OpcodeMetadata(0, Bytecode.OpcodeArgType.EnumeratorId, OpcodeArgType.Label)]
```
wixoaGit, line 226:
```suggestion
    [OpcodeMetadata(0, OpcodeArgType.EnumeratorId, OpcodeArgType.Label)]
```

PR #1863, opened 2024-06-26
Const-fold some proc statements

ike709:
Const-folds `if(FALSE)` and `do {} while(FALSE)`. I also confirmed that folding `while(FALSE)` is pointless since it never occurs.

To test this I compiled Paradise and confirmed that I can wander around without things being horribly broken, and there wasn't any runtime spam.


ike709:
Need to figure out some issues with scopes.

ike709:
I'll revisit this later.

PR #1873, opened 2024-07-02
Implement `client.show_popup_menus`

### LINKED ISSUE #183: `client/show_popup_menus` not implemented
http://www.byond.com/docs/ref/#/client/var/show_popup_menus

NTstation

ike709:
You can't really tell but I am right clicking while it's disabled and the menu is not showing, as expected:

https://github.com/OpenDreamProject/OpenDream/assets/5714543/e3d038c0-e032-444e-8978-1c53040b6172

Closes #183

### DIFF
--- a/DMCompiler/DMStandard/Types/Client.dm
+++ b/DMCompiler/DMStandard/Types/Client.dm
@@ -20,7 +20,7 @@
 	var/pixel_y = 0 as opendream_unimplemented
 	var/pixel_z = 0 as opendream_unimplemented
 	var/pixel_w = 0 as opendream_unimplemented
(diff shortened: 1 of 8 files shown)


wixoaGit:
What does tg use for its right-click interactions? Would be nice if we could start getting that to work in its place.

ZeWaka:
Client/RightClick probably?

wixoaGit:
> Client/RightClick probably?

This doesn't exist according to the DM reference, but apparently setting `show_popup_menus` to 0 causes right clicks to be sent to `/client/Click()`.

wixoaGit (approved):

Issue #1872, opened 2024-07-02
Toggling darkmode via chat doesn't toggle the statpanel browser's theme


ike709:
Tested on Paradise. In BYOND, toggling the chat theme should automatically toggle the DMF color. But in OpenDream, you have to run the "Toggle Darkmode" verb separately for the DMF to change.

Proof of BYOND behavior:
https://github.com/OpenDreamProject/OpenDream/assets/5714543/15eb695f-1461-4518-81c2-53eb46510ff3

ike709:
This was fixed for TG but Paradise must handle it differently because it's still an issue there.

ike709:
Paradise seems to handle it via cookies:
```js
$('#toggleDarkChat').click(function(e) {
        internalOutput('<span class="internal boldnshit">Dark Chat toggled. Reconnecting to chat.</span>', 'internal');
        var backlog = $messages.html()
        if(getCookie('darkChat') == "on"){
            setCookie('darkChat', "off", 365)
        } else {
            setCookie('darkChat', "on", 365)
        }
        localStorage.setItem('backlog', backlog)
        location.reload();
    });
```

ike709:
This seems to be an issue with us canceling the reload if the paths are identical, but _not_ canceling identical paths causes goonchat to have a seizure. I'm not sure how to differentiate paths that should be canceled from paths that shouldn't.

![image](https://github.com/user-attachments/assets/5012530c-cb58-4ed8-9511-69dd3d2cd318)

wixoaGit:
This no longer happens on master OD and our test server's commit, https://github.com/wixoaGit/Paradise/commit/0bf6ddfae53c05e3b723958cb753dd3c46c70c5d.

PR #1875, opened 2024-07-03
Implement `.output` client command

wixoaGit:
Basic implementation of `.output`

Fixes the stat panel's dark mode in tg

image


ike709:
Closes #1872

PR #1877, opened 2024-07-05
Lint for invalid datum[] indexing

### LINKED ISSUE #1876: Lint against `datum["key"]`
516 breaks this apparently. I'd like to know where they are in tg by checking for something that's explicitly typed as not a list.

https://github.com/tgstation/tgstation/pull/84648

ike709:
Closes #1876 

Invalid in BYOND 515.1641+

Don't want to make it fatal since it'd brick OD compilation of codebases targeting older versions.

### DIFF
--- a/DMCompiler/DM/Expressions/Dereference.cs
+++ b/DMCompiler/DM/Expressions/Dereference.cs
@@ -114,8 +114,11 @@ namespace DMCompiler.DM.Expressions {
                     break;
 
                 case IndexOperation indexOperation:
-                    if (NestedPath is not null && NestedPath.Value.IsDescendantOf(DreamPath.Datum)) {
-                        DMCompiler.Emit(WarningCode.InvalidIndexOperation, Location, "Invalid index operation. datum[] index operations are not valid starting in BYOND 515.1641");
+                    if (NestedPath is not null) {
+                        var obj = DMObjectTree.GetDMObject(NestedPath.Value, false);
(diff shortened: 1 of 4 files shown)


ike709:
CI failure is unrelated.

amylizzle:
This is warning on goon's `/datum/db_record` which has `operator[]` defined.

ike709:
Fixed it.

DMCompiler/DM/Expressions/Dereference.cs
```diff
@@ -114,6 +114,13 @@ public Dereference(Location location, DreamPath? path, DMExpression expression,
                     break;
 
                 case IndexOperation indexOperation:
+                    if (NestedPath is not null) {
+                        var obj = DMObjectTree.GetDMObject(NestedPath.Value, false);
+                        if (obj is not null && obj.IsSubtypeOf(DreamPath.Datum) && !obj.HasProc("operator[]")) {
+                            DMCompiler.Emit(WarningCode.InvalidIndexOperation, Location, "Invalid index operation. datum[] index operations are not valid starting in BYOND 515.1641");
+                        }
+                    }
```
wixoaGit, line 122: This should also be done in `EmitReference()` so that we can emit a warning for `datum["foo"] = value`.

wixoaGit (changes requested):

DMCompiler/DM/Expressions/Dereference.cs
amylizzle: don't forget to check for `operator[]=` instead

wixoaGit (approved):

PR #1878, opened 2024-07-05
Fix sandbox errors in .NET 8.0.300

wixoaGit:
string + char now trips the sandbox


OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -441,27 +441,31 @@ internal sealed class DreamInterfaceManager : IDreamInterfaceManager {
...
                         }
-                        currentArg += argsRaw[1][i];
+
+                        currentArg.Append(argsRaw[1][i]);
                     }
-                    if(!string.IsNullOrEmpty(currentArg))
-                        args.Add(HandleEmbeddedWinget(null ,currentArg));
+
+                    if(currentArg.ToString() is { } arg && !string.IsNullOrEmpty(arg))
+                        args.Add(HandleEmbeddedWinget(null ,arg));
```
wixoaGit, line 468:
```suggestion
                        args.Add(HandleEmbeddedWinget(null, arg));
```

PR #1884, opened 2024-07-10
Handle appending `null` to `vis_contents`

ike709:
Appears to be a no-op in BYOND:

image


OpenDreamRuntime/Objects/Types/DreamList.cs
```diff
@@ -827,7 +827,9 @@ public sealed class DreamVisContentsList : DreamList {
                 return; // vis_contents cannot contain duplicates
             _visContents.Add(turf);
             entity = EntityUid.Invalid; // TODO: Support turfs in vis_contents
-        } else {
+        } else if (value == DreamValue.Null) {
+            return; // vis_contents cannot contain nulls
+        }else {
```
ike709, line 832:
```suggestion
        } else {
```

wixoaGit (approved):

PR #1891, opened 2024-07-19
Add CI testing for DMDisassembler

### LINKED ISSUE #1879: CI should verify ProcDecoder is keeping up with opcode modifications
ProcDecoder is often forgotten when modifications are made to opcodes. This leads to a broken debugger and DMDisassembler. 

A test to ensure DMDisassembler's `test-all` works on a SS13 codebase would be nice. An alternative would be to rewrite ProcDecoder to use OpcodeMetadataAttribute so it can work with most opcodes without needing explicit work.

ike709:
Resolves #1879 but I think we should open a separate issue for modifying ProcDecoder to use OpcodeMetadataAttribute.

Changes:
- DMDisassembler's commands can now be passed as CLI arguments
- Added `crash-on-error` which is a `test-all` wrapper that kills DMDisasembler if it finds errors (to keep the GitHub Action simple)
- CI now builds `DMDisassembler` and runs `crash-on-error` on TG

### DIFF
--- a/DMDisassembler/DMDisassembler.csproj
+++ b/DMDisassembler/DMDisassembler.csproj
@@ -4,6 +4,7 @@
     <TargetFramework>net8.0</TargetFramework>
     <Configurations>Debug;Release;Tools</Configurations>
     <Platforms>AnyCPU</Platforms>
+    <OutputPath>..\bin\DMDisassembler\</OutputPath>
   </PropertyGroup>
 
   <ItemGroup>
--- a/DMDisassembler/DMDisassembler.csproj
+++ b/DMDisassembler/DMDisassembler.csproj
@@ -4,6 +4,7 @@
     <TargetFramework>net8.0</TargetFramework>
     <Configurations>Debug;Release;Tools</Configurations>
     <Platforms>AnyCPU</Platforms>
+    <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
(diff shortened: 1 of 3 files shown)


DMDisassembler/DMDisassembler.csproj
```diff
@@ -4,6 +4,7 @@
     <TargetFramework>net8.0</TargetFramework>
     <Configurations>Debug;Release;Tools</Configurations>
     <Platforms>AnyCPU</Platforms>
+    <OutputPath>..\bin\DMDisassembler\</OutputPath>
```
ike709, line 7:
```suggestion
    <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
    <OutputPath>..\bin\DMDisassembler\</OutputPath>
```

DMDisassembler/Program.cs
```diff
@@ -31,31 +31,50 @@ internal class Program {
         LoadAllTypes();
 
         bool acceptingCommands = true;
+
+        // Commands passed as CLI arguments
+        if (args.Length > 1) {
+            ParseCommand(args[1..], out acceptingCommands);
```
wixoaGit, line 37: Allowing a single command to be executed from the command line doesn't make much sense for most of these. I'd rather this just check for a `crash-on-error` argument instead of making it a command.

wixoaGit (changes requested):

wixoaGit (approved):

PR #1903, opened 2024-07-29
HtmlParser: Support named entities in body text

20kdc:
This gives HtmlParser the ability to support HTML character entities in body text. This is primarily useful to clean up text such as "`&quot;`", "`&apos;`" ,and "`&#33;`".

This was tested with [Tomeno's OpenDream branch of Console](https://github.com/Tomeno/console/tree/opendream_cope) but with changes to interface/default.dmf reverted and then the computer_docked window deleted (it attempts to share ownership of controls). When operating a computer and "> "-commands are typed, the window no longer shows HTML entities with this patch.

...I'm sure there are simpler ways to test this, but that's where I found the issue.

### DIFF
--- a/OpenDreamClient/Interface/Html/HtmlParser.cs
+++ b/OpenDreamClient/Interface/Html/HtmlParser.cs
@@ -88,6 +88,7 @@ public static class HtmlParser {
                         // browsers usually allow for some fallibility here
                         break;
                     }
+                    
                     string insideEntity = text.Substring(i + 1, end - (i + 1));
                     i = end;
 
@@ -114,6 +115,7 @@ public static class HtmlParser {
                                 break;
                         }
                     }
+                    
                     break;
                 default:
                     currentText.Append(c);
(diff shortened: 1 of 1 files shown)


OpenDreamClient/Interface/Html/HtmlParser.cs
```diff
@@ -81,6 +81,40 @@ public static class HtmlParser {
...
+                            case "yen": currentText.Append("¥"); break;
+                            case "euro": currentText.Append("€"); break;
+                            case "copyright": currentText.Append("©"); break;
+                            case "trademark": currentText.Append("®"); break;
+                            default:
+                                currentText.Append("&" + insideEntity + ";");
+                                break;
+                        }
+                    }
+                    break;
```
wixoaGit, line 117:
```suggestion
                    }
                    
                    break;
```

wixoaGit (changes requested):

wixoaGit (approved):

wixoaGit:
> ...I'm sure there are simpler ways to test this, but that's where I found the issue.

TestGame's say verb isn't escaped so that works fine

PR #1905, opened 2024-08-05
restores opendream_unimplemented warnings for procs with empty bodies in DMStandard

harryob:
currently some opendream_unimplemented warnings from dmstandard aren't being emitted, as we null both the Statement and SetStatement blocks if the Statement block is empty (due to having an empty proc body)

### DIFF
--- a/Content.Tests/DMTests.cs
+++ b/Content.Tests/DMTests.cs
@@ -26,9 +26,10 @@ public sealed class DMTests : ContentUnitTest {
         NoError = 0,        // Should run without errors
         Ignore = 1,         // Ignore entirely
         CompileError = 2,   // Should fail to compile
-        RuntimeError = 4,   // Should throw an exception at runtime
-        ReturnTrue = 8,     // Should return TRUE
-        NoReturn = 16,      // Shouldn't return (aka stopped by a stack-overflow or runtimes)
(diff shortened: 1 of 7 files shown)


Content.Tests/DMTests.cs
```diff
@@ -26,9 +26,10 @@ public enum DMTestFlags {
         NoError = 0,        // Should run without errors
         Ignore = 1,         // Ignore entirely
         CompileError = 2,   // Should fail to compile
-        RuntimeError = 4,   // Should throw an exception at runtime
-        ReturnTrue = 8,     // Should return TRUE
-        NoReturn = 16,      // Shouldn't return (aka stopped by a stack-overflow or runtimes)
+        CompileWarning = 4, // Should throw a warning during compile
```
wixoaGit, line 29: We usually test for warnings by putting `#pragma EmissionCode error` inside the DM file. I would rather unimplemented warnings be given a code so we can do that instead.

wixoaGit (changes requested):

Content.Tests/DMTests.cs
harryob: that makes sense, addressed

wixoaGit (approved):

PR #1917, opened 2024-08-12
Implement `as` syntax for multiple base types (`mob|obj|turf`)

MyNameIsRomayne:
This is an attempt to support code like
```
for (var/atom/A as mob|obj|turf|area in view(null, null))
```
to continue the implementation parity between OD and DM


MyNameIsRomayne:
This is probably not a great first project (it was supposedly easy at first, but the proper implementation is slightly more complicated) for me, so the code quality is dubious.

Current issues are something along the lines of
`Error OD0000 at code/mob/living/silicon/ai.dm:1102:11: Negative stack size`
though this should all work fine once these errors are resolved (?)

DMCompiler/DM/Builders/DMProcBuilder.cs
```diff
@@ -612,8 +612,10 @@ internal sealed class DMProcBuilder(DMObject dmObject, DMProc proc) {
                 implicitTypeCheck = dmTypes.Value.TypePath;
             } else if (!dmTypes.Value.IsAnything) {
                 // "as anything" performs no check. Other values are unimplemented.
-                DMCompiler.UnimplementedWarning(outputVar.Location,
-                    $"As type {dmTypes} in for loops is unimplemented. No type check will be performed.");
+                byte typeId = (byte) dmTypes.Value.Type;
+                proc.CreateFilteredBaseTypesListEnumerator(typeId, dmTypes.Value.Type);
```
wixoaGit, line 616: This is likely your issue. CreateFilteredBaseTypesEnumerator expects to have a list to pop off the stack, but the `list.EmitPushValue()` is called after this instead of before. While the compiler is keeping track of the stack's expected size, this ends up trying to pop something off an empty stack so the compiler raises that error.

MyNameIsRomayne: I may have interpreted this wrong, but I moved the call to CreateFilteredBaseTypesListEnumerator after EmitPushValue, and I still receive the same error.
```C#
// Depending on the var's type and possibly a given "as [types]", an implicit istype() check is performed
DreamPath? implicitTypeCheck = null;
if (dmTypes == null) {
    // No "as" means the var's type will be used
    implicitTypeCheck = lValue?.Path;
} else if (dmTypes.Value.TypePath != null) {
    // "as /datum" will perform a check for /datum
    implicitTypeCheck = dmTypes.Value.TypePath;
}

list.EmitPushValue(dmObject, proc);
if (implicitTypeCheck != null) {
    if (DMObjectTree.TryGetTypeId(implicitTypeCheck.Value, out var filterTypeId)) {
        // Create an enumerator that will do the implicit istype() for us
        proc.CreateFilteredListEnumerator(filterTypeId, implicitTypeCheck.Value);
    } else {
        DMCompiler.Emit(WarningCode.ItemDoesntExist, outputVar.Location,
            $"Cannot filter enumeration by type {implicitTypeCheck.Value}, it does not exist");
        proc.CreateListEnumerator();
    }
} else if (dmTypes != null && !dmTypes.Value.IsAnything) {
    byte typeId = (byte) dmTypes.Value.Type;
    proc.CreateFilteredBaseTypesListEnumerator(typeId, dmTypes.Value.Type);
} else {
    proc.CreateListEnumerator();
}
```

MyNameIsRomayne: NVM got it to work. was an issue with not having a case for the new annotatedbytecode

MyNameIsRomayne:
It compiles without errors so I'm undrafting this and re-requesting review

DMCompiler/DM/Builders/DMProcBuilder.cs
```diff
@@ -602,32 +602,36 @@ public void ProcessStatementForList(DMExpression list, DMExpression outputVar, D
...
                 // "as /datum" will perform a check for /datum
                 implicitTypeCheck = dmTypes.Value.TypePath;
             } else if (!dmTypes.Value.IsAnything) {
-                // "as anything" performs no check. Other values are unimplemented.
-                DMCompiler.UnimplementedWarning(outputVar.Location,
-                    $"As type {dmTypes} in for loops is unimplemented. No type check will be performed.");
+                // "as anything" performs no check. Other than that, all that's left are type assignments like `as mob|obj|area`.
+                proc.CreateFilteredBaseTypesListEnumerator((byte) dmTypes.Value.Type, dmTypes.Value.Type);
+                alreadyCreatedEnumerator = true;
+                DMCompiler.VerbosePrint($"Created CreateFilteredBaseTypesListEnumerator");
```
ZeWaka, line 619: do we want this print here?

DMCompiler/DM/DMObject.cs
```diff
@@ -230,4 +230,8 @@ public DMValueType GetDMValueType() {
 
         return DMValueType.Anything;
     }
+
+    public DreamPath[] GetDMAncestors() {
+
+    }
```
ZeWaka, line 236: ?

amylizzle:
Abandoned?

MyNameIsRomayne:
Yeah these haven't been worked on for a while, I'm not sure if I'll be able to get to them. Someone else can continue it or work on it if they want

wixoaGit:
Closing in favor of #2228

PR #1919, opened 2024-08-12
Implements `hearers()` and `ohearers()`

### LINKED ISSUE #1755: `hearers()` not implemented
https://www.byond.com/docs/ref/#/proc/hearers

TheTyrant25:
Fixes #1755

Using very similar code to `viewers()` and `oviewers`, a new boolean argument has been added to `CalculateVisibility()`, called `ignore_light` (false by default). Setting it to true will skip all the luminosity checks in the algorithm. As there aren't actually any at the moment, it does nothing different, but once `CalculateVisibility()` is fully finished, it hopefully should.


OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -1014,6 +1014,68 @@ internal static class DreamProcNativeRoot {
...
+                return new(hear);
+
+            var centerPos = bundle.AtomManager.GetAtomPosition(center);
+            if (!depthValue.TryGetValueAsInteger(out var depth))
+                depth = 5;
+
+            foreach (var atom in bundle.AtomManager.EnumerateAtoms(bundle.ObjectTree.Mob)) {
+                var mob = (DreamObjectMob)atom;
+
+                if (centerPos.Z == mob.Z && Math.Abs(centerPos.X - mob.X) <= depth && Math.Abs(centerPos.Y - mob.Y) <= depth) {
```
amylizzle, line 1054: Are you doing `in range(5)` and *then* doing `in range(view string)`? Because that doesn't seem right.
It seems like you should instead be gathering the tiles in range, calculating audibility, *then* iterating the mobs to see if they're on an audible tile

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -1014,6 +1014,68 @@ internal static class DreamProcNativeRoot {
             return new DreamValue(obj.ObjectDefinition.HasProc(procName) ? 1 : 0);
         }
 
+        [DreamProc("hearers")]
+        [DreamProcParameter("Depth", Type = DreamValueTypeFlag.Float)]
+        [DreamProcParameter("Center", Type = DreamValueTypeFlag.DreamObject)]
+        public static DreamValue NativeProc_hearers(NativeProc.Bundle bundle, DreamObject? src, DreamObject? usr) { //TODO: Change depending on center
```
ZeWaka, line 1020: what's this todo

amylizzle:
Also you could probably add a test in testgame? Would be good.

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
TheTyrant25: ah that's copied over from `viewers()`. I left it in because whatever implementation that proc does should do a similar thing to hearers I think

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
TheTyrant25: this is how viewers() implements it. I didn't look too closely but from what I can tell, line 1048 tries to get depth and defaults to 5 if it fails
```cs
            if (!depthValue.TryGetValueAsInteger(out var depth))
                depth = 5;
```

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
TheTyrant25: afaik the center arg doesn't work correctly and I don't know how to fix it

Kapu1178 (changes requested):
Discussed on discord but just for a more permanent note, this is just... not how byond hearers() works. hearers() is viewers() ignoring luminosity (or a lack thereof)

DMCompiler/DMStandard/_Standard.dm
```diff
@@ -31,6 +31,7 @@ proc/fract(n) as num
 proc/ftime(File, IsCreationTime = 0) as num
 proc/gradient(A, index)
 proc/hascall(Object, ProcName) as num
+proc/hearers(Depth = 5, Center = usr) as /list
```
wixoaGit, line 34: According to the ref, the default for this is `world.view` and not 5
```suggestion
proc/hearers(Depth = world.view, Center = usr) as /list
```

DMCompiler/DMStandard/_Standard.dm
```diff
@@ -66,6 +67,7 @@ proc/num2text(N, A, B) as text
 proc/orange(Dist = 5, Center = usr) as /list|null // NOTE: Not sure if return types have BYOND parity
 proc/oview(Dist = 5, Center = usr) as /list
 proc/oviewers(Depth = 5, Center = usr) as /list
+proc/ohearers(Depth = 5, Center = usr) as /list
```
wixoaGit, line 70: Also a default of `world.view`
```suggestion
proc/ohearers(Depth = world.view, Center = usr) as /list
```

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -1010,6 +1010,68 @@ public static DreamValue NativeProc_hascall(NativeProc.Bundle bundle, DreamObjec
         return new DreamValue(obj.ObjectDefinition.HasProc(procName) ? 1 : 0);
     }
 
+    [DreamProc("hearers")]
+    [DreamProcParameter("Depth", Type = DreamValueTypeFlag.Float)]
+    [DreamProcParameter("Center", Type = DreamValueTypeFlag.DreamObject)]
+    public static DreamValue NativeProc_hearers(NativeProc.Bundle bundle, DreamObject? src, DreamObject? usr) { //TODO: Change depending on center
```
wixoaGit, line 1016: Rather than copy-pasting all of `viewers()`, this code should be moved into a method `DreamProcNativeHelpers.cs` that can be called by both native procs.

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -1813,6 +1875,71 @@ public static DreamValue NativeProc_num2text(NativeProc.Bundle bundle, DreamObje
         // Maybe an exception is better?
         return new DreamValue("0");
     }
+ 
+    [DreamProc("ohearers")]
+    [DreamProcParameter("Depth", Type = DreamValueTypeFlag.Float)]
+    [DreamProcParameter("Center", Type = DreamValueTypeFlag.DreamObject)]
+    public static DreamValue NativeProc_ohearers(NativeProc.Bundle bundle, DreamObject? src, DreamObject? usr) { //TODO: Change depending on center
+        DreamValue depthValue = new DreamValue(5);
```
wixoaGit, line 1883: Same here as with `hearers()`

wixoaGit (changes requested):

ike709:
https://github.com/TheTyrant25/OpenDream/pull/1

wixoaGit (approved):

PR #1776, opened 2024-05-11
Softdels part 1 (Datum softdels)

moonheart08:
Implements softdel support for /datum, changing internal refs to be weak as to avoid pinning the objects forever.

PR #1920, opened 2024-08-12
ADOPTED: Datum softdels 

amylizzle:
Adopted from #1776 

Seems to have dropped 2Gb from the memory usage of goon on cog2.

Draft because some mildly weird stuff needs investigating first


amylizzle:
Running goon with *only* softdels (ie, the whole `ifdef opendream del()` thing disabled) reduces memory usage by the same amount. Using goon's debug tools for softdels I get this for OD:
image
compared to this for BYOND:
image

OpenDreamRuntime/DreamManager.Connections.cs
```diff
@@ -358,3 +358,24 @@ public void Execute(IConsoleShell shell, string argStr, string[] args) {
         dreamManager.HotReloadResource(args[0]);
     }
 }
+
+
+public sealed class ForceGarbageCollectionCommand : IConsoleCommand {
```
wixoaGit, line 362: This is already a server command provided by RT, you can run `gc` in the server's console.

OpenDreamRuntime/Procs/DMProc.cs
```diff
@@ -543,6 +543,7 @@ public override void Dispose() {
             _stack = null;
 
             _dreamValuePool.Return(_localVariables);
+            Array.Clear(_localVariables);
```
wixoaGit, line 546: `ArrayPool.Return()` has a `clearArray` parameter.
```suggestion
            _dreamValuePool.Return(_localVariables, true);
```

wixoaGit (changes requested):

OpenDreamRuntime/DreamManager.Connections.cs
amylizzle: oh no worries, I'll delete

wixoaGit (approved):

PR #1921, opened 2024-08-13
Implement replacetextEx_char, reimplement Splittext to match new 515 behavior

MyNameIsRomayne:
Very simple one-liner piggybacking off jointext and spltitext, with an update to Splittext so that it functions properly.

Basically, splittext has been updated in 515 to include the start and end of the string, so the current implementation of Splittext now remembers that and pre/appends those parts to the first/last elements.

The splittext unit tests were updated based off the output from DM for 515:
image

Originally this was just going to be a one-liner for replacetextEx_char, but after being informed of the changes needed to Splittext I decided to go ahead and do that too, so that this works.

### DIFF
--- a/Content.Tests/DMProject/Tests/Text/Splittext.dm
+++ b/Content.Tests/DMProject/Tests/Text/Splittext.dm
@@ -5,19 +5,19 @@
 	ASSERT(test1 ~= test1_expected)
 
 	var/list/test2 = splittext(test_text, " ", 5)
-	var/test2_expected = list("average","of","1,","2,","3,","4,","5","is:","3")
+	var/test2_expected = list("The average","of","1,","2,","3,","4,","5","is:","3")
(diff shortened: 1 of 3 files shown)


MyNameIsRomayne:
Drafting due to a new test I added failing, going to resolve that (Empty strings should be splitting on every char, but instead dont split at all)

MyNameIsRomayne:
image
Reopening because I am genuinely unable to fix this without worsening the code quality beyond the review-necessary state it in. Please advise on this

amylizzle:
Maybe mark as draft if it's not ready for merge

MyNameIsRomayne:
Yeah these haven't been worked on for a while, I'm not sure if I'll be able to get to them

MyNameIsRomayne:
Not working on this anymore, unfortunately. I don't feel confident I could make genuine contributions to the project without significantly taking up other's time

PR #1922, opened 2024-08-14
Convert dreamprocnativeroot.cs to file-scoped namespace

TheTyrant25:
Has the fun effect of unindenting nearly the entire file. Aside from removing two curly braces and adding a semicolon, nothing else was changed. The diff got fucky due to all the blank lines, which is annoying. nothing should have broken though.

Did it because it was throwing a warning, and the file has 39. with this we can reduce it to 38. 👍 
Edit: the file now has 15 warnings. this PR makes it 14.


TheTyrant25:
merge resolution was a bit fucky but i confirmed that it worked as intended by copying the entire file from master, pasting it in, and then doing The Process (untabbing nearly the whole file, removing the outermost curly braces, and adding a semicolon).

wixoaGit (approved):

PR #1923, opened 2024-08-14
Clean up lots of warnings in DreamProcNativeRoot.cs

TheTyrant25:
By adding `!` and `?` at strategic locations, as well as some other changes, the warnings about dereferencing/referencing of null/nullable types are removed. in (presumably) all the cases where the warning exists there are already circumstances that make it so that the errors being warned against can't happen anyway, so these shouldn't hurt. fingers crossed.

The file has gone from having 39 warnings to only 14.

Some of the warnings that are more legitimate in origin have been left in.

### DIFF
--- a/OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
+++ b/OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
@@ -651,13 +651,13 @@ namespace OpenDreamRuntime.Procs.Native {
 
             string filePath;
             if (file.TryGetValueAsDreamResource(out var resource)) {
-                filePath = resource.ResourcePath;
-            } else if(!file.TryGetValueAsString(out filePath)) {
+                filePath = resource.ResourcePath!;
+            } else if(!file.TryGetValueAsString(out filePath!)) {
                 throw new Exception($"{file} is not a valid file");
             }
 
             bool successful;
-            if (filePath.EndsWith("/")) {
+            if (filePath!.EndsWith("/")) {
                 successful = bundle.ResourceManager.DeleteDirectory(filePath);
             } else {
                 successful = bundle.ResourceManager.DeleteFile(filePath);
@@ -673,12 +673,12 @@ namespace OpenDreamRuntime.Procs.Native {
 
             string filePath;
             if (file.TryGetValueAsDreamResource(out var rsc)) {
-                filePath = rsc.ResourcePath;
-            } else if (!file.TryGetValueAsString(out filePath)) {
+                filePath = rsc.ResourcePath!;
+            } else if (!file.TryGetValueAsString(out filePath!)) {
                 return DreamValue.Null;
             }
 
-            return new DreamValue(bundle.ResourceManager.DoesFileExist(filePath) ? 1 : 0);
+            return new DreamValue(bundle.ResourceManager.DoesFileExist(filePath!) ? 1 : 0);
         }
 
         [DreamProc("file")]
@@ -708,7 +708,7 @@ namespace OpenDreamRuntime.Procs.Native {
 
             if (file.TryGetValueAsString(out var rscPath)) {
                 resource = bundle.ResourceManager.LoadResource(rscPath);
-            } else if (!file.TryGetValueAsDreamResource(out resource)) {
+            } else if (!file.TryGetValueAsDreamResource(out resource!)) {
                 return DreamValue.Null;
             }
 
@@ -788,7 +788,7 @@ namespace OpenDreamRuntime.Procs.Native {
             int start = bundle.GetArgument(2, "Start").MustGetValueAsInteger(); //1-indexed
             int end = bundle.GetArgument(3, "End").MustGetValueAsInteger(); //1-indexed
 
-            if (start > text.Length || start == 0) return new DreamValue(0);
+            if (start > text!.Length || start == 0) return new DreamValue(0);
 
             if (start < 0) {
                 start = text.Length + start + 1; //1-indexed
@@ -808,7 +808,7 @@ namespace OpenDreamRuntime.Procs.Native {
                 return match.Success ? new DreamValue(match.Index + 1) : new DreamValue(0);
             }
 
-            int needleIndex = text.IndexOf(needle, start - 1, end - start, StringComparison.OrdinalIgnoreCase);
+            int needleIndex = text.IndexOf(needle!, start - 1, end - start, StringComparison.OrdinalIgnoreCase);
             return new DreamValue(needleIndex + 1); //1-indexed
         }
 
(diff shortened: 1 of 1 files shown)


OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -3298,7 +3298,7 @@
         public static DreamValue NativeProc_winset(NativeProc.Bundle bundle, DreamObject? src, DreamObject? usr) {
             DreamValue player = bundle.GetArgument(0, "player");
             DreamValue controlId = bundle.GetArgument(1, "control_id");
-            string winsetControlId = (!controlId.IsNull) ? controlId.GetValueAsString() : null;
+            string winsetControlId = (!controlId.IsNull) ? controlId.GetValueAsString() : null!;
```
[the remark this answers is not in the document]
TheTyrant25: this warning existed previously and i'm considering it out of scope here, since there's a few other cases of the same thing. May be fixed in another PR

wixoaGit:
Using `!` is generally something to avoid since it's suppressing an actual potential issue. It's better to instead do things differently to ensure the value could never be null, or check if the value is null and handle it accordingly.

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -651,18 +651,12 @@ public static DreamValue NativeProc_fdel(NativeProc.Bundle bundle, DreamObject?
 
             string filePath;
             if (file.TryGetValueAsDreamResource(out var resource)) {
-                filePath = resource.ResourcePath;
-            } else if(!file.TryGetValueAsString(out filePath)) {
+                filePath = resource.ResourcePath!;
```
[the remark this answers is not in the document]
TheTyrant25: it appears that making filePath nullable has removed the warning anyway.

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -2532,7 +2526,7 @@ public static DreamValue NativeProc_splicetext(NativeProc.Bundle bundle, DreamOb
                 else
                     return new DreamValue(insertText);
             else if(text == "")
-                return new DreamValue(insertText);
+                return new DreamValue(insertText!);
```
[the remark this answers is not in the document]
TheTyrant25: i'll revert it to how it was then, same with the others

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -3298,7 +3292,7 @@
         public static DreamValue NativeProc_winset(NativeProc.Bundle bundle, DreamObject? src, DreamObject? usr) {
             DreamValue player = bundle.GetArgument(0, "player");
             DreamValue controlId = bundle.GetArgument(1, "control_id");
-            string winsetControlId = (!controlId.IsNull) ? controlId.GetValueAsString() : null;
+            string? winsetControlId = (!controlId.IsNull) ? controlId.GetValueAsString() : null;
```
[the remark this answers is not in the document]
TheTyrant25: preexisting issue; out of scope

TheTyrant25:
oh huh it's failing ubuntu-latest build test and release

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -649,20 +649,14 @@ public static DreamValue NativeProc_fcopy_rsc(NativeProc.Bundle bundle, DreamObj
...
             }
 
-            bool successful;
-            if (filePath.EndsWith("/")) {
-                successful = bundle.ResourceManager.DeleteDirectory(filePath);
-            } else {
-                successful = bundle.ResourceManager.DeleteFile(filePath);
-            }
-
+            bool successful = filePath!.EndsWith("/") ? bundle.ResourceManager.DeleteDirectory(filePath) : bundle.ResourceManager.DeleteFile(filePath);
```
wixoaGit, line 659: This is still suppressing the issue I pointed out earlier. `resource.ResourcePath` could be null, which is why the compiler warned about it in the first place.

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -673,8 +667,8 @@ public static DreamValue NativeProc_fexists(NativeProc.Bundle bundle, DreamObjec
 
             string filePath;
             if (file.TryGetValueAsDreamResource(out var rsc)) {
-                filePath = rsc.ResourcePath;
-            } else if (!file.TryGetValueAsString(out filePath)) {
+                filePath = rsc.ResourcePath!;
+            } else if (!file.TryGetValueAsString(out filePath!)) {
```
wixoaGit, line 671: Mark `filePath` as nullable like you did in `fdel`

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -708,7 +702,7 @@ public static DreamValue NativeProc_file2text(NativeProc.Bundle bundle, DreamObj
 
             if (file.TryGetValueAsString(out var rscPath)) {
                 resource = bundle.ResourceManager.LoadResource(rscPath);
-            } else if (!file.TryGetValueAsDreamResource(out resource)) {
+            } else if (!file.TryGetValueAsDreamResource(out resource!)) {
```
wixoaGit, line 705: Mark `resource` as nullable

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -2546,7 +2542,7 @@ public static DreamValue NativeProc_splicetext(NativeProc.Bundle bundle, DreamOb
             if(start == 0 || start > text.Length || start > end)
                 throw new Exception("bad text or out of bounds");
 
-            string result = text.Remove(start - 1, (end-start)).Insert(start - 1, insertText);
+            string result = text.Remove(start - 1, (end-start)).Insert(start - 1, insertText!);
```
wixoaGit, line 2545: `insertText` still also could be null here if the DM code didn't give a string.

wixoaGit (changes requested):

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -671,9 +665,9 @@ public static DreamValue NativeProc_fdel(NativeProc.Bundle bundle, DreamObject?
         public static DreamValue NativeProc_fexists(NativeProc.Bundle bundle, DreamObject? src, DreamObject? usr) {
             DreamValue file = bundle.GetArgument(0, "File");
 
-            string filePath;
+            string? filePath;
             if (file.TryGetValueAsDreamResource(out var rsc)) {
-                filePath = rsc.ResourcePath;
+                filePath = rsc.ResourcePath!;
```
wixoaGit, line 670: This could be null
```suggestion
                filePath = rsc.ResourcePath;
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #1924, opened 2024-08-15
Convert DreamValue.cs to file-scoped namespace

TheTyrant25:
Same reasoning and method as #1922, it gets rid of a warning and an indent.

### DIFF
--- a/OpenDreamRuntime/DreamValue.cs
+++ b/OpenDreamRuntime/DreamValue.cs
@@ -16,745 +16,744 @@ using Robust.Shared.Serialization.TypeSerializers.Interfaces;
 using OpenDreamRuntime.Procs.Native;
 using Robust.Shared.Utility;
 
-namespace OpenDreamRuntime {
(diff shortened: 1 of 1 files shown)


TheTyrant25:
those should all be pre existing issues

OpenDreamRuntime/DreamValue.cs
```diff
@@ -16,745 +16,744 @@
...
-            String        = 1,
-            Float         = 2,
-            DreamResource = 3,
-            DreamObject   = 4,
-            DreamType     = 5,
-            DreamProc     = 6,
-            Appearance    = 7
-            // @formatter:on
-        }
+namespace OpenDreamRuntime;
```
wixoaGit, line 19:
```suggestion
namespace OpenDreamRuntime;

```

wixoaGit (changes requested):

wixoaGit (approved):

PR #1927, opened 2024-08-16
Fix compiling TestGame in BYOND

ike709:
The renderer tests use these procs that BYOND is missing.


amylizzle (approved):
Oh that's nice than the way I was gonna do it

wixoaGit (approved):

PR #1933, opened 2024-08-16
fix minor positioning mismatch

amylizzle:
Reverses the order of position X sorting to match BYOND.

Also adds a comment to FLOAT_LAYER handling


amylizzle:
nope

PR #1946, opened 2024-08-27
Fix modifiers as keybinds, add ".winset :type" support

### LINKED ISSUE #1936: `.winset :map` is not implemented
```
macro "default"
    elem ".winset :map.right-click=false"
        name = "SHIFT+Shift"
    elem "Shift"
        name = "SHIFT"
        command = ".winset :map.right-click=false"
    elem "ShiftUp"
        name = "SHIFT+UP"
        command = ".winset :map.right-click=true"
```

This is needed for TG Shift+RMB behavior, which just causes the context menu to appear by temporarily disabling the `right-click` parameter.

ike709:
Resolves #1936 

- In the DMF, modifier keys are now treated as keys when no other key is present. This appears to be more in line with the behavior described in the ref.
- `.winset :[type].whatever` is now implemented, where `:[type]` is the default control of that type

This makes TG shift+RMB behavior functional.

RMB toggles the cabinet's open/closed state. Shift+RMB opens ye olde context menu.

https://github.com/user-attachments/assets/c4a3949a-c088-4f12-9aeb-764bb8b5227c

### DIFF
--- a/OpenDreamClient/Interface/DreamInterfaceManager.cs
+++ b/OpenDreamClient/Interface/DreamInterfaceManager.cs
@@ -331,17 +331,15 @@ internal sealed class DreamInterfaceManager : IDreamInterfaceManager {
             string elementId = split[0];
 
             // ":[element]" returns the default element of that type
-            if (elementId[0] == ':') {
-                switch (elementId[1..]) {
-                    case "map":
-                        return DefaultMap;
-                    case "info":
-                        return DefaultInfo;
-                    case "window":
-                        return DefaultWindow;
-                    case "output":
-                        return DefaultOutput;
-                }
+            switch (elementId) {
+                case ":map":
+                    return DefaultMap;
+                case ":info":
+                    return DefaultInfo;
+                case ":window":
+                    return DefaultWindow;
+                case ":output":
+                    return DefaultOutput;
             }
 
             foreach (ControlWindow window in Windows.Values) {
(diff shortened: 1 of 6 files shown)


OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -302,7 +302,7 @@ public void FrameUpdate(FrameEventArgs frameEventArgs) {
             DefaultMap.Viewport.Eye = _eyeManager.CurrentEye;
     }
 
-    public InterfaceElement? FindElementWithId(string id) {
+    public InterfaceElement? FindElementWithId(string id, bool selectDefault = false) {
```
wixoaGit, line 305: There are a lot of calls to `FindElementWithId()` that would need to explicitly set `selectDefault`. I think it's better to instead have the colon be a part of `id` so it works more universally.

wixoaGit (changes requested):

OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -330,6 +330,20 @@ public void FrameUpdate(FrameEventArgs frameEventArgs) {
         } else {
             string elementId = split[0];
 
+            // ":[element]" returns the default element of that type
+            if (elementId[0] == ':') {
+                switch (elementId[1..]) {
```
wixoaGit, line 335: You were worried about the string operation, right? You could just put `:` at the beginning of each case instead

wixoaGit (approved):

PR #1961, opened 2024-09-09
Add `name` to `IconAppearance`, fix placeholder turf names

### LINKED ISSUE #1960: Context menu shows a placeholder name for turfs
Turfs are supported in the right-click context menu as of https://github.com/OpenDreamProject/OpenDream/pull/1951. However, all turfs use a placeholder name of `turf (OD WIP)`. To resolve this, the `Name` field needs to be re-added to `IconAppearance` so clients are aware of turf names.

ike709:
Resolves #1960 

As far as I can tell it seems to work?

image

### DIFF
--- a/OpenDreamRuntime/AtomManager.cs
+++ b/OpenDreamRuntime/AtomManager.cs
@@ -243,7 +243,8 @@ public sealed class AtomManager {
     public void SetAppearanceVar(IconAppearance appearance, string varName, DreamValue value) {
         switch (varName) {
             case "name":
-                value.TryGetValueAsString(out appearance.Name);
+                value.TryGetValueAsString(out var name);
+                appearance.Name = name ?? string.Empty;
                 break;
             case "icon":
                 if (_resourceManager.TryLoadIcon(value, out var icon)) {
--- a/OpenDreamRuntime/AtomManager.cs
+++ b/OpenDreamRuntime/AtomManager.cs
@@ -208,6 +208,7 @@ public sealed class AtomManager {
 
     public bool IsValidAppearanceVar(string name) {
         switch (name) {
+            case "name":
             case "icon":
             case "icon_state":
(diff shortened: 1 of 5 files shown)


OpenDreamRuntime/AtomManager.cs
```diff
@@ -241,6 +242,9 @@ public bool IsValidAppearanceVar(string name) {
 
     public void SetAppearanceVar(IconAppearance appearance, string varName, DreamValue value) {
         switch (varName) {
+            case "name":
+                value.TryGetValueAsString(out appearance.Name);
```
wixoaGit, line 246: This will set `appearance.Name` to `null` if the value isn't a string.

OpenDreamShared/Dream/IconAppearance.cs
```diff
@@ -58,6 +69,7 @@ public sealed class IconAppearance : IEquatable<IconAppearance> {
     ];
 
     public IconAppearance() {
+        Name = string.Empty;
```
wixoaGit, line 72: You should follow the pattern and put this in the field definition instead (`public string Name = string.Empty`)

OpenDreamShared/Network/Messages/MsgAllAppearances.cs
```diff
@@ -217,6 +221,9 @@ public override void WriteToBuffer(NetOutgoingMessage buffer, IRobustSerializer
 
             lastId = pair.Key;
 
+            buffer.Write((byte)Property.Name);
+            buffer.Write(appearance.Name);
```
wixoaGit, line 225: You could avoid writing this property if it's equal to `IconAppearance.Default.Name`

wixoaGit (changes requested):

wixoaGit (approved):

PR #1962, opened 2024-09-12
Fix `Say`/`Me` DMF button command parsing

ike709:
The code is kind of cursed but it works and `Say`/`Me` now behave identically to BYOND.

Here's the Paradise DMF command for the `Say` button:
`command = ".winset \"saybutton.is-checked=true?input.command=\"!say \\\"\"saybutton.is-checked=true?mebutton.is-checked=false\""`

And the Paradise `Me` button:
``command = ".winset \"mebutton.is-checked=true ? input.command=\"!me \\\"\" : input.command=\"\"mebutton.is-checked=true ? saybutton.is-checked=false\""``

I am pretty sure BYOND does not allow multiple true statements or multiple false statements in ternaries. Otherwise, I have no idea how BYOND is deciding that the `Say` ternary's true statement is over and the second winset has begun.

https://github.com/user-attachments/assets/5d72df97-5e6d-4bb4-b25e-5d3b463d1f84

I also tested it on TG but couldn't be bothered to record a second video. They needed special handling because they appear to use `""` to delimit one of their ternaries. This is TG's `Say` button (note the `\"\"`):
`command = ".winset \"saybutton.is-checked=true ? input.command=\"!say \\\"\" : input.command=\"\"saybutton.is-checked=true ? mebutton.is-checked=false\"\"saybutton.is-checked=true ? oocbutton.is-checked=false\""`


ike709:
I have no idea if this breaks anything else in the DMF but I didn't notice any errors and nothing seemed broken while I was testing.

ike709:
This is not entirely correct, as apparently DMF ternaries *can* have multiple true/false statements which are delimited by apparently either whitespace or semicolons. Those will not work with this PR.

However, the only place I've seen those used are in goon's DMF in places that don't function currently anyways.

So either this PR can be merged as-is and fix some DMF functionality for Paradise & TG without breaking any *additional* things in goon, or it can be closed pending a much larger DMF refactor that I have no actual desire to code myself.

amylizzle:
so this is definitely wrong, I think maybe you've misinterpreted the winsets. AFAIK you can as many true statements in a ternary as you want, but only one false statement. 

The ones you've shown here are `condition ? input.command=value` where `value` is `\"!me \\\"\" : input.command=\"\"mebutton.is-checked=true ? saybutton.is-checked=false\""` as in the escaped winset for *another* ternary.

PR #1973, opened 2024-09-18
Remove all `OpenDreamShared` and `Content.Tests` warnings

ike709:
See title. Also elevates warnings to errors in all of the projects that don't currently have warnings.

### DIFF
--- a/Content.Tests/DMTests.cs
+++ b/Content.Tests/DMTests.cs
@@ -111,7 +111,7 @@ public sealed class DMTests : ContentUnitTest {
 
         DreamThread.Run("RunTest", async (state) => {
             if (_objectTree.TryGetGlobalProc("RunTest", out DreamProc? proc)) {
-                callTask = state.Call(proc, null, null);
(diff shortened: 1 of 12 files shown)


Content.Tests/DMTests.cs
```diff
@@ -111,7 +111,7 @@ public void TestFiles(string sourceFile, DMTestFlags testFlags) {
 
         DreamThread.Run("RunTest", async (state) => {
             if (_objectTree.TryGetGlobalProc("RunTest", out DreamProc? proc)) {
-                callTask = state.Call(proc, null, null);
+                callTask = state.Call(proc, _dreamMan.WorldInstance, null);
```
wixoaGit, line 114: `/proc/RunTest()` is a global proc so giving it `world` as its `src` would be incorrect. The method instead needs this argument marked as nullable.

wixoaGit (changes requested):

wixoaGit (approved):

PR #1978, opened 2024-09-19
Fix setting a static var to a const var

ike709:
See title. Brings Baystation from 211 errors down to 15.

I'm not entirely confident that this doesn't break things, as var parsing is a bit of a convoluted mess. I did confirm that this didn't generate any *new* errors for Baystation, and I tested that Paradise still seems to be playable.

### DIFF
--- a/DMCompiler/DM/Builders/DMExpressionBuilder.cs
+++ b/DMCompiler/DM/Builders/DMExpressionBuilder.cs
@@ -404,7 +404,11 @@ internal static class DMExpressionBuilder {
 
                     var field = dmObject?.GetVariable(name);
                     if (field != null) {
-                        return new Field(identifier.Location, field, field.ValType);
+                        if (field.IsConst)
+                            return new Field(identifier.Location, field, field.ValType);
+
+                        return BadExpression(WarningCode.BadExpression, identifier.Location,
+                            "Var \"{name}\" cannot be used in this context");
                     }
                 }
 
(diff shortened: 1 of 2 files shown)


DMCompiler/DM/Builders/DMExpressionBuilder.cs
```diff
@@ -401,6 +401,11 @@ private static DMExpression BuildIdentifier(DMASTIdentifier identifier, DMObject
                         var global = new GlobalField(identifier.Location, globalVar.Type, globalId.Value, globalVar.ValType);
                         return global;
                     }
+
+                    var field = dmObject?.GetVariable(name);
+                    if (field != null) {
+                        return new Field(identifier.Location, field, field.ValType);
+                    }
```
wixoaGit, line 408: This allows you to also reference a non-static var.
```dm
/obj/var/static/b = a
/obj/var/a = 1
```
This needs to verify that it's constant.
```suggestion
                    var field = dmObject?.GetVariable(name);
                    if (field != null) {
                        if (field.IsConst)
                            return new Field(identifier.Location, field, field.ValType);

                        return BadExpression(WarningCode.BadExpression, identifier.Location,
                            "Var \"{name}\" cannot be used in this context");
                    }
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #1981, opened 2024-09-20
Resolve proc override paths in var values

ike709:
Gets bay compiling when some of my other PRs are also merged.

### DIFF
--- a/Content.Tests/DMProject/Tests/Procs/procpath_without_proc_element.dm
+++ b/Content.Tests/DMProject/Tests/Procs/procpath_without_proc_element.dm
@@ -1,0 +1,11 @@
+// RETURN TRUE
+
+// This test confirms we don't OD0404 if a var is set to a procpath that does not contain the "/proc/" element
+
+/atom/movable/proc/foo()
+	return
+
+/datum/var/bar = /atom/movable/foo 
+
+/proc/RunTest()
+	return TRUE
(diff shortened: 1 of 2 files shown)


Content.Tests/DMProject/Tests/Procs/procpath_without_proc_element.dm
```diff
@@ -0,0 +1,11 @@
+// RETURN TRUE
+
+// This test confirms we don't OD0404 if a var is set to a procpath that does not contain the "/proc/" element
+
+/atom/movable/proc/foo()
+	return
+
+/datum/var/bar = /atom/movable/foo
```
out-of-phaze, line 8: image
this errors in DM, it's invalid (tries to find a type rather than a proc)

in 515 and later the proper way to call the latest override (lateral or descended) is via `call(nameof(/atom/movable/proc/foo))()` (or `call(nameof(/atom/movable.proc/foo))()`, i don't know why that one is sometimes used? i think it's just a legacy thing)

out-of-phaze (changes requested):
this seems to kind of be based on faulty assumptions. is baystation even on 515 yet?

Content.Tests/DMProject/Tests/Procs/procpath_without_proc_element.dm
ike709: Repeating what I said on discord a while back (to keep any PR observers informed), this does work if it's a proc override. My mistake was not erroring when there's no proc override declared.

PR #1982, opened 2024-09-20
Fix an issue with deref operations

### LINKED ISSUE #1977: Unknown identifier error in Baystation
Applies to Baystation.

Error:
`Error OD0404 at code/modules/mob/gender.dm:13:32: Unknown identifier "by_key"`

Code:
```
/hook/startup/proc/populate_gender_datum_list()
	for(var/type in subtypesof(/datum/gender))
		var/datum/gender/G = new type
		GLOB.gender_datums[G.key] = G
		if(!G.formal_term)
			G.formal_term = G.key

		var/datum/pronouns/P = GLOB.pronouns.by_key[G.default_pronouns] // This is the error
		GLOB.pronouns_from_gender[G.key] = P
		GLOB.pronouns_from_gender[G] = P

	return 1
```

The var is declared here:
```
/datum/pronouns_manager
	var/static/list/datum/pronouns/instances = list()
	var/static/list/datum/pronouns/by_key = list()

/datum/pronouns_manager/New()
	instances.Cut()
	by_key.Cut()
	for (var/datum/pronouns/pronouns as anything in subtypesof(/datum/pronouns))
		pronouns = new pronouns
		instances += pronouns
		by_key[pronouns.key] = pronouns

GLOBAL_DATUM_INIT(pronouns, /datum/pronouns_manager, new)
```

At first I assumed this was an issue with `by_key` being declared later in compilation order, but moving that code to the start didn't resolve the issue. I'm not really sure what the cause is here.

ike709:
Resolves #1977 

`astOperationOffset` never reached `2`, which is the index of the `IndexOperation`. Instead it was trying to repeat the previous `FieldOperation` again.

### DIFF
--- a/DMCompiler/DM/Builders/DMExpressionBuilder.cs
+++ b/DMCompiler/DM/Builders/DMExpressionBuilder.cs
@@ -717,8 +717,7 @@ internal static class DMExpressionBuilder {
                             }
 
                             operations = new Dereference.Operation[newOperationCount];
-                            astOperationOffset = i + iteration + 1;
-                            iteration++;
+                            astOperationOffset += i + 1;
                             i = -1;
                             prevPath = property.Type;
                             pathIsFuzzy = prevPath == null;
--- a/DMCompiler/DM/Builders/DMExpressionBuilder.cs
+++ b/DMCompiler/DM/Builders/DMExpressionBuilder.cs
@@ -669,7 +669,6 @@ internal static class DMExpressionBuilder {
             }
         }
 
-        var iteration = 0; // Since we do cursed stuff to `i` for reasons(TM), we need to track the for loop's iterations elsewhere
         for (int i = 0; i < operations.Length; i++) {
             DMASTDereference.Operation astOperation = astOperations[i + astOperationOffset];
             Dereference.Operation operation;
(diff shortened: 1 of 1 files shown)


DMCompiler/DM/Builders/DMExpressionBuilder.cs
```diff
@@ -716,7 +717,8 @@ private static DMExpression BuildDereference(DMASTDereference deref, DMObject dm
                             }
 
                             operations = new Dereference.Operation[newOperationCount];
-                            astOperationOffset = i + 1;
+                            astOperationOffset = i + iteration + 1;
+                            iteration++;
```
wixoaGit, line 721: A simpler solution would be `astOperationOffset += i + 1`
```suggestion
                            astOperationOffset += i + 1;
```

DMCompiler/DM/Builders/DMExpressionBuilder.cs
```diff
@@ -669,6 +669,7 @@ private static DMExpression BuildDereference(DMASTDereference deref, DMObject dm
             }
         }
 
+        var iteration = 0; // Since we do cursed stuff to `i` for reasons(TM), we need to track the for loop's iterations elsewhere
```
wixoaGit, line 672:
```suggestion
```

wixoaGit (changes requested):

DMCompiler/DM/Builders/DMExpressionBuilder.cs
wixoaGit: Hold on, this appears to break /tg/ compile

wixoaGit (approved):

PR #1984, opened 2024-09-23
Handle `var/list/foo[]= value` in proc statements

ike709:
Apparently we already handled this case in `DefinitionParameter()` so I just copied what we were doing there.


Content.Tests/DMProject/Tests/List/DoubleSquareBracketEquals.dm
```diff
@@ -0,0 +1,3 @@
+/proc/RunTest()
+	var/foo[] = new()
```
ike709, line 2:
```suggestion
	var/foo[]=new()
```

DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -286,6 +286,10 @@ public DMASTFile File() {
                     if (Check(TokenType.DM_Equals)) {
                         if (value != null) Warning("List doubly initialized");
 
+                        Whitespace();
+                        value = Expression();
+                        RequireExpression(ref value);
+                    } else if (Check(TokenType.DM_DoubleSquareBracketEquals)) {
```
wixoaGit, line 292: Shouldn't this also have the doubly-initialized warning that the above has?

wixoaGit (changes requested):

wixoaGit (approved):

PR #1985, opened 2024-09-23
Fixes animations being desynced

### LINKED ISSUE #1176: Duplicate icons play their frame animations out of sync
BYOND calculates the current frame using a global timer whereas we calculate it from the time the icon was set.

Cyberboss:
- Calculate every displayed frame from game tick zero rather than the last updated tick.
- Reduce loops required to catch up by calculating the module of the total animation delay.
- Fix `DirtyTexture()` being called too aggressively.

Fixes #1176


Cyberboss:
Everything resets when a new icon is loaded

wixoaGit (approved):

PR #1986, opened 2024-09-25
Fix `findtext()` with /regex

wixoaGit:
A change had made `findtext()` always return 0 when used with a /regex. Now it should properly run the regex.

This fixes the map loader on TG, which was improperly treating TGM-formatted maps as BYOND-formatted DMM maps due to this:
```dm
if(findtext(tfile, matches_tgm))
	map_format = MAP_TGM
else
	map_format = MAP_DMM
```


wixoaGit:
Needs a unit test

PR #1990, opened 2024-09-26
Fix issues found via Nebula testing

out-of-phaze:
Local proc variables that fall out of scope during the proc were not serialized due to an overeager optimization. This reverses that.

Due to using `out Result` instead of `out var result`, indexing a list set the proc return value. This changes the variable used to avoid that. ~~i see why some codebases require explicit `this`/`src` now~~

Due to setting alpha casting a `float` to a `byte`, the new alpha value was just truncated to the least significant byte of its integer conversion, meaning setting the alpha to `256` would set it to `0`, setting it to `355` would set it to `99`, etc. Now, it's clamped between 0 and 255 before the conversion.

### DIFF
--- a/OpenDreamRuntime/Procs/DMProc.cs
+++ b/OpenDreamRuntime/Procs/DMProc.cs
@@ -827,9 +827,9 @@ namespace OpenDreamRuntime.Procs {
                 case DMReference.Type.ListIndex: {
                     GetIndexReferenceValues(reference, out var index, out var indexing, peek);
 
-                    ProcStatus subState = GetIndex(indexing, index, this, out Result);
+                    ProcStatus subState = GetIndex(indexing, index, this, out var indexResult);
                     if (subState == ProcStatus.Continue)
-                        return Result;
+                        return indexResult;
                     else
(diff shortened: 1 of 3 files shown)


ZeWaka (commented):
The best part is, I'm pretty sure byond had that same issue with clamping and casting for a few things.

OpenDreamRuntime/Procs/DMProc.cs
```diff
@@ -827,9 +827,9 @@ public DreamValue GetReferenceValue(DreamReference reference, bool peek = false)
                 case DMReference.Type.ListIndex: {
                     GetIndexReferenceValues(reference, out var index, out var indexing, peek);
 
-                    ProcStatus subState = GetIndex(indexing, index, this, out Result);
+                    ProcStatus subState = GetIndex(indexing, index, this, out var result);
```
amylizzle, line 830: maybe best to change the var name entirely? `indexResult` or something?

also whoops :trollface:

wixoaGit: Would be a good change

wixoaGit (approved):

PR #2000, opened 2024-09-29
Add support for NT Script alongside DM

wixoaGit:
Dream Makers shouldn't be constrained to just one language. They should be given a plethora of options, so they can choose the best tool for creating their Dreams. This PR is a step in that direction, allowing you to write [NT Script](https://tgstation13.org/wiki/NT_Script) and include it in your game the same as you would DM.

All you need to do is `#include` a file with the extension `.ntsl` and it will be included in the global DM scope.

For example, NT Script running in Paradise:
image
image


Laboredih123:
can you feel the dreams

Iamgoofball:
@mrstonedone

MrStonedOne:
@gunhog

AffectedArc07:
God this terrifies me but makes me want to bring back my old tcomms shenanigans

wixoaGit:
Nevermind, I have decided the people must stay under DM's oppressive thumb.

~~Stay tuned for next time when I remove any possibility of using C# as well~~

Laboredih123:
im crrying

PR #2007, opened 2024-09-30
fix null statpanel

amylizzle:
bug, makes joining goon real slow for some reason, throws a bunch of debug asserts.


OpenDreamRuntime/DreamConnection.cs
```diff
@@ -96,6 +96,8 @@ public string SelectedStatPanel {
     public DreamConnection() {
         IoCManager.InjectDependencies(this);
 
+        _selectedStatPanel = "";
```
wixoaGit, line 99: An unmodified `/client.statpanel` var is `null` on byond. This makes the default an empty string on OD.

wixoaGit (changes requested):

wixoaGit (approved):

PR #2009, opened 2024-10-01
`COMPILE ERROR` unit tests now require a specific error code

ike709:
Unit tests that check for a compiler error must now specify the error code that they want. 

Example: `// COMPILE ERROR OD0404`

For the tests that do not explicitly elevate a pragma, I got the error code by running the test in the OD compiler bot (while renaming `RunTest()` to `main()`) and seeing what got emitted. This means that if any test is wildly broken currently, this PR doesn't really fix that.

I recommend reviewing it by commit, they're split up pretty nicely.

### DIFF
--- a/Content.Tests/DMTests.cs
+++ b/Content.Tests/DMTests.cs
@@ -179,7 +179,7 @@ public sealed partial class DMTests : ContentUnitTest {
                 testFlags |= DMTestFlags.Ignore;
             if (firstLine.Contains("COMPILE ERROR", StringComparison.InvariantCulture)) {
                 testFlags |= DMTestFlags.CompileError;
-                Match match = ErrorCodeRegex().Match(firstLine);  // "OD" followed by exactly 4 characters
+                Match match = ErrorCodeRegex().Match(firstLine);  // "OD" followed by exactly 4 numbers
                 if (!match.Success) {
                     Console.WriteLine($"\"COMPILE ERROR\" test \"{sourceFile}\" does not specify an error code");
                 } else {
--- a/Content.Tests/DMTests.cs
+++ b/Content.Tests/DMTests.cs
@@ -64,19 +64,16 @@ public sealed partial class DMTests : ContentUnitTest {
         string initialDirectory = Directory.GetCurrentDirectory();
         TestContext.WriteLine($"--- TEST {sourceFile} | Flags: {testFlags}");
         try {
+            string? compiledFile = Compile(Path.Join(initialDirectory, TestsDirectory, sourceFile));
             if (testFlags.HasFlag(DMTestFlags.CompileError)) {
                 Assert.That(errorCode == -1, Is.False, "Expected an error code");
-                string? compileErrorFile = Compile(Path.Join(initialDirectory, TestsDirectory, sourceFile));
-
                 Assert.That(DMCompiler.DMCompiler.UniqueEmissions.Contains((WarningCode)errorCode), Is.True, $"Expected error code \"{errorCode}\" was not found");
-                Assert.That(compileErrorFile, Is.Null, "Expected an error during DM compilation");
+                Assert.That(compiledFile, Is.Null, "Expected an error during DM compilation");
 
-                Cleanup(compileErrorFile);
+                Cleanup(compiledFile);
                 TestContext.WriteLine($"--- PASS {sourceFile}");
                 return;
             }
-
-            string? compiledFile = Compile(Path.Join(initialDirectory, TestsDirectory, sourceFile));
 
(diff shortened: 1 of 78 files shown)


Content.Tests/DMTests.cs
```diff
@@ -176,6 +198,9 @@ private static DMTestFlags GetDMTestFlags(string sourceFile) {
         return testFlags;
     }
 
+    [GeneratedRegex(@"OD.{4}")]
```
wixoaGit, line 201: Should probably just accept numbers
```suggestion
    [GeneratedRegex(@"OD[0-9]{4}")]
```

Content.Tests/DMTests.cs
```diff
@@ -143,28 +155,38 @@ private static IEnumerable<object[]> GetTests()
...
         using (StreamReader reader = new StreamReader(sourceFile)) {
             string? firstLine = reader.ReadLine();
             if (firstLine == null)
                 return testFlags;
             if (firstLine.Contains("IGNORE", StringComparison.InvariantCulture))
                 testFlags |= DMTestFlags.Ignore;
-            if (firstLine.Contains("COMPILE ERROR", StringComparison.InvariantCulture))
+            if (firstLine.Contains("COMPILE ERROR", StringComparison.InvariantCulture)) {
                 testFlags |= DMTestFlags.CompileError;
+                Match match = ErrorCodeRegex().Match(firstLine);  // "OD" followed by exactly 4 characters
```
wixoaGit, line 182:
```suggestion
                Match match = ErrorCodeRegex().Match(firstLine);  // "OD" followed by exactly 4 numbers
```

Content.Tests/DMTests.cs
```diff
@@ -143,28 +155,38 @@ private static IEnumerable<object[]> GetTests()
...
+                if (!match.Success) {
+                    Console.WriteLine($"\"COMPILE ERROR\" test \"{sourceFile}\" does not specify an error code");
```
wixoaGit, line 184: Is there a point to this print? Is it visible anywhere it makes sense?

Content.Tests/DMTests.cs
```diff
@@ -58,18 +59,29 @@ private static void Cleanup(string? compiledFile) {
...
-                Cleanup(compiledFile);
+                Assert.That(string.IsNullOrEmpty(errorCode), Is.False, "Expected an error code");
+                var originalConsoleOut = Console.Out;
+                StringWriter consoleOutput = new StringWriter();
+                Console.SetOut(consoleOutput);
+                string? compileErrorFile = Compile(Path.Join(initialDirectory, TestsDirectory, sourceFile));
+                string output = consoleOutput.ToString();
+                Console.SetOut(originalConsoleOut);
+                Console.WriteLine(output);
+                Assert.That(output.Contains(errorCode), Is.True, $"Expected error code \"{errorCode}\" was not found");
```
wixoaGit, line 75: Instead of checking the console output for `OD1234` could there instead be a HashSet on DMCompiler tracking which emission codes have been made? It could also be used to ensure no other codes have been raised either.

wixoaGit (changes requested):

DMCompiler/DMCompiler.cs
```diff
@@ -21,6 +21,7 @@ namespace DMCompiler;
 public static class DMCompiler {
     public static int ErrorCount;
     public static int WarningCount;
+    public static HashSet<WarningCode> UniqueEmissions = new();
```
wixoaGit, line 24: This needs cleared on every `Compile()` call

Content.Tests/DMTests.cs
```diff
@@ -58,18 +60,24 @@ private static void Cleanup(string? compiledFile) {
...
+    public void TestFiles(string sourceFile, DMTestFlags testFlags, int errorCode) {
         string initialDirectory = Directory.GetCurrentDirectory();
         TestContext.WriteLine($"--- TEST {sourceFile} | Flags: {testFlags}");
         try {
-            string? compiledFile = Compile(Path.Join(initialDirectory, TestsDirectory, sourceFile));
             if (testFlags.HasFlag(DMTestFlags.CompileError)) {
-                Assert.That(compiledFile, Is.Null, "Expected an error during DM compilation");
-                Cleanup(compiledFile);
+                Assert.That(errorCode == -1, Is.False, "Expected an error code");
+                string? compileErrorFile = Compile(Path.Join(initialDirectory, TestsDirectory, sourceFile));
```
wixoaGit, line 69: This doesn't still need to be moved into here and copied below.

wixoaGit (changes requested):

wixoaGit (approved):

PR #2010, opened 2024-10-02
check if sounds are deleted

amylizzle:
this was basically only affecting me, but still


wixoaGit:
I can confirm that this error is spamming my client's console as well. Doesn't crash though.

OpenDreamClient/Audio/DreamSoundEngine.cs
```diff
@@ -31,7 +32,7 @@ public void Initialize() {
 
     public void StopFinishedChannels() {
         for (int i = 0; i < SoundChannelLimit; i++) {
-            if (_channels[i]?.Source.Component.Playing is false or null)
+            if (_channels[i] is not null && (!_entityManager.Deleted(_channels[i]!.Source.Entity) && !_entityManager.IsQueuedForDeletion(_channels[i]!.Source.Entity) && _channels[i]!.Source.Component.Playing is false))
```
wixoaGit, line 35: `StopFinishedChannels()` was needed before RT refactored audio to be entities. Maybe RT is now deleting the entity when the sound finishes and this is no longer necessary?

The channel would still need to be nulled out.

wixoaGit (approved):

PR #2019, opened 2024-10-06
Implement `world.sleep_offline`

### LINKED ISSUE #194: `world/sleep_offline` not implemented
http://www.byond.com/docs/ref/#/world/var/sleep_offline

NTstation

ike709:
Also handles `world.tick_lag` being set on the object definition.

`world.sleep_offline` works slightly differently from BYOND. According to the ref, if `sleep_offline = 0` then the server pauses *if there are no sleeping procs*. Setting `sleep_offline = 1` causes it to still pause even if there *are* sleeping procs.

This PR just makes `world.sleep_offline = 1` set `CVars.GameAutoPauseEmpty` to `true`. Probably good enough for now.

Resolves #194

### DIFF
--- a/OpenDreamRuntime/EntryPoint.cs
+++ b/OpenDreamRuntime/EntryPoint.cs
@@ -42,7 +42,7 @@ namespace OpenDreamRuntime {
             componentFactory.GenerateNetIds();
 
             _configManager.OverrideDefault(CVars.NetLogLateMsg, false); // Disable since disabling prediction causes timing errors otherwise.
-            _configManager.OverrideDefault(CVars.GameAutoPauseEmpty, false); // TODO: world.sleep_offline can control this
+
             _configManager.OverrideDefault(CVars.DiscordRichPresenceSecondIconId, "opendream");
             _configManager.SetCVar(CVars.GridSplitting, false); // Grid splitting should never be used
             if(String.IsNullOrEmpty(_configManager.GetCVar<string>(OpenDreamCVars.JsonPath))) //if you haven't set the jsonpath cvar, set it to the first valid file path passed as an arg
--- a/OpenDreamRuntime/Objects/Types/DreamObjectWorld.cs
+++ b/OpenDreamRuntime/Objects/Types/DreamObjectWorld.cs
@@ -64,8 +64,10 @@ public sealed class DreamObjectWorld : DreamObject {
         base(objectDefinition) {
         IoCManager.InjectDependencies(this);
 
+        SetTicklag(objectDefinition.Variables["tick_lag"]);
         SetLog(objectDefinition.Variables["log"]);
         SetFps(objectDefinition.Variables["fps"]);
+        SetSleepOffline(objectDefinition.Variables["sleep_offline"]);
 
         DreamValue view = objectDefinition.Variables["view"];
         if (view.TryGetValueAsString(out var viewString)) {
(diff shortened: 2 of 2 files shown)


OpenDreamRuntime/Objects/Types/DreamObjectWorld.cs
```diff
@@ -64,8 +64,10 @@ public DreamObjectWorld(DreamObjectDefinition objectDefinition) :
         base(objectDefinition) {
         IoCManager.InjectDependencies(this);
 
+        SetTicklag(objectDefinition.Variables["tick_lag"]);
         SetLog(objectDefinition.Variables["log"]);
         SetFps(objectDefinition.Variables["fps"]);
```
wixoaGit, line 69: If a project sets `world.tick_lag` but not `world.fps` this would lead to the default fps value overriding it.

OpenDreamRuntime/Objects/Types/DreamObjectWorld.cs
```diff
@@ -239,14 +241,18 @@ protected override void SetVar(string varName, DreamValue value) {
...
-            case "sleep_offline":
             case "status":
             case "version":
             case "visibility":
                 // Set it in the var dictionary, so reading at least gives the same value
                 base.SetVar(varName, value);
                 break;
 
+            case "sleep_offline":
+                base.SetVar(varName, value); // Only 1 and not-1 matter, but set it in case any user code cares what the actual value is
```
wixoaGit, line 252: BYOND is actually always giving 0 when trying to get `world.sleep_offline`

image
image

OpenDreamRuntime/EntryPoint.cs
```diff
@@ -42,7 +42,7 @@ public override void Init() {
             componentFactory.GenerateNetIds();
 
             _configManager.OverrideDefault(CVars.NetLogLateMsg, false); // Disable since disabling prediction causes timing errors otherwise.
-            _configManager.OverrideDefault(CVars.GameAutoPauseEmpty, false); // TODO: world.sleep_offline can control this
+
```
wixoaGit, line 45:
```suggestion
```

wixoaGit (changes requested):

OpenDreamRuntime/Objects/Types/DreamObjectWorld.cs
```diff
@@ -314,4 +317,20 @@ private void SetFps(DreamValue fps) {
...
+
+    private void SetTicklag(DreamValue value) {
+        if (!value.TryGetValueAsFloat(out var tickLag))
+            tickLag = 1; // An invalid tick_lag gets turned into 1
+
+        TickLag = tickLag;
+    }
+
+    private void SetSleepOffline(DreamValue sleepOffline) {
+        if (sleepOffline.TryGetValueAsInteger(out var value) && value == 1) {
```
wixoaGit, line 329: Truthy values are treated as true here.
```suggestion
        if (sleepOffline.IsTruthy()) {
```

OpenDreamRuntime/Objects/Types/DreamObjectWorld.cs
wixoaGit: Turns out BYOND is erroneously giving the opposite value (`0` for on and `1` for off). We should return the correct 0/1 value here.

PR #2023, opened 2024-10-07
Freeze some collections for faster accessing

ike709:
Converts some collections on `DreamObjectTree` and `DreamObjectDefinition` to their frozen variants. On my machine this consistently reduced Paradise round init by roughly ~2 seconds. I only compared third and fourth runs for both before & after so PGO wouldn't impact the results.

Since these collections never change after JSON parsing, it's a free performance boost on any accesses for the remainder of the round.

I'll try to tackle Procs & Variables in a subsequent PR since they need a bit more wrangling.


ike709:
Superseded by https://github.com/OpenDreamProject/OpenDream/pull/2031

PR #2030, opened 2024-10-14
disable lint check on draft prs

harryob:
no sarif annotations while drafted


harryob:
likely redundant - for future reference, you can toggle annotations by pressing `A` in the files screen

PR #2032, opened 2024-10-15
Halves `OpenDreamClient` warnings

ike709:
Takes `OpenDreamClient` from 40 to 20 warnings.

I confirmed that the changes to gliding and the context menu didn't break them.

### DIFF
--- a/OpenDreamClient/Interface/DMF/IDMFProperty.cs
+++ b/OpenDreamClient/Interface/DMF/IDMFProperty.cs
@@ -222,38 +222,20 @@ public struct DMFPropertyVec2 : IDMFProperty, IEquatable<DMFPropertyVec2> {
         return comparisonVec.X == X && comparisonVec.Y == Y;
     }
 
-    public bool Equals(DMFPropertyVec2 other)
-    {
+    public bool Equals(DMFPropertyVec2 other) {
         return X == other.X && Y == other.Y && Delim == other.Delim;
     }
 
-    public override bool Equals(object? obj)
-    {
+    public override bool Equals(object? obj) {
         return obj is DMFPropertyVec2 other && Equals(other);
     }
 
-    public override int GetHashCode()
-    {
+    public override int GetHashCode() {
         return HashCode.Combine(X, Y, Delim);
     }
 }
 
 public struct DMFPropertySize : IDMFProperty, IEquatable<DMFPropertySize> {
-    public bool Equals(DMFPropertySize other)
-    {
-        return _value.Equals(other._value);
-    }
-
-    public override bool Equals(object? obj)
-    {
-        return obj is DMFPropertySize other && Equals(other);
-    }
-
-    public override int GetHashCode()
-    {
-        return _value.GetHashCode();
-    }
-
     private DMFPropertyVec2 _value;
     public int X {get => _value.X; set => _value.X = value;}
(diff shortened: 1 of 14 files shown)


OpenDreamClient/Interface/InterfaceElement.cs
```diff
@@ -42,7 +42,9 @@
 
             MappingDataNode newNode = original.Merge(node);
 
-            ElementDescriptor = (ElementDescriptor)serializationManager.Read(ElementDescriptor.GetType(), newNode);
+            var descriptor = serializationManager.Read(ElementDescriptor.GetType(), newNode);
+            if(descriptor is null) throw new NullReferenceException(); // We're in a try/catch anyway, just play it safe
```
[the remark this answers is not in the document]
ike709: I think complying with this results in yuckier-looking code in this case so I'll leave it to wixoa to decide.

OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -891,21 +891,22 @@ private void LoadInterface(InterfaceDescriptor descriptor) {
     }
 
     private void LoadDescriptor(ElementDescriptor descriptor) {
+        // Note: Id has a getter to ensure Id.Value is never null
```
wixoaGit, line 894: I don't see why `Value` is marked nullable in the first place. The value isn't nullable in the constructor, and nothing ever gives it `null`.

OpenDreamClient/Interface/DMF/IDMFProperty.cs
```diff
@@ -226,9 +226,39 @@ public bool Equals(string comparison) {
         DMFPropertyVec2 comparisonVec = new(comparison);
         return comparisonVec.X == X && comparisonVec.Y == Y;
     }
+
+    public bool Equals(DMFPropertyVec2 other)
```
wixoaGit, line 230: All these new methods need their braces moved to the same line.

OpenDreamClient/Interface/DMF/IDMFProperty.cs
```diff
@@ -226,9 +226,39 @@ public bool Equals(string comparison) {
...
+
+    public override bool Equals(object? obj)
+    {
+        return obj is DMFPropertySize other && Equals(other);
+    }
+
+    public override int GetHashCode()
+    {
+        return _value.GetHashCode();
+    }
```
wixoaGit, line 260: Move these to the bottom

OpenDreamClient/Interface/DMF/IDMFProperty.cs
```diff
@@ -299,7 +329,22 @@ public bool Equals(string comparison) {
...
+
+    public override bool Equals(object? obj)
+    {
+        return obj is DMFPropertyPos other && Equals(other);
+    }
+
+    public override int GetHashCode()
+    {
+        return _value.GetHashCode();
+    }
```
wixoaGit, line 346: Move these to the bottom

OpenDreamClient/Interface/InterfaceElement.cs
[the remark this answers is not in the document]
wixoaGit: I'm fine with it as is.

wixoaGit (changes requested):

wixoaGit (approved):

PR #2039, opened 2024-10-17
Run UI controls in the background

wixoaGit:
In BYOND non-visible browser controls on secondary windows still run & execute javascript in the background. We would only start running browser controls from the first time their window was shown.

Fixing this gets tg's say window working.
image

Depends on https://github.com/space-wizards/RobustToolbox/pull/5489


OpenDreamClient/Interface/Controls/ControlWindow.cs
```diff
@@ -178,16 +179,8 @@ private void UpdateWindowAttributes((OSWindow? osWindow, IClydeWindow? clydeWind
 
         //if our window is null or closed, and we are visible, we need to create a new one. Otherwise we need to update the existing one.
```
wixoaGit, line 180:
```suggestion
        //if our window is null or closed, we need to create a new one. Otherwise we need to update the existing one.
```

PR #2043, opened 2024-10-18
Add a unit test for `\L`

ike709:


Content.Tests/DMProject/Tests/Regex/regex_L_brackets.dm
```diff
@@ -0,0 +1,5 @@
+
+/proc/RunTest()
+	var/static/regex/repeated_consonant_regex = regex(@"\b([^aeiou\L])\1", "gi")
+	var/words = "hey ggggurl"
+	ASSERT(repeated_consonant_regex.Find(words) != 0)
```
ike709, line 5:
```suggestion
	ASSERT(repeated_consonant_regex.Find(words) == 5)
```

Content.Tests/DMProject/Tests/Regex/regex_L_brackets.dm
```diff
@@ -0,0 +1,5 @@
+
+/proc/RunTest()
+	var/static/regex/repeated_consonant_regex = regex(@"\b([^aeiou\L])\1", "gi")
+	var/words = "hey ggggurl"
+	ASSERT(repeated_consonant_regex.Find(words) == 5)
```
wixoaGit, line 5: This would match the same regardless of if `\L` wasn't in the regex. Add something that changes the outcome depending on if `\L` is there.
```suggestion
	var/words = "hey... ggggurl"
	ASSERT(repeated_consonant_regex.Find(words) == 8)
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #2045, opened 2024-10-18
Apply UIScale to ControlWindow size

ike709:
Fixes the size of various UIs like TG's Say popup.

### DIFF
--- a/OpenDreamClient/Interface/Controls/ControlWindow.cs
+++ b/OpenDreamClient/Interface/Controls/ControlWindow.cs
@@ -340,7 +340,7 @@ public sealed class ControlWindow : InterfaceControl {
             case "size":
                 if (_myWindow.osWindow is {ClydeWindow: not null}) {
                     var size = new DMFPropertySize(value);
-                    var uiScale = _configurationManager.GetCVar(CVars.DisplayUIScale);
+                    var uiScale = _myWindow.osWindow.UIScale;
                     size.X = (int)(size.X * uiScale); // TODO: RT should probably do this itself
                     size.Y = (int)(size.Y * uiScale);
                     _myWindow.osWindow.ClydeWindow.Size = size.Vector;
--- a/OpenDreamClient/Interface/Controls/ControlWindow.cs
+++ b/OpenDreamClient/Interface/Controls/ControlWindow.cs
@@ -4,12 +4,15 @@ using OpenDreamClient.Interface.DMF;
 using Robust.Client.Graphics;
 using Robust.Client.UserInterface;
 using Robust.Client.UserInterface.Controls;
+using Robust.Shared;
+using Robust.Shared.Configuration;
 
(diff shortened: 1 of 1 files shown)


OpenDreamClient/Interface/Controls/ControlWindow.cs
```diff
@@ -336,7 +339,11 @@ public override void SetProperty(string property, string value, bool manualWinse
         switch (property) {
             case "size":
                 if (_myWindow.osWindow is {ClydeWindow: not null}) {
-                    _myWindow.osWindow.ClydeWindow.Size = new DMFPropertySize(value).Vector;
+                    var size = new DMFPropertySize(value);
+                    var uiScale = _configurationManager.GetCVar(CVars.DisplayUIScale);
+                    size.X = (int)(size.X * uiScale);
```
wixoaGit, line 344: This is probably something better left for RT to do automatically, but we can just TODO that
```suggestion
                    size.X = (int)(size.X * uiScale); // TODO: RT should probably do this itself
```

OpenDreamClient/Interface/Controls/ControlWindow.cs
```diff
@@ -336,7 +339,11 @@ public override void SetProperty(string property, string value, bool manualWinse
         switch (property) {
             case "size":
                 if (_myWindow.osWindow is {ClydeWindow: not null}) {
-                    _myWindow.osWindow.ClydeWindow.Size = new DMFPropertySize(value).Vector;
+                    var size = new DMFPropertySize(value);
+                    var uiScale = _configurationManager.GetCVar(CVars.DisplayUIScale);
```
wixoaGit, line 343: A few different things can affect the UI scale. This CVar is actually 0 by default, to indicate using your OS settings.
```suggestion
                    var uiScale = _myWindow.osWindow.UIScale;
```

wixoaGit (changes requested):

OpenDreamClient/Interface/Controls/ControlWindow.cs
```diff
@@ -4,12 +4,15 @@
...
 using Robust.Client.UserInterface.Controls;
+using Robust.Shared;
+using Robust.Shared.Configuration;
 
 namespace OpenDreamClient.Interface.Controls;
 
 public sealed class ControlWindow : InterfaceControl {
     [Dependency] private readonly IClyde _clyde = default!;
     [Dependency] private readonly IUserInterfaceManager _uiMgr = default!;
+    [Dependency] private readonly IConfigurationManager _configurationManager = default!;
```
ike709, line 15:
```suggestion

namespace OpenDreamClient.Interface.Controls;

public sealed class ControlWindow : InterfaceControl {
    [Dependency] private readonly IClyde _clyde = default!;
    [Dependency] private readonly IUserInterfaceManager _uiMgr = default!;
```

wixoaGit (approved):

PR #2049, opened 2024-10-20
`ReturnFloat` peephole optimization

ike709:
`PushFloat -> Return` is a common pattern in TG with 21,601 occurrences. This is mainly due to procs returning `TRUE` or `FALSE`.

This PR replaces that pattern with a new `ReturnFloat` opcode to avoid pushing to and immediately popping from the stack on all of those proc returns.


ike709:
test

LemonInTheDark:
😳

wixoaGit (approved):

PR #2050, opened 2024-10-20
Change the movable contents to special list, and make view include Contents

SaphireLattice:
New list implementation was required to make `+=` and such work. Not sure if `SetVar` would have helped? Actually didn't test if a reference to the list still updates the object on reference implementation, but it feels like it should. Then again, never know what weird shit might happen.

Also made `view(...)` include `contents` just like `range(...)` - these probably should share the implementation code, other than a boolean. They are pretty much the same, just with view culling.

Oh, this fixes:
- TGUI of worn stuff (via view(), mostly)
- Drones and probably borgs getting UI for RPD
- Fire extinguishers dropping under your mob when you put them into the cabinet, while also being referenced by the cabinet, leading to fun teleport shenanigans

Though, worrding that it's shitcode to have `public ChildEnumerator` and `public ChildCount` exposed by the `DreamObjectMovable`. Maybe I could just give the `MovableContentsList` its Transform component?


OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -3023,6 +3023,14 @@ public static DreamValue NativeProc_view(NativeProc.Bundle bundle, DreamObject?
         if (center is null)
             return new(view);
 
+        if (center.TryGetVariable("contents", out var centerContents) && centerContents.TryGetValueAsDreamList(out var centerContentsList)) {
+            foreach (var content in centerContentsList.GetValues()) {
+                view.AddValue(content);
+            }
```
ike709, line 3029:
```suggestion
             view.AddRange(centerContentsList.GetValues());
```
pretty sure this can just be an `AddRange()`?

OpenDreamRuntime/Objects/Types/DreamList.cs
```diff
@@ -1263,6 +1263,91 @@ public IEnumerable<DreamObjectTurf> GetTurfs() {
...
+
+                index--;
+            }
+        }
+
+        throw new Exception($"Out of bounds index on movable contents list after iterating: {key}");
+    }
+
+    public override List<DreamValue> GetValues() {
+        List<DreamValue> values = [];
```
ike709, line 1297:
```suggestion
        List<DreamValue> values = new List<DreamValue>(_movable.ChildCount);
```
Pre-allocate the list size

SaphireLattice:
Oh yeah, remembered another thing

I have *no damn clue* as to the stability of the iteration keys and trying to get items "by index" where it's just another iterator again and uhhh... yeah. It didn't seem to explode at least? Like, taking off stuff and moving things around in hands worked fine. Didn't check it with backpacks too much though.

OpenDreamRuntime/Objects/Types/DreamList.cs
```diff
@@ -1263,6 +1271,91 @@ public IEnumerable<DreamObjectTurf> GetTurfs() {
...
+
+    public override void SetValue(DreamValue key, DreamValue value, bool allowGrowth = false) {
+        throw new Exception("Cannot set an index of movable contents list");
+    }
+
+    public override void AddValue(DreamValue value) {
+        if (!value.TryGetValueAsDreamObject<DreamObjectMovable>(out var movable))
+            throw new Exception($"Cannot add {value} to movable contents");
+
+        movable.SetVariable("loc", new (_movable));
```
wixoaGit, line 1327: This should use the `SetLoc()` method on `movable` instead. Feel free to make it public.

OpenDreamRuntime/Objects/Types/DreamList.cs
```diff
@@ -142,6 +142,14 @@ public virtual void AddValue(DreamValue value) {
         _values.Add(value);
     }
 
+    public virtual void AddValueRange(DreamList valueRange) {
+        _values.AddRange(valueRange.GetValues());
+    }
+
+    public virtual void AddValueRange(IEnumerable<DreamValue> valueRange) {
+        _values.AddRange(valueRange);
+    }
```
wixoaGit, line 151: Adding helper functions like this that perform directly on `_values` should be avoided. These would have to be overridden on every DreamList type to handle their special behaviors.

OpenDreamRuntime/Objects/Types/DreamList.cs
```diff
@@ -1263,6 +1271,91 @@ public IEnumerable<DreamObjectTurf> GetTurfs() {
...
+    public override DreamValue GetValue(DreamValue key) {
+        if (!key.TryGetValueAsInteger(out var index))
+            throw new Exception($"Invalid index into movable contents list: {key}");
+
+
+        if (index < 1 || index > _movable.ChildCount)
+            throw new Exception($"Out of bounds index on movable contents list: {index}");
+
+
+        using (var childEnumerator = _movable.ChildEnumerator) {
```
wixoaGit, line 1286: It should be noted that `ChildEnumerator` is an enumerator of a `HashSet` which has no defined order. This means the order could theoretically change at any time, breaking this.

I don't think this is a new bug so it doesn't really need to be worried about in this PR.

OpenDreamRuntime/Objects/Types/DreamList.cs
```diff
@@ -1263,6 +1271,91 @@ public IEnumerable<DreamObjectTurf> GetTurfs() {
...
+            throw new Exception($"Cannot add {value} to movable contents");
+
+        movable.SetVariable("loc", new (_movable));
+    }
+
+    public override void RemoveValue(DreamValue value) {
+        if (!value.TryGetValueAsDreamObject<DreamObjectMovable>(out var movable))
+            throw new Exception($"Cannot remove {value} from movable contents");
+
+        movable.SetVariable("loc", DreamValue.Null);
```
wixoaGit, line 1334: This doesn't verify that this movable was even in our contents to begin with.

Also, use `SetLoc()` here as well.

OpenDreamRuntime/Objects/Types/DreamList.cs
```diff
@@ -1263,6 +1271,91 @@ public IEnumerable<DreamObjectTurf> GetTurfs() {
...
+        if (!value.TryGetValueAsDreamObject<DreamObjectMovable>(out var movable))
+            return false;
+
+        if (!movable.TryGetVariable("loc", out var _locVariable))
+            return false;
+
+        if (!_locVariable.TryGetValueAsDreamObject<DreamObjectAtom>(out var _loc))
+            return false;
+
+        return _loc == _movable;
```
wixoaGit, line 1347: `DreamObjectMovable.Loc` can be used instead of indirectly going through variables.
```suggestion
        return dreamObject.Loc == movable;
```

TestGame/code.dm
wixoaGit, line 1: Undo the changes to this.

OpenDreamRuntime/Objects/Types/DreamObjectMovable.cs
```diff
@@ -19,6 +19,10 @@ public class DreamObjectMovable : DreamObjectAtom {
 
     private readonly TransformComponent _transformComponent;
 
+    public readonly MovableContentsList Contents;
```
wixoaGit, line 22: Move this up with the other public properties

OpenDreamRuntime/Objects/Types/DreamObjectMovable.cs
```diff
@@ -19,6 +19,10 @@ public class DreamObjectMovable : DreamObjectAtom {
...
+
+    public TransformChildrenEnumerator ChildEnumerator => _transformComponent.ChildEnumerator;
+    public int ChildCount => _transformComponent.ChildCount;
```
wixoaGit, line 25: > Though, worrding that it's shitcode to have public ChildEnumerator and public ChildCount exposed by the DreamObjectMovable. Maybe I could just give the MovableContentsList its Transform component?

I would prefer for you to do this.

wixoaGit (changes requested):

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
wixoaGit: `view` is a DreamList which does not have any range-adding methods. Closest would be `OperatorAppend()` but that should probably be only be used by DM.

OpenDreamRuntime/Objects/Types/DreamList.cs
SaphireLattice: Oh forgot about that. Guess it would be a no-op if it isn't?

And thanks for telling about SetLoc, was not aware! Swear I've seen this pattern elsewhere.

wixoaGit (approved):

PR #2054, opened 2024-10-21
Peephole optimizer & proc decoder fixes

ike709:
This PR does a few things:

1. Fixes peephole optimizations not being applied to opcodes created by peephole optimizations. E.g.:
`PushNull -> Assign -> Pop` was being optimized to `PushNull -> AssignPop`, but the optimizer offset wasn't rolled back far enough to consider optimizing `PushNull -> AssignPop` to `NullRef`. It would only be rolled back to the new `AssignPop` opcode. This meant that the `NullRef` opcode was never used, which is now fixed.

2. Fixes a few opcodes not being implemented in `ProcDecoder`. Because a few opcodes were specific to optimizations that were never applied, they never triggered any errors in the disassembler. `DreamProcOpcode.NullRef` and `DreamProcOpcode.PushNOfStringFloats` can now be decoded.

3. Fixes `PushNOfStringFloats` only being applied to pairs of `StringFloat`s. Because of how the peephole optimizer works, it would (now that it actually runs) try to run the `PushNOfStringFloats` optimization before every `PushString` and `PushFloat` in the sequence had been converted to `StringFloat`s. The `PushNOfStringFloat`s optimization has been folded into the `StringFloat` optimization, and it'll just use the relevant opcode depending on how many pairs of `PushString` and `PushFloat` opcodes there are. This has the added benefit of reducing the amount of optimizations that need to be applied.

We really need to implement a system for unit testing the peephole optimizer on input bytecode, but that is out of scope for this PR.

I made sure that TG doesn't explode, the `NullRef` opcode is now used, and that the `PushNOfStringFormat` optimization actually works correctly & disassembles with sane formatting:
```
	PushFloat 0 
	PushNOfStringFloats "y" 0 "size" 1 "repeat" 2 "radius" 0 "falloff" 1 "flags" 0
	CreateAssociativeList 7 
```

### DIFF
--- a/DMCompiler/Optimizer/PeepholeOptimizations.cs
+++ b/DMCompiler/Optimizer/PeepholeOptimizations.cs
@@ -273,7 +273,6 @@ internal sealed class PushStringFloat : IPeepholeOptimization {
 
         int stackDelta = 0;
         List<IAnnotatedBytecode> args = new List<IAnnotatedBytecode>(2 * count + 1) { new AnnotatedBytecodeInteger(count, input[index].GetLocation()) };
-        args.Add(new AnnotatedBytecodeInteger(count, input[index].GetLocation()));
 
         for (int i = 0; i < count; i++) {
             AnnotatedBytecodeInstruction stringInstruction = (AnnotatedBytecodeInstruction)(input[index + i*2]);
--- a/DMCompiler/Optimizer/PeepholeOptimizations.cs
+++ b/DMCompiler/Optimizer/PeepholeOptimizations.cs
@@ -236,6 +236,8 @@ internal sealed class PushNRef : IPeepholeOptimization {
 // PushString [string]
 // PushFloat [float]
 // -> PushStringFloat [string] [float]
+// or if there's multiple
+// -> PushNOfStringFloat [count] [string] [float] ... [string] [float]
 internal sealed class PushStringFloat : IPeepholeOptimization {
     public ReadOnlySpan<DreamProcOpcode> GetOpcodes() {
         return [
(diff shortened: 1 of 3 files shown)


DMCompiler/Optimizer/PeepholeOptimizations.cs
```diff
@@ -249,13 +251,40 @@
...
+
+            input.RemoveRange(index, 2);
+            input.Insert(index, new AnnotatedBytecodeInstruction(DreamProcOpcode.PushStringFloat, [pushVal1, pushVal2]));
+            return;
+        }
+
+        // Otherwise, replace with PushNOfStringFloat
+
+        int stackDelta = 0;
+        List<IAnnotatedBytecode> args = new List<IAnnotatedBytecode>(2 * count + 1);
```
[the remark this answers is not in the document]
ike709:
```suggestion
        List<IAnnotatedBytecode> args = new List<IAnnotatedBytecode>(2 * count + 1) { new AnnotatedBytecodeInteger(count, input[index].GetLocation()) };
```

DMCompiler/Optimizer/PeepholeOptimizations.cs
```diff
@@ -249,13 +251,40 @@ public void Apply(List<IAnnotatedBytecode> input, int index) {
...
+            input.RemoveRange(index, 2);
+            input.Insert(index, new AnnotatedBytecodeInstruction(DreamProcOpcode.PushStringFloat, [pushVal1, pushVal2]));
+            return;
+        }
+
+        // Otherwise, replace with PushNOfStringFloat
+
+        int stackDelta = 0;
+        List<IAnnotatedBytecode> args = new List<IAnnotatedBytecode>(2 * count + 1) { new AnnotatedBytecodeInteger(count, input[index].GetLocation()) };
+        args.Add(new AnnotatedBytecodeInteger(count, input[index].GetLocation()));
```
ike709, line 276:
```suggestion
```

DMCompiler/Optimizer/PeepholeOptimizations.cs
```diff
@@ -267,13 +269,39 @@ public void Apply(List<IAnnotatedBytecode> input, int index) {
...
+
+        int stackDelta = 0;
+        List<IAnnotatedBytecode> args = new List<IAnnotatedBytecode>(2 * count + 1) { new AnnotatedBytecodeInteger(count, input[index].GetLocation()) };
+
+        for (int i = 0; i < count; i++) {
+            AnnotatedBytecodeInstruction stringInstruction = (AnnotatedBytecodeInstruction)(input[index + i*2]);
+            AnnotatedBytecodeInstruction floatInstruction = (AnnotatedBytecodeInstruction)(input[index + i*2 + 1]);
+            args.Add(stringInstruction.GetArg<AnnotatedBytecodeString>(0));
+            args.Add(floatInstruction.GetArg<AnnotatedBytecodeFloat>(0));
+            stackDelta += 2;
```
wixoaGit, line 300: None of these args are pushed to the stack and so shouldn't have an effect on the stack.

wixoaGit (changes requested):

DMCompiler/Optimizer/PeepholeOptimizations.cs
wixoaGit: Turns out an opcode starting with `Push...` does indeed push

wixoaGit (approved):

PR #2057, opened 2024-10-22
Add an `AppendNoPush` optimization

ike709:
There are ~10.5k usages of `Append` in Paradise bytecode. Of these, all but ~160 immediately pop the value that `Append` pushes to the stack.

This adds a simple `AppendNoPush` peephole opt and opcode, with identical behavior except the opcode handler does not push the result to the stack.

### DIFF
--- a/OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
+++ b/OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
@@ -742,41 +742,8 @@ namespace OpenDreamRuntime.Procs {
             throw new Exception("Invalid add operation on " + first + " and " + second);
         }
 
-        /// <remarks>Any changes here should be replicated in <see cref="AppendNoPush"/></remarks>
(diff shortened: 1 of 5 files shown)


OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -742,6 +742,7 @@ private static void ThrowInvalidAddOperation(DreamValue first, DreamValue second
             throw new Exception("Invalid add operation on " + first + " and " + second);
         }
 
+        /// <remarks>Any changes here should be replicated in <see cref="AppendNoPush"/></remarks>
```
wixoaGit, line 745: You should instead move the shared code to a new method.

wixoaGit (changes requested):

wixoaGit (approved):

wixoaGit (approved):

PR #2063, opened 2024-10-26
Fix upward path search operator without lefthand side

### LINKED ISSUE #617: Path operators don't work without a left-hand side
image

ike709:
The upward path search operator didn't resolve paths correctly specifically when there was nothing on the lefthand side of the operator. 

This is because:
```
/obj/foo
/obj/bar
    var/meep = .foo
```
would resolve to `/obj/.foo` instead of `/obj/foo`

Now we strip the search operator out of the path string if there's nothing on the lefthand side, so it can be combined into the correct path. 

The other upward search unit tests confirm that this doesn't break anything when there _is_ a lefthand side. I also confirmed TG didn't explode.

Fixes #617

### DIFF
--- a/Content.Tests/DMProject/Tests/Tree/upward_search_path3.dm
+++ b/Content.Tests/DMProject/Tests/Tree/upward_search_path3.dm
@@ -1,0 +1,8 @@
+
+//# issue 617
+
+/datum/foo
+/datum/bar/var/meep = .foo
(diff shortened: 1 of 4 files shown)


ike709:
Fixed on master and unit test added in https://github.com/OpenDreamProject/OpenDream/pull/2087

PR #2064, opened 2024-10-27
Lint for `/var/` in proc args

ike709:
Lints for this BYOND bug: https://www.byond.com/forum/post/2830750

The following code prints `5` in BYOND but errors with `Unknown identifier "global.bar"` in OpenDream.
```
/datum/proc/foo(/var/bar = 5)
    return

/proc/main()
    world.log << global.bar // This prints 5 in BYOND
```

This PR doesn't replicate BYOND behavior, it just adds a lint. Ref https://github.com/OpenDreamProject/OpenDream/issues/2065 for tracking the status of replicating BYOND behavior.

Credit to @out-of-phaze for reminding me of this BYOND bug's existence.

I tested about half a dozen codebases and only found a single hit in goonstation.

### DIFF
--- a/DMCompiler/Compiler/CompilerError.cs
+++ b/DMCompiler/Compiler/CompilerError.cs
@@ -49,7 +49,7 @@ public enum WarningCode {
     FallbackBuiltinArgument = 2208, // A builtin (sin(), cos(), etc) with an invalid/fallback argument
     PointlessScopeOperator = 2209,
     PointlessPositionalArgument = 2210,
-    ProcArgumentGlobal = 2211, // Prepending "/" on a proc arg (e.g. "/proc(/var/foo)" makes the arg a global var. Ref https://www.byond.com/forum/post/2830750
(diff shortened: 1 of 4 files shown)


DMCompiler/Compiler/CompilerError.cs
```diff
@@ -49,6 +49,7 @@ public enum WarningCode {
     FallbackBuiltinArgument = 2208, // A builtin (sin(), cos(), etc) with an invalid/fallback argument
     PointlessScopeOperator = 2209,
     PointlessPositionalArgument = 2210,
+    ProcArgumentGlobal = 2211, // Prepending "/" on a proc arg (e.g. "/proc(/var/foo)" makes the arg a global var. Ref https://www.byond.com/forum/post/2830750
```
ike709, line 52:
```suggestion
    ProcArgumentGlobal = 2211, // Prepending "/" on a proc arg (e.g. "/proc/example(/var/foo)" makes the arg a global var. Ref https://www.byond.com/forum/post/2830750
```

ZeWaka:
fixed on goon

wixoaGit (approved):

PR #2070, opened 2024-10-28
Try to disable the GC in DMCompiler

ike709:
This PR attempts to disable the garbage collector during compilation if sufficient memory is available. It's no guarantee and it's currently set to disable the GC if ~4.5 GB is available. This is to give us some breathing room on top of the ~3.9 GB of total allocs for compiling TG currently. 

Note that this memory value does not cause it to allocate the specified amount of memory nor cause any sort of issues if that amount of memory isn't available. It'll just run the GC in that case.

This takes TG from a consistent 22s down to 18s on my machine, and Paradise from 13s to 11s.

Added a `--use-gc` compiler arg if people want the old behavior for some reason.


wixoaGit:
I'd like to try reducing the largest allocators before doing this.

PR #2072, opened 2024-10-31
Typemaker 1.1 - fixes and improvements

out-of-phaze:
A bunch of fixes/improvements to typemaker inference, plus:
- `DMValueType.Instance` is now the default when doing `as /some/typepath`.
- `as path(/some/typepath)` can be used to specifically denote 'a typepath of type `/some/typepath`'.
- Lists can now have types provided: `as /list(some union astypes here)` for normal lists, `as /list(some union astypes here, some others here)` for associative lists. Empty lists currently always match all typed lists because type info is not tracked at runtime. Similarly, a list with only keys and no associated values can match a typed associative list (this was necessary to fix some stuff on TG, and besides, it's hard to differentiate between `list("foo")` and `list("foo" = null)`.
- An experimental `as params[1]` astypes syntax has been added. You can also even do `as params[1] as instance` to denote that, if `params[1]` is a path, this value is an instance with the same type. (This is useful for singleton getters.)

I beg you, please help me find some other way to do the params stuff other than just threading `procParameters` through the entire parser. It's SO BAD. I hate it

### DIFF
--- a/DMCompiler/Compiler/DM/AST/DMAST.ObjectStatements.cs
+++ b/DMCompiler/Compiler/DM/AST/DMAST.ObjectStatements.cs
@@ -52,7 +52,7 @@ public sealed class DMASTObjectVarDefinition(
     Location location,
     DreamPath path,
     DMASTExpression value,
-    DMComplexValueType valType,
+    DMComplexValueType? valType,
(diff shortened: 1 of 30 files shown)


wixoaGit:
Closing in favor of #2208

PR #2073, opened 2024-11-01
Use a `switch` in `ProcStatement()`

ike709:
Doesn't really impact compiletime either way, but it means we no longer have to worry about the order we're parsing proc statements.


wixoaGit:
> Doesn't really impact compiletime either way, but it means we no longer have to worry about the order we're parsing proc statements.

Did we ever have to worry about it if it doesn't impact compile time?

ike709:
> Did we ever have to worry about it if it doesn't impact compile time?

The order of proc statements in the original code did have a measurable impact on compile time (at least back when I first rearranged them based on frequency, dunno about current year).

DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -747,22 +747,55 @@ public DMASTFile File() {
...
+                        procStatement = TryCatch();
+                        break;
+                    case TokenType.DM_Goto:
+                        procStatement = Goto();
+                        break;
+                    case TokenType.DM_Slash:
+                    case TokenType.DM_Var:
+                        procStatement = ProcVarDeclaration();
+                        break;
+                }
```
wixoaGit, line 798: I think a switch expression is a better fit here, readability wise
```suggestion
                DMASTProcStatement? procStatement = Current().Type switch {
                    TokenType.DM_If => If(),
                    TokenType.DM_Return => Return(),
                    TokenType.DM_For => For(),
                    TokenType.DM_Set => Set(),
                    TokenType.DM_Switch => Switch(),
                    TokenType.DM_Continue => Continue(),
                    TokenType.DM_Break => Break(),
                    TokenType.DM_Spawn => Spawn(),
                    TokenType.DM_While => While(),
                    TokenType.DM_Do => DoWhile(),
                    TokenType.DM_Throw => Throw(),
                    TokenType.DM_Del => Del(),
                    TokenType.DM_Try => TryCatch(),
                    TokenType.DM_Goto => Goto(),
                    TokenType.DM_Slash or TokenType.DM_Var => ProcVarDeclaration(),
                    _ => null
                };
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #2090, opened 2024-11-14
Fix some sources of runtime errors

SpaceManiac:
- Implement `list() << output(...)`
  - Like BYOND, the `list()` form only supports mobs and clients, not world, and is not recursive
- Fix `vars[1]` erroring instead of returning the name of the first var
  - No attempt was made to match BYOND's variable order, just to do something sensible instead of erroring
  - Code example: `for(var/name in (vars & data))`
- Fix `typesof("/foo/verb")` returning non-verb subtypes when there are no verbs
- Fix `Cut(2, 1)` throwing (negative count to RemoveRange) instead of doing nothing
- Fix bad offset math when DMIParser skips a `tEXt` chunk, leading to UTF-8 decode exceptions
- Fix `animate(filter, ...)` not setting LastAnimatedObject
  - Animating filters still isn't implemented, but at least a chained `animate()` call doesn't error

### DIFF
--- a/OpenDreamRuntime/Objects/Types/DreamList.cs
+++ b/OpenDreamRuntime/Objects/Types/DreamList.cs
@@ -174,7 +174,8 @@ public class DreamList : DreamObject {
                 _associativeValues.Remove(_values[i - 1]);
         }
 
-        _values.RemoveRange(start - 1, end - start);
+        if (end > start)
+            _values.RemoveRange(start - 1, end - start);
     }
 
     public void Insert(int index, DreamValue value) {
--- a/OpenDreamRuntime/Objects/Types/DreamList.cs
+++ b/OpenDreamRuntime/Objects/Types/DreamList.cs
@@ -416,6 +416,7 @@ internal sealed class DreamListVars(DreamObjectDefinition listDef, DreamObject d
             if (DreamObject.TryGetVariable(varName, out var objectVar)) {
                 return objectVar;
             }
+            
             throw new Exception($"Cannot get value of undefined var \"{key}\" on type {DreamObject.ObjectDefinition.Type}");
         } else {
             throw new Exception($"Invalid var index {key}");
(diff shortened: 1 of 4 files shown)


OpenDreamRuntime/Objects/Types/DreamList.cs
```diff
@@ -409,16 +410,16 @@ public override bool ContainsValue(DreamValue value) {
...
 
     public override DreamValue GetValue(DreamValue key) {
-        if (!key.TryGetValueAsString(out var varName)) {
+        if (key.TryGetValueAsInteger(out int keyInteger)) {
+            return new DreamValue(DreamObject.GetVariableNames().ElementAt(keyInteger - 1)); //1-indexed
+        } else if (key.TryGetValueAsString(out var varName)) {
+            if (DreamObject.TryGetVariable(varName, out var objectVar)) {
+                return objectVar;
+            }
+            throw new Exception($"Cannot get value of undefined var \"{key}\" on type {DreamObject.ObjectDefinition.Type}");
```
wixoaGit, line 419:
```suggestion
            
            throw new Exception($"Cannot get value of undefined var \"{key}\" on type {DreamObject.ObjectDefinition.Type}");
```

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -156,10 +156,10 @@ public static DreamValue NativeProc_animate(NativeProc.Bundle bundle, DreamObjec
                 return DreamValue.Null;
             chainAnim = true;
         }
+        bundle.LastAnimatedObject = new DreamValue(obj);
```
wixoaGit, line 159:
```suggestion
        
        bundle.LastAnimatedObject = new DreamValue(obj);
```

wixoaGit (changes requested):

PR #2091, opened 2024-11-14
Fix static vars trying to reference local vars, and add pragma

### LINKED ISSUE #997: In BYOND, RHS of Static Vars Can Only Access `world` Vars
Consider this code:
```
/world
    name = "wew lad"

var/global/name = "wumbo"

/obj
    name = "meep"
/obj/proc/bar()
    var/static/L = name
    world.log << L
/proc/main()
    (new /obj()).bar()
```
What do you think it prints? The answer is `"wew lad"` because `var/static/L = name` is actually `var/static/L = world.name` in BYOND.

SaphireLattice:
Pragma `OD2212` triggers as an error when the code accesses a global var while there's an identically named local one

Closes #997

### DIFF
--- a/DMCompiler/Compiler/CompilerError.cs
+++ b/DMCompiler/Compiler/CompilerError.cs
@@ -50,7 +50,7 @@ public enum WarningCode {
     PointlessScopeOperator = 2209,
     PointlessPositionalArgument = 2210,
     ProcArgumentGlobal = 2211, // Prepending "/" on a proc arg (e.g. "/proc/example(/var/foo)" makes the arg a global var. Ref https://www.byond.com/forum/post/2830750
-    AmbiguousVarStatic = 2212, // Assigning an instance variable to a static one (which gives unknown ref) while a global exists (which makes it work, with unexpected value)
+    AmbiguousVarStatic = 2212, // Referencing a static variable when an instance variable with the same name exists
     MalformedRange = 2300,
     InvalidRange = 2301,
     InvalidSetStatement = 2302,
--- a/DMCompiler/Compiler/CompilerError.cs
+++ b/DMCompiler/Compiler/CompilerError.cs
@@ -50,6 +50,7 @@ public enum WarningCode {
     PointlessScopeOperator = 2209,
     PointlessPositionalArgument = 2210,
     ProcArgumentGlobal = 2211, // Prepending "/" on a proc arg (e.g. "/proc/example(/var/foo)" makes the arg a global var. Ref https://www.byond.com/forum/post/2830750
+    AmbiguousVarStatic = 2212, // Assigning an instance variable to a static one (which gives unknown ref) while a global exists (which makes it work, with unexpected value)
     MalformedRange = 2300,
     InvalidRange = 2301,
     InvalidSetStatement = 2302,
--- a/DMCompiler/DM/Builders/DMExpressionBuilder.cs
+++ b/DMCompiler/DM/Builders/DMExpressionBuilder.cs
@@ -554,13 +554,16 @@ internal static class DMExpressionBuilder {
                 }
 
                 var field = dmObject.GetVariable(name);
-                if (field != null) {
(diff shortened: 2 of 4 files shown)


DMCompiler/DM/Builders/DMExpressionBuilder.cs
```diff
@@ -554,13 +554,16 @@ private static DMExpression BuildIdentifier(DMASTIdentifier identifier, DMObject
...
-                if (field != null) {
+                if (field != null && CurrentScopeMode == ScopeMode.Normal) {
                     return new Field(identifier.Location, field, field.ValType);
                 }
 
                 var globalId = proc?.GetGlobalVariableId(name) ?? dmObject.GetGlobalVariableId(name);
 
                 if (globalId != null) {
+                    if (field is not null)
+                        DMCompiler.Emit(WarningCode.AmbiguousVarStatic, identifier.Location, $"Static var definition can not reference instance variable \"{name}\" but a global exists");
```
wixoaGit, line 565:
```suggestion
                        DMCompiler.Emit(WarningCode.AmbiguousVarStatic, identifier.Location, $"Static var definition cannot reference instance variable \"{name}\" but a global exists");
```

DMCompiler/DMStandard/DefaultPragmaConfig.dm
```diff
@@ -32,6 +32,7 @@
 #pragma SuspiciousSwitchCase warning
 #pragma PointlessPositionalArgument warning
 #pragma ProcArgumentGlobal warning // Ref BYOND issue https://www.byond.com/forum/post/2830750
+#pragma AmbiguousVarStatic error // https://github.com/OpenDreamProject/OpenDream/issues/997
```
wixoaGit, line 35: I think this should be a notice by default. When a static of the same name exists it's fine to use that.

When I said it should be an error, I was thinking of situations where there was only an instance variable with the name.
```suggestion
#pragma AmbiguousVarStatic notice // https://github.com/OpenDreamProject/OpenDream/issues/997
```

DMCompiler/Compiler/CompilerError.cs
```diff
@@ -50,6 +50,7 @@ public enum WarningCode {
     PointlessScopeOperator = 2209,
     PointlessPositionalArgument = 2210,
     ProcArgumentGlobal = 2211, // Prepending "/" on a proc arg (e.g. "/proc/example(/var/foo)" makes the arg a global var. Ref https://www.byond.com/forum/post/2830750
+    AmbiguousVarStatic = 2212, // Assigning an instance variable to a static one (which gives unknown ref) while a global exists (which makes it work, with unexpected value)
```
wixoaGit, line 53:
```suggestion
    AmbiguousVarStatic = 2212, // Referencing a static variable when an instance variable with the same name exists
```

wixoaGit (changes requested):

DMCompiler/DMStandard/DefaultPragmaConfig.dm
SaphireLattice: I think if a codebase needs to use ambiguous names like this, they could do this on their own. Otherwise it prevents awful footguns.

The thing didn't trigger on Paradise at all. Not sure about other codebases though.

wixoaGit: The general philosophy with these pragmas has been to not error on things that byond doesn't. If a codebase wants the extra guarantee of erroring on it, they can raise it themselves (as they've been doing up to now).

I said `notice` because I felt it was a little pedantic of a lint, but I can see a case being made for `warning`.

SaphireLattice: Ah, I see. 

Probably want a pedantic mode in compiler that uses harsher pragmas, eventually

Probably should be a warning. It's gonna be pretty ambiguous when trying to read it unless you keep this behavior in mind.

wixoaGit (approved):

Issue #2095, opened 2024-11-16
`world.opendream_topic_port` is being const-folded to `null`


ike709:
This is breaking TGS tests.

Issue #2094, opened 2024-11-16
`world.system_type` should be const but isn't


ike709:
OpenDream just declares it as `/world/var/system_type` but it's clearly const in BYOND:

![image](https://github.com/user-attachments/assets/e992604c-3e36-471e-8955-7c9b9df5549f)

PR #2096, opened 2024-11-16
Unit test some `world` vars against bad const folding, fix `world.system_type`

ike709:
Won't work until #2095 is fixed but it will prevent that from reoccurring.

Also tests against repeating the issue when #2094 is resolved.

Also fixes `world.system_type` not being a string. Woops.


Content.Tests/DMProject/Tests/Builtins/world_system_type.dm
```diff
@@ -0,0 +1,3 @@
+
+/proc/RunTest()
+	ASSERT(isnum(world.system_type))
```
wixoaGit, line 3: `world.system_type` is a string, not a var

wixoaGit (changes requested):

Content.Tests/DMProject/Tests/Builtins/world_system_type.dm
ike709, line 3:
```suggestion
	ASSERT(istext(world.system_type) && length(world.system_type))
```

wixoaGit (approved):

PR #2104, opened 2024-11-19
Implement `atom.appearance.overlays`

wixoaGit:
BYOND just gives you a normal list populated with the overlays/underlays appearances

### DIFF
--- a/OpenDreamRuntime/AtomManager.cs
+++ b/OpenDreamRuntime/AtomManager.cs
@@ -376,8 +376,8 @@ public sealed class AtomManager {
             case "underlays":
                 throw new Exception($"Cannot assign the {varName} var on an appearance");
 
-            // TODO: overlays, underlays, filters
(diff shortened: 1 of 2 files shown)


OpenDreamRuntime/AtomManager.cs
```diff
@@ -370,6 +370,12 @@ public void SetAppearanceVar(IconAppearance appearance, string varName, DreamVal
...
             case "appearance":
                 throw new Exception("Cannot assign the appearance var on an appearance");
+
+            // These should be handled by the DreamObject if being accessed through that
+            case "overlays":
+            case "underlays":
+                throw new Exception($"Cannot assign the {varName} var on an appearance");
+
             // TODO: overlays, underlays, filters
             //       Those are handled separately by whatever is calling SetAppearanceVar currently
```
wixoaGit, line 380:
```suggestion
            // TODO: filters
            //       It's handled separately by whatever is calling SetAppearanceVar currently
```

Issue #1773, opened 2024-05-08
Nebula: Map-edge screen_loc anchors do not work (NORTH, etc)


out-of-phaze:
See ref:
> The `LEFT`, `RIGHT`, `TOP`, and `BOTTOM` keywords (also `TOPLEFT`, `TOPRIGHT`, `BOTTOMLEFT`, `BOTTOMRIGHT`) can be used to anchor a screen object to the edge of the map control if the map is zoomed in so that some pixels are cut off. When you use these edge-alignment keywords, the icon size _is_ taken into account, and the alignment of the icon changes to fit whichever edge you use. Because these keywords do not conform to the normal tile-based structure of the HUD, they can't be used for a range of tiles with the `"to"` format.

Currently these are just identical to NORTH/SOUTH/EAST/WEST.

out-of-phaze:
Working on this, btw.

wixoaGit:
As of #2106 these anchors use icon size instead of tile size. The other half of fixing this issue would be to have the icons stick to the edge of the visible map when the map's container is smaller than the map itself.

PR #2106, opened 2024-11-21
Implement the `screen_loc` edge anchors

wixoaGit:
`LEFT`/`RIGHT`/`TOP`/`BOTTOM` are no longer the same as  `WEST`/`EAST`/`NORTH`/`SOUTH`
They take rendered icon size into account instead of the map's icon size.

Fixes tg's lobby menu
image

Partially addresses #1773, the icons aren't kept on the edge if the map's parent control is smaller than the map itself.

### DIFF
--- a/OpenDreamShared/Dream/ScreenLocation.cs
+++ b/OpenDreamShared/Dream/ScreenLocation.cs
@@ -72,6 +72,8 @@ public sealed class ScreenLocation {
     }
 
     public Vector2 GetViewPosition(Vector2 viewOffset, ViewRange view, float tileSize, Vector2i iconSize) {
+        // TODO: LEFT/RIGHT/TOP/BOTTOM need to stick to the edge of the visible map if the map's container is smaller than the map itself
+        
         float x = (X + PixelOffsetX / tileSize);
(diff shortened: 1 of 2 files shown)


OpenDreamShared/Dream/ScreenLocation.cs
```diff
@@ -66,20 +71,22 @@ public ScreenLocation(string screenLocation) {
         ParseScreenLoc(screenLocation);
     }
 
-    public Vector2 GetViewPosition(Vector2 viewOffset, ViewRange view, float iconSize) {
-        float x = (X + PixelOffsetX / iconSize);
+    public Vector2 GetViewPosition(Vector2 viewOffset, ViewRange view, float tileSize, Vector2i iconSize) {
```
wixoaGit, line 74:
```suggestion
    public Vector2 GetViewPosition(Vector2 viewOffset, ViewRange view, float tileSize, Vector2i iconSize) {
        // TODO: LEFT/RIGHT/TOP/BOTTOM need to stick to the edge of the visible map if the map's container is smaller than the map itself
        
```

PR #2113, opened 2024-11-29
Fix a bad proc param test

ike709:
Another Broken Test test that didn't actually work in BYOND. Also changed the obj to a datum for no particular reason.

Original:
image

This PR:
image


amylizzle (approved):
huh I swear byond used to do some fuckery if you passed `null` as an arg name

ike709:
Not in this context. That involves shenanigans with proc overrides, like I told you on discord the other day.

wixoaGit:
> Not in this context. That involves shenanigans with proc overrides, like I told you on discord the other day.

I'm fairly sure using `null` as the name of an argument in any context would have that become the new `null`. This was probably fixed in newer versions of BYOND.

wixoaGit (approved):

PR #2116, opened 2024-11-30
Cope with extra tokens after proc statements

ike709:
Trailing `.` and `:` tokens after `if()` and `for()` statements is valid in BYOND:
image

We already handled `if():` so I just expanded our copium and fixed & extended a broken unit test in the process. Added the `ExtraToken` stylistic pragma for when this happens. No hits on TG/goon/para.

### DIFF
--- a/DMCompiler/Compiler/DM/DMParser.cs
+++ b/DMCompiler/Compiler/DM/DMParser.cs
@@ -1125,8 +1125,10 @@ namespace DMCompiler.Compiler.DM {
 
             BracketWhitespace();
             ConsumeRightParenthesis();
-            Whitespace();
-            Check(TokenType.DM_Colon);
+            if (Check(TokenType.DM_Colon) || Check(TokenType.DM_Period)) {
+                Emit(WarningCode.ExtraToken, loc, "Extra token at end of proc statement");
+            }
+
             Whitespace();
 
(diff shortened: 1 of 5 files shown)


DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -1125,8 +1125,10 @@ private DMASTProcStatementIf If() {
 
             BracketWhitespace();
             ConsumeRightParenthesis();
-            Whitespace();
-            Check(TokenType.DM_Colon);
+            if (Check(TokenType.DM_Colon) || Check(TokenType.DM_Period)) {
+                Emit(WarningCode.ExtraToken, loc, "Extra token at end of proc statement");
+            }
```
ike709, line 1130: Checking for whitespace first would cause `if(foo)       . += "bar"` to be caught, which is paraphrased from actual SS13 code.

wixoaGit: This does work though:
```
if(0)     .
    world.log << 1
```

amylizzle: `WhitespaceNoNewline()` time?

wixoaGit (approved):

PR #2122, opened 2024-12-06
Fix __IMPLIED_TYPE__ in proc arguments

out-of-phaze:
it just wasn't ever passed

### DIFF
--- a/Content.Tests/DMProject/Tests/Builtins/__IMPLIED_TYPE__.dm
+++ b/Content.Tests/DMProject/Tests/Builtins/__IMPLIED_TYPE__.dm
@@ -1,7 +1,10 @@
 
 
-/datum/test/var/bar = "foobar"
+/datum/test
 /proc/RunTest()
 	var/datum/test/D = __IMPLIED_TYPE__
-	ASSERT(D.bar == "foobar")
+	ASSERT(D == /datum/test)
+	D = ArgumentTest(__IMPLIED_TYPE__)
 
+/proc/ArgumentTest(some_argument)
+	ASSERT(some_argument == /datum/test)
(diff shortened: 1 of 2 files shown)


Content.Tests/DMProject/Tests/Builtins/__IMPLIED_TYPE__.dm
```diff
@@ -4,4 +4,8 @@
 /proc/RunTest()
 	var/datum/test/D = __IMPLIED_TYPE__
 	ASSERT(D.bar == "foobar")
+	D = ArgumentTest(__IMPLIED_TYPE__)
+	ASSERT(D.bar == "foobar")
```
wixoaGit, line 8: This second assert is redundant

Content.Tests/DMProject/Tests/Builtins/__IMPLIED_TYPE__.dm
```diff
@@ -4,4 +4,8 @@
...
 
+/proc/ArgumentTest(some_argument)
+	return some_argument
```
wixoaGit, line 11: This should instead assert that the argument is the correct value

wixoaGit (changes requested):

Content.Tests/DMProject/Tests/Builtins/__IMPLIED_TYPE__.dm
out-of-phaze: it's not redundant, we reassign D. we *must* reassign D or else there is no implied type for the arguments, and since we reassign D, redoing the assert is the easiest way i could think of to check it
that said i think it would probably be better both times to just check `ASSERT(D == /datum/test)`

wixoaGit (approved):

PR #2125, opened 2024-12-07
Don't instantiate multidimensional lists in proc params

ike709:
In BYOND `/proc/foo(bar[5])` treats `bar` as null but OD was instantiating a multidimensional list (of 1 dimension).

Fixed 3 unit tests.


ike709:
CI failing due to https://github.com/NebulaSS13/Nebula/pull/4619

wixoaGit (approved):

PR #2127, opened 2024-12-07
Tracy

amylizzle:
with thanks to https://github.com/wixoaGit/OpenDreamFork/tree/tracy for doing most of the work

When in TOOLS build activates proc logging for DM with source locations.

image
image
also it does native procs so we can use it to make OD faster
image
Memory tracing:
image
image

Note: DreamObject memory footprint is not accurate. It assumes that each object allocates its total possible memory.


OpenDreamRuntime/Objects/DreamObjectTree.cs
```diff
@@ -144,44 +144,46 @@ public IEnumerable<TreeEntry> GetAllDescendants(TreeEntry treeEntry) {
...
+            if (type.ObjectDefinition.IsSubtypeOf(Client))
+                throw new Exception("Cannot create objects of type /client");
+            if (type.ObjectDefinition.IsSubtypeOf(Turf))
+                throw new Exception("New turfs must be created by the map manager");
+            if (type.ObjectDefinition.IsSubtypeOf(Exception))
+                return new DreamObjectException(type.ObjectDefinition);
+            if (type.ObjectDefinition.IsSubtypeOf(Vector))
+                return new DreamObjectVector(type.ObjectDefinition);
+
+                return new DreamObject(type.ObjectDefinition);
```
amylizzle, line 185:
```suggestion
            return new DreamObject(type.ObjectDefinition);
```

OpenDreamRuntime/Profile.cs
```diff
@@ -0,0 +1,295 @@
...
+    public static void Activate(){
+        _isActivated = true;
+    }
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static bool IsActivated(){
+        return _isActivated;
+    }
+
+
```
amylizzle, line 98:
```suggestion
```

wixoaGit (approved):

PR #2132, opened 2024-12-10
`IndexRefWithString` peephole opt

ike709:
`PushReferenceValue -> PushString -> DereferenceIndex` is a pretty common pattern in TG. This combines them into one new `IndexRefWithString` opcode.

The new opcode occurs 6,170 times in TG. This means a pretty decent reduction in pushing/popping the stack.

I joined TG and confirmed it didn't seem to explode, and I could still run around.

### DIFF
--- a/DMCompiler/Optimizer/PeepholeOptimizations.cs
+++ b/DMCompiler/Optimizer/PeepholeOptimizations.cs
@@ -117,6 +117,34 @@ internal sealed class PushField : IOptimization {
 }
 
 // PushReferenceValue [ref]
+// PushString [string]
+// DereferenceIndex [value] [string]
+// -> IndexRefWithString [ref, string]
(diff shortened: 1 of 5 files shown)


DMCompiler/Optimizer/PeepholeOptimizations.cs
```diff
@@ -116,6 +116,34 @@ public void Apply(DMCompiler compiler, List<IAnnotatedBytecode> input, int index
     }
 }
 
+// PushReferenceValue [ref]
+// PushString [string]
+// DereferenceIndex [value] [string]
```
wixoaGit, line 121: DereferenceIndex doesn't have any arguments
```suggestion
// DereferenceIndex
```

wixoaGit (approved):

PR #2133, opened 2024-12-10
Remove `Jump` after `Return`

ike709:
`if()`/`else` blocks with a `return` in the `if()` are left with a dead `Jump` opcode that can never be reached. 

This peephole optimization deals with that and removes over 900 `Jump` opcodes from TG.

I tested that TG doesn't explode with these changes.


ike709:
WYCI

wixoaGit (approved):

PR #2135, opened 2024-12-10
Convert some ProcJson classes to structs

ike709:
I see no reason to allocate a morbillion classes on the heap for what could easily be structs. I don't have dotmemory anymore to measure the exact memory alloc reduction, but not allocating all of these classes will definitely be less memory.

Also converts `set;` to `init;` everywhere the IDE suggested, and fixed a warning for a missing `Equals()`.

Tested that TG doesn't explode with these changes.

### DIFF
--- a/DMCompiler/Json/DreamProcJson.cs
+++ b/DMCompiler/Json/DreamProcJson.cs
@@ -3,44 +3,48 @@
 namespace DMCompiler.Json;
 
 public sealed class ProcDefinitionJson {
-    public int OwningTypeId { get; set; }
-    public required string Name { get; set; }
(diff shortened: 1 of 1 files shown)


wixoaGit:
> I don't have dotmemory anymore to measure the exact memory alloc reduction, but not allocating all of these classes will definitely be less memory.

I'm not so sure about that, since making these structs just shafts the allocation into the arrays holding them. This could theoretically be better for cache locality, but I don't think we access these often enough for it to be an issue.

That being said I don't see a downside to this and I think it makes better sense anyways.

wixoaGit (approved):

PR #2144, opened 2024-12-17
Fix off-by-one in regex replace and add test

amylizzle:
This was subtly breaking DMM suite when there was a cell with more than one var-edited string var, which caused **one** tile to fail to load properly on goon and made the nuclear reactor control computer not work on devtest.

What a ride that was.


Content.Tests/DMProject/Tests/Regex/regex_find_replace.dm
```diff
@@ -0,0 +1,18 @@
...
+	do
+		found = noStrings.Find(models, noStrings.next)
+		if(found)
+			var indexText = {""[stringIndex]""}
+			stringIndex++
+			var match = copytext(noStrings.match, 2, -1) // Strip quotes
+			models = noStrings.Replace(models, indexText, found)
+			originalStrings[indexText] = (match)
+	while(found)
+	ASSERT(models == result_match)
```
wixoaGit, line 18: Couldn't hurt to ensure these variables are also as they should be
```suggestion
	ASSERT(models == result_match)
	ASSERT(originalStrings ~= list("\"1\"" = "2-8", "\"2\"" = "4-8"))
	ASSERT(stringIndex == 3)
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #2145, opened 2024-12-18
Fix atom name/desc not copying on appearance set

### LINKED ISSUE #2107: `thing.appearance = otherthing.appearance` doesn't copy name and desc
```
/obj/thingtocopy
    name = "hello"
    desc = "this is a thing"

/proc/test()
    var/obj/thingtocopy/T = new()
    var/obj/otherthing = new()
    otherthing.appearance = T.appearance
    ASSERT(otherthing.name == T.name)
    ASSERT(otherthing.desc == T.desc)
```

There's probably some other things that are copied in appearance assignments like this that we've missed too.

TobleroneSwordfish:
Fixes #2107 by adding a desc var to appearances and changing atom var access methods to always defer to appearance.
There's a warning on the `IconAppearance` class that begins "Woe, weary traveler" so hopefully I've done all the steps right to avoid doom or whatever.
Includes a DM test to hopefully prove it works.

### DIFF
--- a/Content.Tests/DMProject/Tests/Image/appearance.dm
+++ b/Content.Tests/DMProject/Tests/Image/appearance.dm
@@ -1,0 +1,10 @@
+/obj/thingtocopy
+    name = "hello"
+    desc = "this is a thing"
+
+/proc/RunTest()
+    var/obj/thingtocopy/T = new()
+    var/obj/otherthing = new()
+    otherthing.appearance = T.appearance
+    ASSERT(otherthing.name == T.name)
+    ASSERT(otherthing.desc == T.desc)
--- a/OpenDreamRuntime/AtomManager.cs
+++ b/OpenDreamRuntime/AtomManager.cs
@@ -209,6 +209,7 @@ public sealed class AtomManager {
     public bool IsValidAppearanceVar(string name) {
         switch (name) {
             case "name":
+            case "desc":
             case "icon":
             case "icon_state":
             case "dir":
@@ -247,6 +248,10 @@ public sealed class AtomManager {
             case "name":
                 value.TryGetValueAsString(out var name);
                 appearance.Name = name ?? string.Empty;
+                break;
+            case "desc":
+                value.TryGetValueAsString(out var desc);
+                appearance.Desc = desc ?? string.Empty;
                 break;
             case "icon":
                 if (_resourceManager.TryLoadIcon(value, out var icon)) {
@@ -387,6 +392,8 @@ public sealed class AtomManager {
         switch (varName) {
             case "name":
                 return new(appearance.Name);
+            case "desc":
+                return new(appearance.Desc);
             case "icon":
                 if (appearance.Icon == null)
                     return DreamValue.Null;
@@ -589,6 +596,7 @@ public sealed class AtomManager {
             return appearance;
 
         def.TryGetVariable("name", out var nameVar);
+        def.TryGetVariable("desc", out var descVar);
         def.TryGetVariable("icon", out var iconVar);
         def.TryGetVariable("icon_state", out var stateVar);
         def.TryGetVariable("color", out var colorVar);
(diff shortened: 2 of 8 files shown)


OpenDreamRuntime/Objects/DreamObject.cs
```diff
@@ -13,6 +13,7 @@
 using Robust.Shared.Map;
 using Robust.Shared.Serialization.Manager;
 using Robust.Shared.Utility;
+using System.ComponentModel;
```
amylizzle, line 16:
```suggestion
```
vscode likes to add usings without telling you *a lot*

OpenDreamRuntime/Objects/Types/DreamObjectAtom.cs
```diff
@@ -5,7 +5,6 @@ namespace OpenDreamRuntime.Objects.Types;
 
 [Virtual]
 public class DreamObjectAtom : DreamObject {
-    public string? Name;
     public string? Desc;
```
amylizzle, line 8:
```suggestion
```

could probably clip this also?

amylizzle (changes requested):

OpenDreamRuntime/Objects/DreamObject.cs
TobleroneSwordfish: guh

OpenDreamRuntime/Objects/Types/DreamObjectAtom.cs
TobleroneSwordfish: Ohh, yep I put that one back in to find all the usages of it when it errored and forgot to remove it again

Content.Tests/DMProject/Tests/Image/appearance.dm
```diff
@@ -0,0 +1,10 @@
+/obj/thingtocopy
```
wixoaGit, line 1: Using atom-derivatives in a unit test is generally not a good idea. Our unit testing framework doesn't start up a whole server, and so the entity systems/entities responsible for atoms aren't created. This is the reason tests are failing on this PR.

If you really need to test atom functionality, you need to add to the `Content.IntegrationTests` project.

OpenDreamRuntime/AtomManager.cs
```diff
@@ -248,6 +249,10 @@ public void SetAppearanceVar(IconAppearance appearance, string varName, DreamVal
                 value.TryGetValueAsString(out var name);
                 appearance.Name = name ?? string.Empty;
                 break;
+            case "desc":
+                value.TryGetValueAsString(out var desc);
+                appearance.Desc = desc ?? string.Empty;
```
wixoaGit, line 254: Setting `desc` to an invalid value like `123` sets it to `null` in BYOND. `Desc` needs to be a nullable string, `string?`.

OpenDreamRuntime/Objects/Types/DreamObjectAtom.cs
```diff
@@ -38,6 +31,12 @@ protected override void HandleDeletion(bool possiblyThreaded) {
 
         base.HandleDeletion(possiblyThreaded);
     }
+    public string GetDesc() {
```
wixoaGit, line 34: Some spacing between methods, but this should also be renamed to `GetRTEntityDesc()` to make it clearer that it's only used for RT's entities.

```suggestion
    }
    
    public string GetDesc() {
```

OpenDreamRuntime/Objects/Types/DreamObjectAtom.cs
```diff
@@ -38,6 +31,12 @@ protected override void HandleDeletion(bool possiblyThreaded) {
...
+        if (!TryGetVariable("desc", out DreamValue descVar) || !descVar.TryGetValueAsString(out string? desc))
+            return ObjectDefinition.Type.ToString();
```
wixoaGit, line 36: `ObjectDefinition.Type` is already a string.
```suggestion
            return ObjectDefinition.Type;
```

wixoaGit (changes requested):

Content.Tests/DMProject/Tests/Image/appearance.dm
TobleroneSwordfish: Okay the test probably isn't all that necessary, but I'll move it over to integration tests just to learn how to do that 👍

TobleroneSwordfish:
guh

wixoaGit (approved):

PR #2150, opened 2024-12-29
Division unit test fixes

ike709:
- Added error codes to `// COMPILE ERROR` division tests that didn't have them
- Fixed `// RUNTIME ERROR` unit tests not passing if the runtime error occurred prior to `RunTest()` execution. `_dreamMan.LastDMException` is now reset prior to running the current test, and the test now checks if it's null.
- Moved the now-passing divide-by-zero tests to the `Tests` directory


DMCompiler/Optimizer/PeepholeOptimizations.cs
```diff
@@ -458,6 +459,10 @@ public void Apply(DMCompiler compiler, List<IAnnotatedBytecode> input, int index
 
         IOptimization.GetInstructionAndValue(input[index + 1], out var pushVal2);
 
+        if (pushVal2 == 0) {
+            compiler.Emit(WarningCode.BadExpression, input[index + 1].GetLocation(), "Division by zero");
+        }
```
ike709, line 464: This fixes `div_zero4.dm`

It would fix the other broken tests if we actually did const folding on them.

wixoaGit: We want peephole optimizations to be disable-able with CLI arguments someday right? Tying compile errors to them probably isn't a good idea.

wixoaGit (changes requested):

DMCompiler/Optimizer/PeepholeOptimizations.cs
```diff
@@ -1,4 +1,5 @@
 using DMCompiler.Bytecode;
+using DMCompiler.Compiler;
```
ike709, line 2:
```suggestion
```

wixoaGit (approved):

PR #2153, opened 2025-01-04
List append self

### LINKED ISSUE #2149: Appending a list to itself causes an error
Running the below code throws a modified collection error. It should instead duplicate each element, adding them to the end of the list.
```
var/list/L = list(1, 2)
L += L
```
Should end up with the list `1, 2, 1, 2`

ritorizo:
Fixes #2149, the runtime that  was caused by appending a list to itself by appending a copy of the list instead.

### DIFF
--- a/Content.Tests/DMProject/Tests/List/ListAppendSelf.dm
+++ b/Content.Tests/DMProject/Tests/List/ListAppendSelf.dm
@@ -5,4 +5,4 @@
 
 	L2 = list(a=1, b=2)
 	L2 += L2
-	ASSERT(L2 = list(a=1, b=2))
+	ASSERT(L2 ~= list(a=1, b=2, "a", "b"))
--- a/Content.Tests/DMProject/Tests/List/ListAppendSelf.dm
+++ b/Content.Tests/DMProject/Tests/List/ListAppendSelf.dm
@@ -1,8 +1,8 @@
 /proc/RunTest()
-	L1 = list(1, 2)
+	var/L1 = list(1, 2)
 	L1 += L1
 	ASSERT(L1 ~= list(1, 2, 1, 2))
 
-	L2 = list(a=1, b=2)
+	var/L2 = list(a=1, b=2)
 	L2 += L2
 	ASSERT(L2 ~= list(a=1, b=2, "a", "b"))
--- a/Content.Tests/DMProject/Tests/List/ListAppendSelf.dm
+++ b/Content.Tests/DMProject/Tests/List/ListAppendSelf.dm
@@ -1,7 +1,7 @@
 /proc/RunTest()
 	L1 = list(1, 2)
 	L1 += L1
-	ASSERT(L1 = list(1, 2, 1, 2)
+	ASSERT(L1 ~= list(1, 2, 1, 2))
 
 	L2 = list(a=1, b=2)
 	L2 += L2
--- a/Content.Tests/DMProject/Tests/List/ListAppendSelf.dm
+++ b/Content.Tests/DMProject/Tests/List/ListAppendSelf.dm
@@ -1,0 +1,8 @@
+/proc/RunTest()
+	L1 = list(1, 2)
+	L1 += L1
+	ASSERT(L1 = list(1, 2, 1, 2)
+
+	L2 = list(a=1, b=2)
+	L2 += L2
+	ASSERT(L2 = list(a=1, b=2))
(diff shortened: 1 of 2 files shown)


Content.Tests/DMProject/Tests/List/ListAppendSelf.dm
```diff
@@ -0,0 +1,8 @@
+/proc/RunTest()
+	L1 = list(1, 2)
+	L1 += L1
+	ASSERT(L1 = list(1, 2, 1, 2)
```
wixoaGit, line 4: This is missing a closing parenthesis. You're also assigning here, not checking for equality.

`==` wouldn't be correct here either since that checks for reference equality with lists. You want `~=` which checks for equality of the contents.
```suggestion
	ASSERT(L1 ~= list(1, 2, 1, 2))
```

Content.Tests/DMProject/Tests/List/ListAppendSelf.dm
```diff
@@ -0,0 +1,8 @@
...
+
+	L2 = list(a=1, b=2)
+	L2 += L2
+	ASSERT(L2 = list(a=1, b=2))
```
wixoaGit, line 8: Same issue with `=` vs `~=`. This is also incorrect, BYOND results in `list(a=1, b=2, "a", "b")`.
```suggestion
	ASSERT(L2 ~= list(a=1, b=2))
```

OpenDreamRuntime/Objects/Types/DreamList.cs
```diff
@@ -297,6 +297,7 @@ public override DreamValue OperatorOr(DreamValue b, DMProcState state) {
 
     public override DreamValue OperatorAppend(DreamValue b) {
         if (b.TryGetValueAsDreamList(out var bList)) {
+            if (bList == this) bList = this.CreateCopy();
```
wixoaGit, line 300: Creating a copy of the list that then gets thrown out is wasteful, especially with large lists (which I imagine is a common operation). You could instead do this:
```cs
var values = bList.GetValues();
var valueCount = values.Count; // Some lists return a reference to their internal values list which could change with each loop
for (int i = 0; i < valuesCount; i++) {
    var value = values[i];
}
```

Another issue I found while testing this is that values get added even when they have an associated value. `SetValue()` doesn't do this because the key already exists, so this should be done inside the loop instead:
```cs
AddValue(value); // Always add the value
if (bList._associativeValues?.TryGetValue(value, out var assocValue) is true) { // Ensure the associated value is correct
    _associativeValues ??= new();
    _associativeValues[value] = assocValue;
}
```

wixoaGit (changes requested):

OpenDreamRuntime/Objects/Types/DreamList.cs
ritorizo: Thank you, the changes are committed. But can `_associativeValues` ever be null inside the if statement  since we execute if only if we find a correct key ?

OpenDreamRuntime/Objects/Types/DreamList.cs
```diff
@@ -574,9 +577,9 @@ public override int GetLength() {
 // atom's verbs list
 // Keeps track of an appearance's verbs (atom.verbs, mutable_appearance.verbs, etc)
 public sealed class VerbsList(DreamObjectTree objectTree, AtomManager atomManager, ServerVerbSystem? verbSystem, DreamObjectAtom atom) : DreamList(objectTree.List.ObjectDefinition, 0) {
-   public override DreamValue GetValue(DreamValue key) {
+    public override DreamValue GetValue(DreamValue key) {
```
ritorizo, line 580: Also sorry about that my IDE do it automatically ;-;

OpenDreamRuntime/Objects/Types/DreamList.cs
wixoaGit: The if statement checks `bList._associativeValues` and uses `_associativeValues` inside. If you're appending the list to itself then these two are the same, otherwise they wouldn't be.

OpenDreamRuntime/Objects/Types/DreamList.cs
wixoaGit: The change is correct, this should have always been 4 spaces

wixoaGit (approved):

PR #2157, opened 2025-01-06
Look for ressources in the dirrectory of the environement

### LINKED ISSUE #2077: There should be an implicit `#define FILE_DIR .`
creating an environment without `#define FILE_DIR .` and referencing an icon in a subdir throws a file not found on OD, but not on BYOND.

ritorizo:
Fix #2077 by adding the dirrectory of the dme to the ressource directories, therefore making it unecessary to declare `#define FILE_DIR' in the dme.


TestGame/environment.dme
```diff
@@ -1,8 +1,5 @@
 // BEGIN_INTERNALS
 // END_INTERNALS
-// BEGIN_FILE_DIR
-#define FILE_DIR .
-// END_FILE_DIR
```
wixoaGit, line 5: These lines are generated by the Dream Maker IDE and should be left there so we aren't fighting it.

Content.Tests/DMProject/environment.dme
```diff
@@ -1,8 +1,5 @@
 // BEGIN_INTERNALS
 // END_INTERNALS
-// BEGIN_FILE_DIR
-#define FILE_DIR .
-// END_FILE_DIR
```
wixoaGit, line 5: Same here

DMCompiler/DMCompiler.cs
```diff
@@ -128,9 +128,13 @@ public void AddResourceDirectory(string dir) {
                 preproc.IncludeFile(includeDir, fileName, false);
             }
 
+            // Adds the root of the DM project to FILE_DIR
+            compiler.AddResourceDirectory(Path.GetDirectoryName(files[0]));
```
wixoaGit, line 132: `GetDirectoryName()` could return null if the path is in the root directory. Null coalesce it to `"/"`.
```suggestion
            compiler.AddResourceDirectory(Path.GetDirectoryName(files[0]) ?? "/");
```

DMCompiler/DMCompiler.cs
```diff
@@ -128,9 +128,13 @@ public void AddResourceDirectory(string dir) {
...
+
             string compilerDirectory = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location) ?? string.Empty;
             string dmStandardDirectory = Path.Join(compilerDirectory, "DMStandard");
 
+
```
wixoaGit, line 137: Unnecessary empty line
```suggestion
```

DMCompiler/DMCompiler.cs
ritorizo: Thanks I'll read the function reference next time ^^'

wixoaGit (approved):

PR #2162, opened 2025-01-08
Quickfix of a bug introduced in #2157

amylizzle:
It appears this slipped through OD's test compile of the codebases and only failed on the actual codebase's tests themselves.


amylizzle:
nope

PR #2167, opened 2025-01-14
Server hop support in `link()`

ike709:
In BYOND, calling `link()` with a `byond://` server address will cause the client to connect to that server. Similarly, we now support connecting to other RT/SS14 servers via `ss14://` and `ss14s://`

Note that you _must_ join the initial game via the SS14 launcher for server hop (redial) to work; RT doesn't allow it from e.g. your dev TestGame server that you connected to directly from the client.

https://github.com/user-attachments/assets/f1639e50-e7df-4d0b-a3e3-8ae6736a3541

### DIFF
--- a/OpenDreamClient/Interface/DreamInterfaceManager.cs
+++ b/OpenDreamClient/Interface/DreamInterfaceManager.cs
@@ -41,6 +41,7 @@ internal sealed class DreamInterfaceManager : IDreamInterfaceManager {
     [Dependency] private readonly IRobustRandom _random = default!;
     [Dependency] private readonly ITimerManager _timerManager = default!;
     [Dependency] private readonly IUriOpener _uriOpener = default!;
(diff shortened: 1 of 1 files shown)


OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -270,8 +271,10 @@ private void RxLink(MsgLink message) {
         // TODO: This can be a topic call or a connection to another server
         if (uri.Scheme is "http" or "https") {
             _uriOpener.OpenUri(message.Url);
+        } else if (uri.Scheme is "ss14" or "ss14s") {
```
wixoaGit, line 274: You can check `_gameController.LaunchState.FromLauncher` to display a message if it can't redial.

wixoaGit (changes requested):

wixoaGit (approved):

PR #2174, opened 2025-01-17
Fix text box on goon by fixing element lookup

amylizzle:
`winclone()` was failing because lookup was failing because we were using `Name` instead of `Id` for the search.
image


OpenDreamClient/Interface/Descriptors/InterfaceDescriptor.cs
```diff
@@ -18,7 +18,7 @@ public InterfaceDescriptor(List<WindowDescriptor> windowDescriptors, List<MacroS
 
     public ElementDescriptor? GetElementDescriptor(string name) {
```
ZeWaka, line 19: Rename this?

wixoaGit (approved):

PR #2191, opened 2025-01-28
Avoid recalculating tile visibility when nothing is changing

wixoaGit:
On my less-powerful laptop this shaves off many milliseconds on most frames. It also saves on kilobytes of memory allocations per frame.

### DIFF
--- a/OpenDreamClient/Rendering/DreamViewOverlay.cs
+++ b/OpenDreamClient/Rendering/DreamViewOverlay.cs
@@ -21,7 +21,7 @@ namespace OpenDreamClient.Rendering;
 /// <summary>
 /// Overlay for rendering world atoms
 /// </summary>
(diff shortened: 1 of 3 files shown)


OpenDreamClient/Rendering/DreamViewOverlay.cs
```diff
@@ -21,7 +21,7 @@ namespace OpenDreamClient.Rendering;
 /// <summary>
 /// Overlay for rendering world atoms
 /// </summary>
-internal sealed class DreamViewOverlay : Overlay {
+internal sealed partial class DreamViewOverlay : Overlay, IEntityEventSubscriber {
```
wixoaGit, line 24:
```suggestion
internal sealed partial class DreamViewOverlay : Overlay {
```

PR #2196, opened 2025-01-31
mark `issaved()` as implemented

amylizzle:
it is, we've got a test for it and everything


wixoaGit:
This proc should be removed from the DM entirely. `issaved()` is implemented in bytecode, the proc does nothing.

DMCompiler/DMStandard/UnsortedAdditions.dm
```diff
@@ -21,7 +21,6 @@
 /proc/generator(type, A, B, rand)
 	set opendream_unimplemented = TRUE
 /proc/issaved(v)
```
amylizzle, line 23:
```suggestion
```

wixoaGit (approved):

Issue #2194, opened 2025-01-31
Unimplemnted Procs and Vars Master Issue


amylizzle:
## Global Functions
- [ ]   /proc/flick(Icon, Object)
- [ ]   /proc/noise_hash(...) as num
- [ ]   /proc/walk_to(Ref, Trg, Min = 0, Lag = 0, Speed = 0)
- [ ]   /proc/replacetextEx_char(Haystack as text, Needle, Replacement, Start = 1, End = 0) as text
- [ ]   /proc/walk_away(Ref,Trg,Max=5,Lag=0,Speed=0)
- [ ]   ﻿/proc/bounds(Ref=src, Dist=0)
- [ ]   /proc/bounds_dist(Ref, Target)
- [ ]   /proc/cmptextEx(T1)
- [ ]   /proc/eval(script)
- [ ]   /proc/findlasttext_char(Haystack,Needle,Start=0,End=1)
- [ ]   /proc/findlasttextEx_char(Haystack,Needle,Start=1,End=0)
- [ ]   /proc/findtext_char(Haystack,Needle,Start=1,End=0)
- [ ]   /proc/findtextEx_char(Haystack,Needle,Start=1,End=0)
- [ ]   /proc/generator(type, A, B, rand)
- [ ]   /proc/issaved(v)
- [ ]   /proc/load_resource(File)
- [ ]   proc/missile(Type, Start, End)
- [ ]   /proc/obounds(Ref=src, Dist=0)
- [ ]   /proc/replacetext_char(Haystack,Needle,Replacement,Start=1,End=0)
- [ ]   /proc/run(File)
- [ ]   /proc/shell(command)
- [ ]   /proc/nonspantext_char(Haystack,Needles,Start=1)
- [ ]   /proc/splittext_char(Text,Start=1,End=0,Insert="")
- [ ]   /proc/_dm_db_new_con()
- [ ]   /proc/_dm_db_connect()
- [ ]   /proc/_dm_db_close()
- [ ]   /proc/_dm_db_is_connected()
- [ ]   /proc/_dm_db_quote()
- [ ]   /proc/_dm_db_new_query()
- [ ]   /proc/_dm_db_execute()
- [ ]   /proc/_dm_db_next_row()
- [ ]   /proc/_dm_db_rows_affected()
- [ ]   /proc/_dm_db_row_count()
- [ ]   /proc/_dm_db_error_msg()
- [ ]   /proc/_dm_db_columns()

## /atom
[...]

PR #2201, opened 2025-02-03
Implement `list.Splice()`

ike709:
https://github.com/OpenDreamProject/OpenDream/issues/2194

https://www.byond.com/docs/ref/#/list/proc/Splice


Content.Tests/DMProject/Tests/List/ListSplice.dm
wixoaGit, line 1: Can we also add a `Splice()` call that has a list as one of its items

wixoaGit (approved):

PR #2202, opened 2025-02-03
Implement `/sound.offset`

ike709:
https://www.byond.com/docs/ref/#/sound/var/offset

https://github.com/OpenDreamProject/OpenDream/issues/2194

Tested it to confirm it works. Didn't test what BYOND does when given non-numbers.


OpenDreamRuntime/DreamConnection.cs
```diff
@@ -212,11 +212,13 @@ public void OutputDreamValue(DreamValue value) {
         if (value.TryGetValueAsDreamObject<DreamObjectSound>(out var outputObject)) {
             ushort channel = (ushort)outputObject.GetVariable("channel").GetValueAsInteger();
             ushort volume = (ushort)outputObject.GetVariable("volume").GetValueAsInteger();
+            int offset = (int)outputObject.GetVariable("offset").UnsafeGetValueAsFloat();
```
wixoaGit, line 215: The DM reference says this value is in seconds. Does this not take a float for fraction-of-a-second values?

wixoaGit (changes requested):

wixoaGit (approved):

PR #2207, opened 2025-02-05
Implement `sign()`

ike709:
https://github.com/OpenDreamProject/OpenDream/issues/2203

https://www.byond.com/docs/ref/#/proc/sign

PR #2209, opened 2025-02-08
Pretend to be 516 for Nebula compilation

ike709:
Resolves the CI failure in https://github.com/OpenDreamProject/OpenDream/pull/2207


ike709:
Adding this to https://github.com/OpenDreamProject/OpenDream/pull/2207 instead so CI actually passes.

PR #2210, opened 2025-02-09
Bump dotnet version in forgotten files

Furrior:
I think these were missed in #2165. Also adjusted the readme a bit to follow common md styling

### DIFF
--- a/DMCompiler/copy_standard.bat
+++ b/DMCompiler/copy_standard.bat
@@ -1,5 +1,5 @@
 @echo off
-if not exist bin\Debug\net8.0\DMStandard mkdir bin\Debug\net8.0\DMStandard
-xcopy DMStandard bin\Debug\net8.0\DMStandard /y /s /e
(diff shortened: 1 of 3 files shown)


DMCompiler/copy_standard.bat
wixoaGit, line 1: These paths are actually even more out-of-date than this. The DMCompiler's build path was changed a while ago to be `../bin/DMCompiler/` to match the rest of the client/server projects.

wixoaGit (changes requested):

DMCompiler/copy_standard.bat
Furrior: Suspected so, but thought that the person who actually uses these scripts also uses custom build paths

wixoaGit: Turns out no one's been using these for over a year. They can just be deleted.

wixoaGit (approved):

Issue #2203, opened 2025-02-03
516 Support Master Issue


amylizzle:
BYOND 516 mostly just upgrades the browser elements to Webview2, but also adds a few new features:

- [ ] `/vector`
- [ ] new `screen_loc` keywords: `SCREEEN_SOUTHWEST`, `SCREEN_NORTHEAST`, etc
- [ ] `sound.atom` and `sound.transform` for moving sounds
- [ ] `icon_w` and `icon_z`
- [ ] some list operations for doing fast math on assoc lists `values_sum`, `values_product`, etc.
- [ ] `lerp()`
- [ ] `sign()`
- [ ] `/callee` in error handling
- [ ] `operator<=>`
- [ ] ~~asslist~~ `alist` - it's a dictionary. Why isn't it called `dict`? Nobody knows.
- [ ] `#pragma syntax C` ?????
- [ ] `for(k,v in list)` syntax
- [ ] `load_ext()`
- [ ] `astype()`
- [ ] `animate()` now supports a tag argument for named animation sequences
- [ ] `/pixloc` - seems to just be wrapper for `pos + (step_x, step_y)`
- [ ] `/atom.pixloc` 
- [ ] `filter()` now supports naming of filters and named filter lookup
- [ ] `The ~= and ~! operators now take associated values into account when checking associative lists`
- [ ] Byondapi gets `Byond_CRASH()`
- [ ] `.sound` client-side command
- [ ] `SIDE_MAP` and `ISOMETRIC_MAP` have had changes to their positioning behavior

TiviPlus:
mouse-pos (516.1651) and mousex (undocumented, https://www.byond.com/forum/post/2902663) missing

PR #2212, opened 2025-02-11
Implement 516 behavior for `~=` and `~!`

ike709:
https://github.com/OpenDreamProject/OpenDream/issues/2203

### DIFF
--- a/OpenDreamRuntime/Objects/Types/DreamList.cs
+++ b/OpenDreamRuntime/Objects/Types/DreamList.cs
@@ -372,12 +372,23 @@ public class DreamList : DreamObject {
             return DreamValue.False;
         if (GetLength() != secondList.GetLength())
             return DreamValue.False;
+        if(IsAssociative && (!secondList.IsAssociative || secondList.GetAssociativeValues().Count != GetAssociativeValues().Count))
+            return DreamValue.False;
 
         var firstValues = GetValues();
         var secondValues = secondList.GetValues();
         for (var i = 0; i < firstValues.Count; i++) {
             if (!firstValues[i].Equals(secondValues[i]))
                 return DreamValue.False;
+        }
+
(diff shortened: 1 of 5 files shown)


OpenDreamRuntime/Objects/Types/DreamList.cs
```diff
@@ -372,6 +372,8 @@ public override DreamValue OperatorEquivalent(DreamValue b) {
             return DreamValue.False;
         if (GetLength() != secondList.GetLength())
             return DreamValue.False;
+        if(IsAssociative && (!secondList.IsAssociative || secondList.GetAssociativeValues().Count != GetAssociativeValues().Count))
```
wixoaGit, line 375: `list(a=null) ~= list("a")` is `TRUE` so returning early if `IsAssociative != secondList.IsAssociative` is incorrect. At least as long as `IsAssociative` doesn't take all-null associative lists into account.

OpenDreamRuntime/Objects/Types/DreamList.cs
```diff
@@ -380,6 +382,15 @@ public override DreamValue OperatorEquivalent(DreamValue b) {
                 return DreamValue.False;
         }
 
+        // Starting with 516, equivalence checks assoc values
+        if (IsAssociative) {
+            var secondListAssoc = secondList.GetAssociativeValues();
+            foreach (var kvp in GetAssociativeValues()) {
+                if(!secondListAssoc.TryGetValue(kvp.Key, out var assocValue) || !assocValue.Equals(kvp.Value))
```
wixoaGit, line 389: The lack of a value in either assoc lists should be treated as null.

wixoaGit (changes requested):

wixoaGit (approved):

Issue #2187, opened 2025-01-25
HTML parser confused by `<br>` not needing a closing tag


amylizzle:
`[ERRO] opendream.html_parser: Invalid closing tag </span>, expected </br>`

PR #2217, opened 2025-02-17
Adds html parser check for void elements

Ruzihm:
#2187

### DIFF
--- a/OpenDreamClient/Interface/Html/HtmlParser.cs
+++ b/OpenDreamClient/Interface/Html/HtmlParser.cs
@@ -84,7 +84,8 @@ public static class HtmlParser {
                     } else {
                         tags.Push(tagType);
 
-                        appendTo.PushTag(new MarkupNode(tagType, null, ParseAttributes(attributes)), selfClosing: attributes[^1] == "/");
+                        bool isSelfClosing = IsSelfClosing(tagType, attributes);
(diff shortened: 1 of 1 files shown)


OpenDreamClient/Interface/Html/HtmlParser.cs
```diff
@@ -139,6 +140,39 @@ void PushCurrentText() {
...
+     * Returns if a tag is written in old self-closing form, or if the tag
+     * represents a void element, which must have no children
+     * </summary>
+     */
+    private static bool IsSelfClosing(string tagType, string[] attributes) {
+        if (attributes[^1] == "/") {
+            return true;
+        }
+
+        switch (tagType) {
```
wixoaGit, line 154: According to [this page of the DM reference](https://www.byond.com/docs/ref/#/DM/text/tags), the only tags in here that the client supports are `<br>` and ``. Though it doesn't hint at this format of `` being supported, so I'm not sure if that should be here either.

wixoaGit (changes requested):

OpenDreamClient/Interface/Html/HtmlParser.cs
wixoaGit: Though it's good to be able to parse these, so this is fine.

wixoaGit (approved):

Ruzihm:
Will have to continue on this - the problem still isn't fixed.

Issue #2062, opened 2024-10-26
Pointless `Jump` instructions


ike709:
This sort of pattern shows up in TG:
```
    Jump 0x43d
0x43d
```

Need to optimize these out in the bytecode optimizer.

PR #2218, opened 2025-02-17
Additional label optimization

Ruzihm:
Re-runs label optimizations again after peephole optimizations to take advantage of any removed jumps.

Also adds an optimization which removes any jumps that point to an immediately following bytecode label.

#2062


Ruzihm:
Got pulled away after accidentally committing unnecessary local changes. Will remove when back

Ruzihm:
Got pulled away after accidentally committing unnecessary local changes. Will remove when back

PR #2219, opened 2025-02-18
Particles

### LINKED ISSUE #2143: `/particle` is not implemented
We have the typedef but no functionality. Note that this will probably benefit from some in-engine shenanigans for efficiency's sake and also so SS14 can have particles too.

amylizzle:
PR'ing cos I'm a good chunk of the way there and it's easier to keep track of

Relies on https://github.com/space-wizards/RobustToolbox/pull/5653

Fixes #2143 

#2194


amylizzle:
TODO: remove gooncode and sprites from this for license stuff

wixoaGit:
I haven't done a thorough look through the code yet, but I tried testing this by booting up Paradise and shooting a girder. This should create some particles, but it broke the renderer instead. Some planes stopped rendering and there were errors every frame.

amylizzle:
Ah that's because I hadn't implemented the generator type and had it throw instead - my bad, I just need to figure out the math and implement

Issue #1738, opened 2024-03-27
Goon: water coolers are not transparent


amylizzle:
![image](https://github.com/OpenDreamProject/OpenDream/assets/3855802/a9b93c5b-ec24-4880-a6ce-953ae6694572)

They're supposed to be.

wixoaGit:
#2222 Gave them some transparency, though I don't know if they're usually even more transparent than this.

![Image](https://github.com/user-attachments/assets/64d2d613-5dc8-4f98-bc82-e3a8c91af63e)

wixoaGit:
![Image](https://github.com/user-attachments/assets/cfc08fe5-54ee-4518-a7e9-59a84638ca0e)

PR #2222, opened 2025-02-20
Updates atom alpha when atom color is set

Ruzihm:
When setting atom color (but not with matrix), update alpha variable with alpha channel of color

#1738


wixoaGit:
Do you have an example of what this is fixing?

Ruzihm:
From the discord:

> amylizzle: oh I guess atom.alpha is just an alias for atom.color.A

So this isn't appropriate. Going to mark as draft and revisit.

wixoaGit (approved):

Issue #1490, opened 2023-10-14
Server contains DMCompiler in its build output


wixoaGit:
It would be preferred to keep DMCompiler outside of OpenDreamServer's build output. It's currently in there because OpenDreamServer has DMCompiler as a dependency.

PR #2226, opened 2025-02-23
Refactor common code

Ruzihm:
Puts code common among server, client, compiler into an OpenDreamCommon assembly
Puts code common among only server and compiler into a NoClient assembly (currently just for the MD5 call)
Removes compiler from server (#1490)
Deduplicates DreamValueType and DMValueType into just DMValueType. (these were different for some reason)


OpenDreamCommon/Bytecode/OpcodeMetadata.cs
```diff
@@ -1,10 +1,10 @@
-namespace DMCompiler.Bytecode;
+namespace OpenDreamShared.Common.Bytecode;
 
 /// <summary>
 /// Custom attribute for declaring <see cref="OpcodeMetadata"/> metadata for individual opcodes
 /// </summary>
 [AttributeUsage(AttributeTargets.Field)]
-internal sealed class OpcodeMetadataAttribute : Attribute {
+public sealed class OpcodeMetadataAttribute : Attribute {
```
Ruzihm, line 7: This change was to allow referencing `OpcodeMetadataAttribute` into this common assembly class from the compiler assembly. Let me know if there's a better approach.

Ruzihm:
worried for a sec but it looks like goon broke the compilation https://github.com/goonstation/goonstation/commit/8f6e2a65c5ca08856fbfea583163bbb9718b1979#diff-c082a149de1413ea394f478403100935075b0f4dfd201395a6876e667d1a51eaR989🫢

wixoaGit:
Thinking about this more, I don't think this really makes sense to do. Feels like project bloat, and the server will likely have to reference DMCompiler anyway if we want to add things like expression evaluation to the DM debugger.

Sorry to have wasted your time with this.

PR #2227, opened 2025-02-23
Adds check for local variable in scope operator check

Ruzihm:
Adds check for local variable in scope operator check. 

Example code from goon:

```
proc/generate_copyrighted_colors()
	src.copyrighted_colors = list()
	for (var/obj/item/clothing/head/type as anything in src.copyrighted_berets)
		var/icon/icon = icon(type::icon, type::icon_state) // <-- type now interpreted as a local var
		copyrighted_colors[global.get_average_color(icon)] = rand(10000, 999999)
```


DMCompiler/DM/Builders/DMExpressionBuilder.cs
```diff
@@ -645,8 +645,13 @@ private DMExpression BuildScopeIdentifier(DMASTScopeIdentifier scopeIdentifier,
...
             // We want to give end users an explanation at least.
             if (scopeMode is Normal && ctx.Proc != null)
-                return BadExpression(WarningCode.BadExpression, identifier.Location,
-                    "Use of \"type::\" and \"parent_type::\" outside of a context is forbidden");
+                if (ctx.Proc.GetLocalVariable(identifier.Identifier) != null) {
+                    // actually - it's referring to a local variable named "type" or "parent_type"... just do the usual thing
+                    expression = BuildExpression(scopeIdentifier.Expression, inferredPath);
+                } else {
+                    return BadExpression(WarningCode.BadExpression, identifier.Location,
+                            "Use of \"type::\" and \"parent_type::\" outside of a context is forbidden");
```
amylizzle, line 653:
```suggestion
                            "Use of \"type::\" and \"parent_type::\" outside of an object proc definition is forbidden");
```
make the error message a little clearer maybe?

amylizzle (changes requested):
lgtm, but add a unit test

DMCompiler/DM/Builders/DMExpressionBuilder.cs
Ruzihm: @amylizzle  I'm still learning DM but since this fails to compile (byond):

```
/datum/foo
	var/price = 60
	better
		proc/test_proc() 
			price = parent_type::price + 40
```
with:
```
error: parent_type::price: undefined var
error: ::: bad var or proc() after :: operator
```

but no errors here:

```
/datum/foo
	var/price = 60
	better
		price = parent_type::price + 40
```

I'm inclined to have the error be `Use of "type::" or "parent_type::" inside a proc definition requires a local variable with the name "type" or "parent_type"` or something along those lines.

amylizzle: huh, I guess I misunderstood what that does? Yeah that's a better message then!

Content.Tests/DMProject/Tests/Operators/scope.dm
```diff
@@ -52,10 +57,12 @@ var/static/one = "one"
...
+	// this does not compile in BYOND, that is just a bug
 	test::three::four.five::six = "7 8 9 10"
 	ASSERT(test::three::four.five::six == "7 8 9 10")
 	ASSERT(/datum::three::four.five::six == "7 8 9 10")
-	
+
 	var/datum/three/threetest = new
 	threetest.typetest()
+
+	var/datum/six/better/test
```
amylizzle, line 68:
```suggestion
	var/datum/six/better/test = new()
	ASSERT(test.price == 100)
```

Content.Tests/DMProject/Tests/Operators/scope_parent_type_fails.dm
```diff
@@ -0,0 +1,10 @@
+// COMPILE ERROR OD0011
+
+/datum/thing
+	var/price = 60
+	better
+		proc/test_proc()
+			price = parent_type::price + 40
+
+/proc/RunTest()
+	var/datum/thing/better/test
```
amylizzle, line 10:
```suggestion
```
doesn't do anything

amylizzle (changes requested):

Content.Tests/DMProject/Tests/Operators/scope_parent_type_fails.dm
Ruzihm: Good point - apparently I can also get rid of `/proc/RunTest()` entirely and it works as expected

confirmed by changing to `// COMPILE ERROR OD0012` and the test case fails.

amylizzle (approved):

DMCompiler/DM/DMExpression.cs
```diff
@@ -138,10 +138,10 @@ private void VerifyArgType(DMCompiler compiler, DMProc targetProc, int index, st
     }
 }
 
-internal readonly struct ExpressionContext(DMCompiler compiler, DMObject type, DMProc proc) {
+internal readonly struct ExpressionContext(DMCompiler compiler, DMObject type, DMProc? proc) {
     public readonly DMCompiler Compiler = compiler;
     public readonly DMObject Type = type;
-    public readonly DMProc Proc = proc;
+    public readonly DMProc? Proc = proc;
```
wixoaGit, line 144: This introduces a million nullability warnings elsewhere. There are only 3 cases where an ExpressionContext is created with a null `proc` argument. I think it would be better to instead provide the type's `<init>` proc for those.

I'd prefer that to be done in another PR though.

DMCompiler/DM/Builders/DMExpressionBuilder.cs
```diff
@@ -644,11 +644,16 @@ private DMExpression BuildScopeIdentifier(DMASTScopeIdentifier scopeIdentifier,
...
             // We want to give end users an explanation at least.
-            if (scopeMode is Normal && ctx.Proc != null)
-                return BadExpression(WarningCode.BadExpression, identifier.Location,
-                    "Use of \"type::\" and \"parent_type::\" outside of a context is forbidden");
-
-            if (identifier.Identifier == "parent_type") {
+            if (scopeMode is Normal && ctx.Proc != null) {
+                if (ctx.Proc.GetLocalVariable(identifier.Identifier) != null) {
+                    // actually - it's referring to a local variable named "type" or "parent_type"... just do the usual thing
+                    expression = BuildExpression(scopeIdentifier.Expression, inferredPath);
```
wixoaGit, line 650: We should emit a warning for this, since it's likely unintentional.

wixoaGit (changes requested):

DMCompiler/DM/Builders/DMExpressionBuilder.cs
Ruzihm: I did `SoftResolvedKeyword` which made goon [fail to compile](https://github.com/OpenDreamProject/OpenDream/actions/runs/13489680448/job/37685809461?pr=2227#step:16:19) due to its `#pragma SoftReservedKeyword error`. Let me know if I should pick something else.

Ruzihm: After discussing, went with a new and specific `OD2001: ScopeOperandNamedType`

wixoaGit (approved):

PR #2239, opened 2025-03-07
Enhancement: `vis_contents_plane_offset`

amylizzle:
Adds a new var to `/atom`: `vis_contents_plane_offset`

Defaults to 0. Applies a simple offset to the plane of items in that atom's `vis_contents`

Requested by TheFinalPotato for cursed multi-z shit

### DIFF
--- a/DMCompiler/DMStandard/Types/Atoms/_Atom.dm
+++ b/DMCompiler/DMStandard/Types/Atoms/_Atom.dm
@@ -62,6 +62,8 @@
 	var/render_target
 	var/vis_flags as opendream_unimplemented
 
+	var/vis_contents_plane_offset = 0 //opendream only
+
 	proc/Click(location, control, params)
 	proc/MouseDrop(over_object,src_location,over_location,src_control,over_control,params)
 	proc/MouseEntered(location,control,params)
--- a/OpenDreamClient/Rendering/DreamViewOverlay.cs
+++ b/OpenDreamClient/Rendering/DreamViewOverlay.cs
@@ -174,7 +174,7 @@ internal sealed partial class DreamViewOverlay : Overlay {
     }
 
     //handles underlays, overlays, appearance flags, images. Adds them to the result list, so they can be sorted and drawn with DrawIcon()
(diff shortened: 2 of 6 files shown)


LemonInTheDark:
does this work on overlays and such? also they said it couldn't be done, good god, etc etc etc

amylizzle:
If the thing in `vis_contents` has overlays, they will also be offset by this. It's functionally equivalent to setting the plane on the atom in `vis_contents`

LemonInTheDark:
> If the thing in `vis_contents` has overlays, they will also be offset by this. It's functionally equivalent to setting the plane on the atom in `vis_contents`

those two aren't the same though, right?
like
```
berry.plane = 10 
bush.overlays += berry
```
the berry's plane here would be uneffected by setting the plane of the bush

sorry if I'm being persnickety here but I don't wanna give false impressions of what I'm trying to do/what this would be useful for.

amylizzle:
no, no precision is good and welcome.

As things currently stand in our render logic, if an overlay is applied to an atom the overlay gets this for it's plane: (in this case, `icon` is `berry` and `parentIcon` is `bush`
```
if ((icon.Appearance.Plane < -10000)) //FLOAT_PLANE - Note: yes, this really is how it works. Yes it's dumb as shit.
        current.Plane = parentIcon.Plane + (icon.Appearance.Plane + 32767);
    else
        current.Plane = icon.Appearance.Plane;
```

so yes, you're right, the plane of the overlay would only be affected if it was `FLOAT_PLANE`, otherwise it would still be 10 (in your example)

LemonInTheDark:
that's annoying. What I'm trying to do here is basically FLOAT_PLANE if it didn't suck (FLOAT_PLANE applied to a floated plane with offset inherits the offset value from its parent). I want a way to offset all the planes in a "chain" by the same amount is I suppose the best way of putting this. vis_contents is I think the best usecase for doing this (because it lets you setup "lenses" but it isn't inherently related. I don't wanna just torture you bout it if it's a lot of work that isn't a problem but I figure I should be clear.

PR #2240, opened 2025-03-08
bumps RT

Ruzihm:
Current RT submodule version has a CVE, bumping to the most recent gets rid of it and makes OD's github workflow happy.


wixoaGit:
This isn't set to a commit that bumps the RT version, which is required for the launcher to continue working.

Ruzihm:
oof. marking as a draft til then

wixoaGit:
Due to https://github.com/space-wizards/RobustToolbox/issues/5738 you'll also have to disable the main viewport in `EntryPoint.PostInit()`

SS14 does that with:
```cs
// Disable engine-default viewport since we use our own custom viewport control.
_userInterfaceManager.MainViewport.Visible = false;
```

Ruzihm:
@wixoaGit Ready for review when you have the time

wixoaGit (approved):

PR #2253, opened 2025-03-27
Make preprocessor dump skip build

Furrior:
Allows to debug preprocessor easier and to use OD as a fast preprocessor potentially


ike709 (changes requested):
I cannot articulate how incredibly annoying this would be to have to modify my Rider config every time I want to switch between preproc debugging and being able to compile.

Furrior:
> I cannot articulate how incredibly annoying this would be to have to modify my Rider config every time I want to switch between preproc debugging and being able to compile.

image
Probably a new arg is the way then

PR #2260, opened 2025-04-13
Implement `/proc/final/`

### LINKED ISSUE #2160: `proc/final` is not supported
Putting `final` in a proc path makes it not possible to override that proc. Putting this keyword on an override does nothing.

```dm
/datum/proc/final/foo()

/datum/a/foo() // Error here
```

https://www.byond.com/docs/ref/#/var/final

ike709:
Resolves #2160

Only difference from BYOND afaict is that I put both proc declaration locations in 1 error message instead of emitting 2 separate errors.

### DIFF
--- a/DMCompiler/DM/DMCodeTree.Procs.cs
+++ b/DMCompiler/DM/DMCodeTree.Procs.cs
@@ -32,13 +32,10 @@ internal partial class DMCodeTree {
             if (procDef.IsOverride) {
                 var procs = dmObject.GetProcs(procDef.Name);
                 if (procs != null) {
-                    foreach (var procId in procs) {
-                        var parent = compiler.DMObjectTree.AllProcs[procId];
-                        if (parent.IsFinal)
-                            compiler.Emit(WarningCode.FinalOverride, procDef.Location,
-                                $"Proc \"{procDef.Name}()\" is final and cannot be overridden. Final declaration: {parent.Location}");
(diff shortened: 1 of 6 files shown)


DMCompiler/DM/DMCodeTree.Procs.cs
```diff
@@ -29,6 +29,19 @@ public bool TryDefineProc(DMCompiler compiler) {
...
+            if (procDef.IsOverride) {
+                var procs = dmObject.GetProcs(procDef.Name);
+                if (procs != null) {
+                    foreach (var procId in procs) {
+                        var parent = compiler.DMObjectTree.AllProcs[procId];
+                        if (parent.IsFinal)
+                            compiler.Emit(WarningCode.FinalOverride, procDef.Location,
+                                $"Proc \"{procDef.Name}()\" is final and cannot be overridden. Final declaration: {parent.Location}");
+                        break;
+                    }
```
ike709, line 41:
```suggestion
                      var parent = compiler.DMObjectTree.AllProcs[procs[0]];
                      if (parent.IsFinal)
                          compiler.Emit(WarningCode.FinalOverride, procDef.Location,
                              $"Proc \"{procDef.Name}()\" is final and cannot be overridden. Final declaration: {parent.Location}");
```

wixoaGit (approved):

PR #2268, opened 2025-05-02
Fixes DMF text-color not applying to statpanel / output

taislin:
By default, text-color is being read from the .DMF file, but not being applied in parts of the UI, even though background color is.
This is problematic when the background color is dark, as text-color defaults to black.

This PR makes the UI system correctly apply text-color as well.

### DIFF
--- a/OpenDreamClient/Interface/Controls/ControlInfo.cs
+++ b/OpenDreamClient/Interface/Controls/ControlInfo.cs
@@ -49,6 +49,9 @@ internal sealed class StatPanel : InfoPanel {
             //       I couldn't find a way to do this without recreating the FormattedMessage
             ValueLabel.MouseFilter = MouseFilterMode.Stop;
             ValueLabel.OnKeyBindDown += OnKeyBindDown;
+            if (_owner.InfoDescriptor.TextColor.Value != Color.Black) {
+                _textColor = _owner.InfoDescriptor.TextColor.Value;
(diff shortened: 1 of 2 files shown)


OpenDreamClient/Interface/Controls/ControlInfo.cs
```diff
@@ -83,6 +83,7 @@ private void UpdateLabels() {
             _valueText.Clear();
 
             // Use the default color and font
+            SetTextColor(_owner.InfoDescriptor.TextColor.Value);
```
wixoaGit, line 86: Isn't this already covered by the `UpdateElementDescriptor()`? Was that not working?

wixoaGit (changes requested):

OpenDreamClient/Interface/Controls/ControlInfo.cs
taislin: Did not work for me, no. I don't think its being called in the status tab.

wixoaGit: Think I found the issue: if the StatEntry is created after the color was set then it keeps the default black.

Calling SetTextColor here ends up running UpdateLabel twice so it would be better to instead set the `_textColor` in StatEntry's constructor.

wixoaGit (approved):

PR #2279, opened 2025-05-17
Implements some of the vector procs and operators & correction initial values

TiviPlus:
implements
```
Dot()
Interpolate()
Normalize()
+
-
/
*
```

Implemented procs in softcode mostly cus thats what the byond ref listed

uses null for len and size


DMCompiler/DMStandard/Types/Vector.dm
```diff
@@ -1,7 +1,6 @@
 /vector
-	// TODO: Verify these default values
-	var/len = 2 as num
-	var/size = 0 as num
+	var/len = null as num
+	var/size = null as num
```
wixoaGit, line 3: Typemaker test are failing because `null` isn't a num.
```suggestion
	var/len = null as num|null
	var/size = null as num|null
```

OpenDreamRuntime/Objects/Types/DreamObjectVector.cs
```diff
@@ -84,6 +84,124 @@ public override void Initialize(DreamProcArguments args) {
...
+            };
+            return new DreamValue(output);
+        }
+
+        return base.OperatorMultiply(b, state);
+    }
+
+    public override DreamValue OperatorMultiplyRef(DreamValue b, DMProcState state) {
+        return OperatorMultiply(b, state);
+    }
```
wixoaGit, line 143: This is the `*=` operator which assigns the values to this vector instead of creating a new vector. This needs to assign the `X`, `Y`, and `Z` values instead of using `OperatorMultiply()`.

OpenDreamRuntime/Objects/Types/DreamObjectVector.cs
```diff
@@ -84,6 +84,124 @@ public override void Initialize(DreamProcArguments args) {
...
+            };
+            return new DreamValue(output);
+        }
+
+        return base.OperatorDivide(b, state);
+    }
+
+    public override DreamValue OperatorDivideRef(DreamValue b, DMProcState state) {
+        return OperatorDivide(b, state);
+    }
```
wixoaGit, line 175: This also needs to assign the values on this vector instead of using `OperatorDivide()`.

wixoaGit (changes requested):

wixoaGit (approved):

PR #2289, opened 2025-05-26
Rewrite repeating verbs to use VerbSystem, incl. overrides

HeyBanditoz:
* Repeating macros can now run any verb
* Repeating behavior closely mimicks BYOND, where the last sent repeated key is the one being repeated
* Verbs now execute its last override on an atom/client


OpenDreamRuntime/AtomManager.cs
```diff
@@ -389,6 +389,7 @@ public void SetAppearanceVar(MutableAppearance appearance, string varName, Dream
                         if (!verbValue.TryGetValueAsProc(out var verb))
                             continue;
 
+
```
wixoaGit, line 392:
```suggestion
```

OpenDreamRuntime/ServerVerbSystem.cs
```diff
@@ -130,35 +132,89 @@ private void OnPlayerStatusChanged(object? sender, SessionStatusEventArgs e) {
...
     }
 
-    private void OnVerbExecuted(ExecuteVerbEvent msg, EntitySessionEventArgs args) {
-        var connection = _dreamManager.GetConnectionBySession(args.SenderSession);
-        var src = _dreamManager.GetFromClientReference(connection, msg.Src);
-        if (src == null || !_verbIdToProc.TryGetValue(msg.VerbId, out var verb) || !CanExecute(connection, src, verb))
-            return;
+    public void RunRepeatingVerbs() {
+        using (Profiler.BeginZone("Repeating Verbs", color: (uint)Color.OrangeRed.ToArgb())) {
+            foreach (var repeatingVerb in _repeatingVerbs) {
```
wixoaGit, line 137: This should pull out the contents of `OnVerbExecuted()` into a shared method and use that instead to execute verbs, to ensure behavior is consistent and doesn't copy code.

OpenDreamRuntime/ServerVerbSystem.cs
```diff
@@ -130,35 +132,89 @@ private void OnPlayerStatusChanged(object? sender, SessionStatusEventArgs e) {
...
+                if (repeatingVerb.Value.Count == 0)
+                    return;
+                var client = repeatingVerb.Key;
+                var verbId = repeatingVerb.Value.Last();
+
+                var src = _dreamManager.GetFromClientReference(client, ClientObjectReference.Client);
```
wixoaGit, line 143: This assumes the verb is always on the client. It needs to use `RegisterRepeatVerbEvent`'s `Src` instead of `ClientObjectReference.Client`.

OpenDreamRuntime/ServerVerbSystem.cs
```diff
@@ -130,35 +132,89 @@ private void OnPlayerStatusChanged(object? sender, SessionStatusEventArgs e) {
...
+    }
+
+    private void OnRepeatVerbStart(RegisterRepeatVerbEvent msg, EntitySessionEventArgs args) {
+        if (!_verbIdToProc.ContainsKey(msg.VerbId))
             return;
+        var conn = _dreamManager.GetConnectionBySession(args.SenderSession);
+        if (_repeatingVerbs.TryGetValue(conn, out var verb)) {
+            if (!verb.Remove(msg.VerbId)) {
+                verb.Add(msg.VerbId);
+            }
```
wixoaGit: Why does this _stop_ executing the verb if it's already in the list?

OpenDreamRuntime/ServerVerbSystem.cs
```diff
@@ -130,35 +132,89 @@ private void OnPlayerStatusChanged(object? sender, SessionStatusEventArgs e) {
...
+        } else {
+            List<int> list = new();
+            _repeatingVerbs.Add(conn, list);
+            list.Add(msg.VerbId);
         }
```
wixoaGit: A little nicer, in my opinion

```suggestion
        
        var conn = _dreamManager.GetConnectionBySession(args.SenderSession);
        if (!_repeatingVerbs.TryGetValue(conn, out var list)) {
            list = new();
            _repeatingVerbs.Add(conn, list);
        }
        
        list.Add(msg.VerbId);
```

wixoaGit (changes requested):

OpenDreamRuntime/ServerVerbSystem.cs
HeyBanditoz: I opted not to for this, since `OnVerbExecuted` handles argument checking, of which verbs wouldn't have any, so we can skip that block entirely.
I also wanted to distinct regular verbs and repeating verbs in the name (1st param) of the `DreamThread#Run` call.

OpenDreamRuntime/ServerVerbSystem.cs
HeyBanditoz: Whoops, meant for `Contains` here, thanks.

HeyBanditoz:
Superseded by #2365

PR #2295, opened 2025-06-02
Basic `alist()` support

wixoaGit:
A basic implementation of `alist()` that allows creating them, setting values, getting values, and `json_encode()`ing them.

An `IDreamList` interface was added, and should in the future be used for anything that allows using either `/list` or `/alist`.

### DIFF
--- a/OpenDreamRuntime/Objects/DreamObjectTree.cs
+++ b/OpenDreamRuntime/Objects/DreamObjectTree.cs
@@ -272,7 +272,7 @@ public sealed class DreamObjectTree {
                                 if (listValue.ValueKind != JsonValueKind.Object ||
                                     !listValue.TryGetProperty("key", out var jsonKey) ||
                                     !listValue.TryGetProperty("value", out var jsonValue)) {
-                                    throw new Exception("List value was missing a key or value property");
+                                    throw new Exception("AList value was missing a key or value property");
                                 }
 
                                 aList.SetValue(GetDreamValueFromJsonElement(jsonKey),


DMCompiler/DM/DMCodeTree.Vars.cs
```diff
@@ -173,7 +173,7 @@ private bool CheckCantDefine(DMCompiler compiler, DMObject dmObject) {
                     return true;
                 }
 
-                // Vars on /world and /list can only be defined in DMStandard
+                // Vars on /world, /list, and /alist can only be defined in DMStandard
                 if ((dmObject.Path == DreamPath.World || dmObject.Path == DreamPath.List) && !inStandard) {
```
wixoaGit:
```suggestion
                if ((dmObject.Path == DreamPath.World || dmObject.Path == DreamPath.List || dmObject.Path == DreamPath.AList) && !inStandard) {
```

OpenDreamRuntime/Objects/DreamObjectTree.cs
```diff
@@ -258,6 +263,25 @@ public DreamValue GetDreamValueFromJsonElement(object? value) {
...
+                    }
+                    case JsonVariableType.AList: {
+                        DreamAssocList aList = CreateAssocList();
+
+                        if (jsonElement.TryGetProperty("values", out JsonElement values)) {
+                            foreach (JsonElement listValue in values.EnumerateArray()) {
+                                if (listValue.ValueKind != JsonValueKind.Object ||
+                                    !listValue.TryGetProperty("key", out var jsonKey) ||
+                                    !listValue.TryGetProperty("value", out var jsonValue)) {
+                                    throw new Exception("List value was missing a key or value property");
```
wixoaGit:
```suggestion
                                    throw new Exception("AList value was missing a key or value property");
```

PR #2313, opened 2025-06-16
Typemaker 1.1 Rebased Redux

ike709:
Essentially just revives https://github.com/OpenDreamProject/OpenDream/pull/2208 with some key differences:

- No merge conflicts (yet)
- Does not do `as params` at all. This PR is large enough as-is and I don't want to figure out the passing around `procParameters` mess
- Tweaked the error message for typemaker
- Fixed the morbillion formatting warnings
- Fixed an NRE on compiling testgame
- Fixed some other stuff that I forgot about between committing and writing this desc
- Support `astype()` in typemaker
- Handle `ImplicitNullType` for proc args

Note that passing the `--skip-anything-typecheck` compiler option is still strongly recommended for trying to actually use typemaker.

Also don't forget to elevate the typemaker pragmas to warnings/errors for this to actually do anything.

Future PR things:
- `alist()` support
- More unit tests
- Revisit `as params` from the original PR


wixoaGit:
This uses quite a few hacky solutions (like `DMProcBuilder.CheckBlockReturnRecursive()`) that would likely be better done in other ways like with the creation of a control flow graph. I've tried adding one to this PR, but it's turning out to be a very in-depth change.

I don't think this is mergeable in its current state. It would be ideal if this could be split up into multiple PRs so the problematic changes can be worked on more effectively and the non-problematic ones can be merged in.

wixoaGit:
This PR appears to have been abandoned for the time being.

PR #2318, opened 2025-06-16
Fix bit-shift operators to match BYOND

amylizzle:
BYOND zeros out the top 8 bits after a bit shift because everything is a 32 bit float

This modifies our `<<`, `>>`, and `<<=`, `>>=` operators to do that too

### DIFF
--- a/Content.Tests/DMProject/Tests/Operators/valid_and_null_assign_const.dm
+++ b/Content.Tests/DMProject/Tests/Operators/valid_and_null_assign_const.dm
@@ -1,0 +1,128 @@
+//simple test of all basic assignment operators with valid and C(null) arguments
+//Const fold everying
+#define C(X) X
+/proc/RunTest()
+	var/a = C(1)
+	a += C(1)
+	ASSERT(a == C(2))
+	a = null
+	a += C(1)
+	ASSERT(a == C(1))
+	a = C(1)
+	a += C(null)
+	ASSERT(a == C(1))
+
+	a = C(1)
+	a -= C(1)
+	ASSERT(a == C(0))
+	a = null
+	a -= C(1)
+	ASSERT(a == C(-1))
+	a = C(1)
+	a -= C(null)
+	ASSERT(a == C(1))
+
+	a = C(2)
+	a *= C(3)
+	ASSERT(a == C(6))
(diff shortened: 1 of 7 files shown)


DMCompiler/SharedOperations.cs
```diff
@@ -61,4 +61,14 @@ public static float Log(float y, float baseValue) {
     public static float Abs(float a) {
         return MathF.Abs(a);
     }
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static int BitShiftLeft(int left, int right) {
+        return (left << right) & 0x00FFFFFF;
+    }
```
ZeWaka, line 70: Perhaps we could comment a potential future note in case we ever want to remove the behavior?

Content.Tests/DMProject/Tests/Operators/valid_and_null_assign_const.dm
```diff
@@ -0,0 +1,128 @@
+//simple test of all basic assignment operators with valid and C(null) arguments
+//Const fold everying
+#define C(X) X
```
wixoaGit, line 3: What's the point of this define? It's no different than just writing the value.

Content.Tests/DMProject/Tests/Operators/valid_and_null_const.dm
```diff
@@ -0,0 +1,103 @@
+//simple test of all basic operators with valid and C(null) arguments
+//Const fold everying
+#define C(X) X
```
wixoaGit, line 3: Same here

Content.Tests/DMProject/Tests/Operators/valid_and_null_assign_const.dm
amylizzle: It's just a copy of the other valid and null tests, so I changed the define instead of rewriting it for lazy reasons

Content.Tests/DMProject/Tests/Operators/valid_and_null_const.dm
amylizzle: ditto

wixoaGit (approved):

PR #2319, opened 2025-06-18
modified types????

### LINKED ISSUE #473: Modified types aren't properly implemented
See https://github.com/wixoaGit/OpenDream/issues/34

Currently the compiler just eats the modifications and throws a warning that it's unimplemented.

Laboredih123:
Fixes #473
Implements modified types by smuggling a json of the modified types all the way to createobject

*please* review my code thoroughly, i suck! there's probably a lot of things that can be improved!


Laboredih123:
so true github

DMCompiler/Compiler/DM/AST/DMAST.ExpressionConstant.cs
```diff
@@ -20,8 +20,9 @@ public sealed class DMASTConstantResource(Location location, string path) : DMAS
 
 public sealed class DMASTConstantNull(Location location) : DMASTExpressionConstant(location);
 
-public sealed class DMASTConstantPath(Location location, DMASTPath value) : DMASTExpressionConstant(location) {
+public sealed class DMASTConstantPath(Location location, DMASTPath value, Dictionary<string, DMASTExpression>? varOverrides) : DMASTExpressionConstant(location) {
```
wixoaGit: A modified type should probably be its own DMASTExpression type instead of an addition to DMASTConstantPath

DMCompiler/DM/Expressions/Builtins.cs
```diff
@@ -86,10 +89,19 @@ public override void EmitPushValue(ExpressionContext ctx) {
                 var newProc = ctx.ObjectTree.GetNewProc(typeReference.Value.Id);
 
                 (argumentsType, stackSize) = arguments.EmitArguments(ctx, newProc);
+
+                ctx.Proc.PushString(JsonSerializer.Serialize(variableOverrides));
```
wixoaGit: I would prefer for this to push null if there aren't any overrides.

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -195,12 +196,18 @@ public static ProcStatus CreateRangeEnumerator(DMProcState state) {
         public static ProcStatus CreateObject(DMProcState state) {
             var argumentInfo = state.ReadProcArguments();
             var val = state.Pop();
+            Dictionary<string, object>? overrides = null;
```
wixoaGit: This should be `Dictionary<string, object?>?`, that's what the compiler is encoding.

Laboredih123: oh i missed that

DMCompiler/DM/Builders/DMExpressionBuilder.cs
```diff
@@ -284,9 +284,56 @@ private DMExpression BuildExpression(DMASTExpression expression, DreamPath? infe
                     break;
                 }
 
-                result = new NewPath(Compiler, newPath.Location, path,
+                result = new NewPath(Compiler, newPath.Location, path, new Dictionary<string, object?>(),
```
amylizzle: it'd probably be better if the variable overrides arg was nullable and this pass a null instead of creating an empty dictionary every time a new object is created

amylizzle (changes requested):
It is generally upsetting to take expressions and smuggle them into the runtime in json format. I realise this was always going to be kinda hacky, but ew.

The modified types tests also contain atoms now, which they shouldn't anymore. Please change them to datums instead of objects.

This is behaviour that works in BYOND but not here: 
```
/obj
	var/a = 5

/proc/RunTest()
	var/obj/o = /obj{a=6}
	ASSERT((new o).a == 6)
```

Seems to work good on paradise and goon though!

amylizzle (approved):

DMCompiler/DM/Expressions/Builtins.cs
```diff
@@ -87,10 +90,23 @@ public override void EmitPushValue(ExpressionContext ctx) {
...
+                    ctx.Proc.PushString(JsonSerializer.Serialize(variableOverrides));
+                }
+
                 ctx.Proc.PushType(typeReference.Value.Id);
                 break;
             case ConstantProcReference procReference: // "new /proc/new_verb(Destination)" is a thing
                 (argumentsType, stackSize) = arguments.EmitArguments(ctx, ctx.ObjectTree.AllProcs[procReference.Value.Id]);
+                if(variableOverrides is not null && variableOverrides.Count > 0) {
+                    ctx.Compiler.Emit(WarningCode.BadExpression, Location, "Cannot add a Var Override to a proc");
+                    ctx.Proc.PushNull();
```
wixoaGit: Better to error if this code somehow makes it out of the compiler and the runtime stumbles upon it.
```suggestion
                    ctx.Proc.Error();
```

DMCompiler/DM/Builders/DMExpressionBuilder.cs
```diff
@@ -456,6 +503,7 @@ private DMExpression BuildConstant(DMASTExpressionConstant constant) {
             case DMASTConstantString constString: return new String(constant.Location, constString.Value);
             case DMASTConstantResource constResource: return new Resource(Compiler, constant.Location, constResource.Path);
             case DMASTConstantPath constPath: return BuildPath(constant.Location, constPath.Value.Path);
+            case DMASTModifiedType constModifiedPath: return BuildPath(constant.Location, constModifiedPath.Value.Path);
```
wixoaGit: `var/type = /datum{foo="bar"}` would just end up being equal to `var/type = /datum`? Should probably add an unimplemented warning for that.

DMCompiler/DM/Builders/DMExpressionBuilder.cs
```diff
@@ -305,7 +352,7 @@ private DMExpression BuildExpression(DMASTExpression expression, DreamPath? infe
                     break;
                 }
 
-                result = new NewPath(Compiler, newInferred.Location, inferredType,
+                result = new NewPath(Compiler, newInferred.Location, inferredType, new Dictionary<string, object?>(),
```
wixoaGit: Avoid the empty dictionary
```suggestion
                result = new NewPath(Compiler, newInferred.Location, inferredType, null,
```

DMCompiler/DM/Builders/DMExpressionBuilder.cs
```diff
@@ -284,9 +284,56 @@ private DMExpression BuildExpression(DMASTExpression expression, DreamPath? infe
...
+                if (BuildExpression(newModifiedType.Type, inferredPath) is not IConstantPath typePath) {
+                    result = BadExpression(WarningCode.BadExpression, newModifiedType.Type.Location,
+                        "Expected a path expression");
+                    break;
+                }
+
+                Dictionary<string, object?> overrides = new();
+                if (newModifiedType.Type.VarOverrides is null) {
+                    result = new NewPath(Compiler, newModifiedType.Location, typePath, overrides,
+                        BuildArgumentList(newModifiedType.Location, newModifiedType.Parameters, inferredPath));
```
wixoaGit: You could also avoid the empty dictionary here if you put the `if` earlier.

DMCompiler/DM/Builders/DMExpressionBuilder.cs
```diff
@@ -284,9 +284,56 @@ private DMExpression BuildExpression(DMASTExpression expression, DreamPath? infe
...
+                var failed = false;
+                foreach (KeyValuePair<string, DMASTExpression> varOverride in newModifiedType.Type.VarOverrides) {
+                    if (!owner.HasLocalVariable(varOverride.Key)) {
+                        return UnknownIdentifier(newModifiedType.Type.Location, varOverride.Key);
+                    }
+
+                    if (!BuildExpression(varOverride.Value, inferredPath)
+                            .TryAsConstant(Compiler, out var jsonConstant)) {
+                        if (!BuildExpression(varOverride.Value, inferredPath)
+                                .TryAsJsonRepresentation(Compiler, out var jsonValue)) {
```
wixoaGit: You're needlessly doing `BuildExpression()` twice here. The `TryAsConstant()` also isn't necessary here, you can just use `TryAsJsonRepresentation()`.

wixoaGit (changes requested):

wixoaGit (approved):

PR #2321, opened 2025-06-19
Implements bounds_dist

Laboredih123:
Yeah I did that.


ZeWaka (commented):
Tests?

wixoaGit:
> uh how do

`bounds_dist()` requires atoms so you'll have to write the test in the `Content.IntegrationTests` project. That's basically just a DM project that errors if something goes wrong. Make a new file in `DMProject/Tests/`, include that in the DME, and call it in `/world/New()` inside `code.dm`

Content.IntegrationTests/DMProject/Tests/bounds_dist.dm
```diff
@@ -0,0 +1,10 @@
+/proc/test_bounds_dist()
+	world.maxx = world.maxy = 5
+	world.maxz = 3
```
wixoaGit: We already have a 5x5x1 map.dmm. Add a second & third zlevel to that instead.

wixoaGit (changes requested):

Issue #2332, opened 2025-06-22
Decrementing a list's `len` below 0 throws


amylizzle:
This throws an error in OD. It doesn't in BYOND.
```
var/list/test = list()
test.len--
```

This should probably still be an error, but maybe a configurable one?

PR #2333, opened 2025-06-22
Cleans up the error message on setting a list to a negative size

amylizzle:
Makes the error tons easier to understand, but could probably still benefit from a configurable warning system

Related #2332


OpenDreamRuntime/Objects/Types/DreamList.cs
```diff
@@ -217,6 +217,10 @@ public void Resize(int size) {
                 AddValue(DreamValue.Null);
             }
         } else {
+            if (size < 0) {
+                //TODO emit configurable warning here
+                throw new InvalidOperationException("Setting a list size to a negative value is invalid");
+            }
             Cut(size + 1);
```
```suggestion
            }
            
            Cut(size + 1);
```

wixoaGit (approved):

PR #2349, opened 2025-07-22
Better initial() parity

Kapu1178:
Fixes initial(A.var) throwing an exception instead of returning null if A is a primitive. Retains exception behavior via optional runtime pragma.

### DIFF
--- a/OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
+++ b/OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
@@ -613,7 +613,6 @@ namespace OpenDreamRuntime.Procs {
             if (!key.TryGetValueAsString(out string? property)) {
                 throw new Exception("Invalid var for initial() call: " + key);
             }
-
             DreamObjectDefinition objectDefinition;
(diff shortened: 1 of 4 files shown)


OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -632,7 +631,9 @@ public static ProcStatus Initial(DMProcState state) {
             } else if (owner.TryGetValueAsType(out var ownerType)) {
                 objectDefinition = ownerType.ObjectDefinition;
             } else {
-                throw new Exception($"Invalid owner for initial() call {owner}");
+                state.DreamManager.OptionalException<ArgumentException>(DMCompiler.Compiler.WarningCode.InitialVarOnPrimitiveException, "Initial() attempted to get the initial value of a variable on a primative.");
```
wixoaGit:
```suggestion
                state.DreamManager.OptionalException<ArgumentException>(DMCompiler.Compiler.WarningCode.InitialVarOnPrimitiveException, "Initial() attempted to get the initial value of a variable on a primitive.");
```

wixoaGit (approved):

PR #2372, opened 2025-08-24
Remove all `/atom`s from tests

### LINKED ISSUE #1142: Unit tests should not reference `/atom` or its descendants

amylizzle:
Fixes #1142 

The vast majority were just find/replace, only a few needed any attention paying

Affected tests were found by forcing `DreamObjectAtom` to throw on creation.


Content.IntegrationTests/DMProject/Tests/output_string.dm
wixoaGit, line 1: I don't think this needs moved to IntegrationTests, this is just testing `null << "test"` doesn't error.

wixoaGit (changes requested):
The term `obj` is used in a bunch of filenames and var/proc names, those should probably be renamed to `datum`.

amylizzle:
> The term `obj` is used in a bunch of filenames and var/proc names, those should probably be renamed to `datum`.

eh, /datum is a `DreamObject`, the distinction is largely meaningless

wixoaGit (approved):

PR #2381, opened 2025-08-29
Rework IntegrationTests to be friendlier

amylizzle:
Just tweaks the `Content.IntegrationTests` so that it's easier to use, a bit more like standard `Content.Tests`

Now to add a new test, you just create a new file in the Tests folder that runs 
```
/world/New()
    ..()
```

It's a little clunky, but it works


Content.IntegrationTests/SetupCompileDM.cs
```diff
@@ -7,19 +7,29 @@
...
     public static readonly string BaseDirectory = AppDomain.CurrentDomain.BaseDirectory;
+    public static readonly string BaseDmEnvironment = Path.Join(BaseDirectory, TestProject, BaseEnvironment);
     public static readonly string DmEnvironment = Path.Join(BaseDirectory, TestProject, Environment);
+    public static readonly string TestFilesDirectory = Path.Join(BaseDirectory, TestProject, TestDirectory);
     public static readonly string CompiledProject = Path.ChangeExtension(DmEnvironment, "json");
 
     [OneTimeSetUp]
     public void Compile() {
         DMCompiler.DMCompiler compiler = new();
+        // open the DME and append include statements for each test
```
wixoaGit: You could include the files in the `Files` list a few lines down instead

wixoaGit (changes requested):

Content.IntegrationTests/SetupCompileDM.cs
amylizzle: I always forget you can do that

DMCompiler/DMCompiler.cs
```diff
@@ -21,7 +21,7 @@ namespace DMCompiler;
 public class DMCompiler {
     public readonly HashSet<WarningCode> UniqueEmissions = new();
     public DMCompilerSettings Settings;
-    public IReadOnlyList<string> ResourceDirectories => _resourceDirectories;
+    public IReadOnlyList<string> ResourceDirectories => _resourceDirectories.Distinct().ToList().AsReadOnly();
```
wixoaGit: Enumerating & creating a new list for every access on `ResourceDirectories` isn't ideal, and probably has an impact on projects with lots of resources. If this needs to prevent duplicate values then using a HashSet would be better.

wixoaGit (approved):

PR #2383, opened 2025-09-02
Add ignore rules to .codecov.yml

amylizzle:
Don't cover Tests, Packaging, or Server

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -9,3 +9,7 @@ comment: false
 codecov:
   max_report_age: off
   require_ci_to_pass: no
+ignore:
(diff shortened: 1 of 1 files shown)


.codecov.yml
```diff
@@ -9,3 +9,7 @@ comment: false
 codecov:
   max_report_age: off
   require_ci_to_pass: no
+ignore:
+  - "Content.Test"
```
wixoaGit:
```suggestion
  - "Content.Tests"
```

wixoaGit (changes requested):

PR #2384, opened 2025-09-03
Fixes blank line warnings and the easy obsolete warnings

Laboredih123:
we have a LOT of warnings on the opendream folder, bout time we shave em down a bit.


wixoaGit:
Did you ensure the changes from `GetValueAs()` -> `MustGetValueAs()` are correct? The reason the new method was created was so we could go through each use of the deprecated version, and make sure throwing an exception on an invalid value type was the correct behavior.

Laboredih123:
the ones i changed literally just call their mustgetvalueas versions and nothing else

ZeWaka (commented):


Issue #2385, opened 2025-09-03
Intermittent error in AtomGlideSystem.cs


amylizzle:
I think this one's a race condition.

<img width="1313" height="392" alt="Image" src="https://github.com/user-attachments/assets/6c988c91-def3-4a88-99b8-26bc43eef46e" />

<img width="1207" height="154" alt="Image" src="https://github.com/user-attachments/assets/c8718b8e-0324-424c-b0fc-9fdac4722639" />

<img width="836" height="78" alt="Image" src="https://github.com/user-attachments/assets/c85d5582-75b2-4805-8da2-ac071d3bb076" />

Essentially the entity that's gliding has no `MetaDataComponent` attached - don't know how or why, but that seems to be hardcoded to be expected and this bodge to kill the RT gliding is hitting on it

PR #2387, opened 2025-09-04
Fix Glide Crash

amylizzle:
Attempt at fixing #2385 

Hard to tell if it actually fixes anything because the error is super hard to trigger reliably, but I ran Paradise for a while with this fix and didn't get the crash, while running it without for a while would cause it, so maybe?


wixoaGit (changes requested):
This should remove the glide if the entity is deleted. Move the check to the `if` on line 55.

PR #2388, opened 2025-09-06
Misc builtin tests

ike709:
- Codecov number go up
- `locate(nonexistent_type)` no longer causes a stack size error
- Weighted pick with one arg is now a forced error in line with BYOND, but the error message is more helpful


DMCompiler/DM/Expressions/Builtins.cs
```diff
@@ -257,7 +257,7 @@ public override void EmitPushValue(ExpressionContext ctx) {
 
         if (weighted) {
             if (values.Length == 1) {
-                ctx.Compiler.ForcedWarning(Location, "Weighted pick() with one argument");
+                ctx.Compiler.Emit(WarningCode.BadArgument, Location, "Weighted pick() with one argument"); // BYOND errors with "extra args"
```
wixoaGit: Maybe WarningCode.InvalidArgumentCount would be better?

```suggestion
                ctx.Compiler.Emit(WarningCode.InvalidArgumentCount, Location, "Weighted pick() with one argument"); // BYOND errors with "extra args"
```

DMCompiler/DM/Expressions/Builtins.cs
```diff
@@ -109,7 +109,7 @@ internal sealed class LocateInferred(Location location, DreamPath path, DMExpres
     public override void EmitPushValue(ExpressionContext ctx) {
         if (!ctx.ObjectTree.TryGetTypeId(path, out var typeId)) {
             ctx.Compiler.Emit(WarningCode.ItemDoesntExist, Location, $"Type {path} does not exist");
-
+            ctx.Proc.PushNull(); // prevents a negative stack size error
```
wixoaGit, line 112: This should also write an `Error` opcode, in case the code ends up output somehow.
```suggestion
            ctx.Proc.PushNull(); // prevents a negative stack size error
            ctx.Proc.Error();
```

Content.Tests/DMProject/Tests/Expression/weighted_pick2.dm
```diff
@@ -0,0 +1,3 @@
+// COMPILE ERROR OD0100
```
wixoaGit:
```suggestion
// COMPILE ERROR OD0013
```

PR #2390, opened 2025-09-08
Nuke DMASTFolder

ike709:
Nukes `DMASTFolder`. It's deprecated and removing it shaves a consistent 6 seconds off my third-run TG compile time (45s -> 39s).

This uncovered an OD bug in the following preprocessed TG code:
```dm
/proc/collect_vv(obj/item/item)
    var/static/list/ignored_vars = list(
        (@#animate_movement# ||item.animate_movement),
        (@#datum_flags# ||item.datum_flags),
        (@#fingerprintslast# ||item.fingerprintslast),
        (@#layer# ||item.layer),
        (@#plane# ||item.plane),
        (@#screen_loc# ||item.screen_loc),
        (@#tip_timer# ||item.tip_timer),
        (@#vars# ||item.vars),
        (@#x# ||item.x),
        (@#y# ||item.y),
        (@#z# ||item.z),
    )
```
`DMASTFolder` was returning just the LHS as a constant string, which hid the fact that OD couldn't resolve the deref in the RHS. I _think_ I got it working, but `BuildIdentifier()` is such a mess I'm not 100% sure this is the proper fix.


DMCompiler/DM/Builders/DMExpressionBuilder.cs
```diff
@@ -550,7 +550,7 @@ private DMExpression BuildPath(Location location, DreamPath path) {
 
     private DMExpression BuildIdentifier(DMASTIdentifier identifier, DreamPath? inferredPath = null) {
         var name = identifier.Identifier;
-        if (scopeMode == Normal) {
+        if (scopeMode == Normal || scopeMode == Static) {
```
wixoaGit: This is not valid even in BYOND, local vars/args aren't available to static vars. Making this change will cause the global init proc to attempt to use one of its own locals, which is incorrect.

I think what BYOND is doing here is a first check that the var exists when the expression is created (that is ignoring the static context), and a second check when the bytecode is generated that is being skipped because of constant folding. As far as I can tell, the first step ignoring the static context is a bug. I'm not sure of the best way to replicate this with the way OD is structured.

wixoaGit (changes requested):

DMCompiler/DM/Builders/DMExpressionBuilder.cs
ike709: I do not know how to fix this.

wixoaGit (approved):

Issue #1027, opened 2023-01-13
Infinite loop in compiler with recursive `#define`


amylizzle:
```
#define foo bar foo
foo
```

causes compiler to hang as it infinitely loops.

Probably maximum recursion depth needed, or maybe just disallow self-references in defines?

PR #2395, opened 2025-09-14
Add compiler messages to integration test

amylizzle:
No less hacky than `UniqueEmissions`, but effective

### DIFF
--- a/DMCompiler/DMCompiler.cs
+++ b/DMCompiler/DMCompiler.cs
@@ -214,10 +214,10 @@ public class DMCompiler {
 
         if (Settings.StoreMessages)
             if (CompilerMessages.Count < 1000)
-                CompilerMessages.Add(new CompilerEmission(ErrorLevel.Error, loc, message).ToString());
+                CompilerMessages.Add(emission.ToString());
             else {
                 Settings.StoreMessages = false;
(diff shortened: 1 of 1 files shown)


wixoaGit:
We used to do this but removed it because of the huge RAM usage in some cases. There should be a limit to how many messages can be stored.

wixoaGit:
If you somehow trigger an infinite loop of errors in the compiler, this could end up consuming all your RAM. Potentially triggerable by #1027? I'm not sure, but it's happened in the past.

amylizzle:
we should probably abort on that anyway, but I take your point. I'll put a cap on it

DMCompiler/DMCompiler.cs
```diff
@@ -211,6 +212,13 @@ public bool Emit(WarningCode code, Location loc, string message) {
                 break;
         }
 
+        if (Settings.StoreMessages)
+            if (CompilerMessages.Count < 1000)
+                CompilerMessages.Add(new CompilerEmission(ErrorLevel.Error, loc, message).ToString());
```
wixoaGit: Is there a reason this is creating a new CompilerEmission that's always an error instead of using the already created `emission`?

amylizzle: there is not, I have no idea why I did that

wixoaGit (approved):

PR #2406, opened 2025-10-04
Stop infinite looping when errors occur in `/world/Error()`

### LINKED ISSUE #1640: Throwing an error in `/world/Error()` results in infinite loop
```
/world/Error(exception/E)
  var/list/not_init
  not_init["null deref"] = "oops"

/proc/main()
  var/list/also_not_init
  world.log << also_not_init["ohno"]
```

In BYOND: Error in `/world/Error` is printed normally
In OD: Infinite loop

amylizzle:
Fixes #1640

### DIFF
--- a/OpenDreamRuntime/DreamThread.cs
+++ b/OpenDreamRuntime/DreamThread.cs
@@ -479,7 +479,7 @@ namespace OpenDreamRuntime {
                 line = source.Item2;
             }
 
-            dreamMan.HandleException(exception, msg, file, line);
+            dreamMan.HandleException(exception, msg, file, line, inWorldError: _current?.Proc?.OwningType == dreamMan.WorldInstance.ObjectDefinition.TreeEntry && _current.Proc.Name == "Error");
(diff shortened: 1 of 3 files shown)


OpenDreamRuntime/DreamThread.cs
```diff
@@ -479,7 +479,7 @@ private void HandleException(Exception exception) {
                 line = source.Item2;
             }
 
-            dreamMan.HandleException(exception, msg, file, line);
+            dreamMan.HandleException(exception, msg, file, line, inWorldError: _current?.Proc?.OwningType == dreamMan.WorldInstance.ObjectDefinition.TreeEntry && _current.Proc.Name == "Error");
```
wixoaGit: If `/world/Error()` calls another proc and _that_ one errors, it also needs to not call `/world/Error()` again. Maybe check every proc state in `_stack`?

wixoaGit (changes requested):

Issue #2396, opened 2025-09-14
RT's `CreateRenderTarget()` fails causing client crash on goonstation


amylizzle:
When connecting to a local instance of goon using 
```
#define GOTTA_GO_FAST_BUT_ZLEVELS_TOO_SLOW 1  
#define IM_REALLY_IN_A_FUCKING_HURRY_HERE 1  
```
in `__build.dm`, a client crash occurs when the player join window, changelog window, and character creation windows should be shown.

The crash seems to be related to the creation of a tooltip `browse()` window.

Two errors occur. First when the window is created, a non-fatal error occurs leaving the `_renderTargets` dictionary in a bad state
```
Robust.Shared.Utility.DebugAssertException: new framebuffer has bad status 0
   at Robust.Shared.Utility.DebugTools.Assert(Boolean condition, AssertInterpolatedStringHandler& message) in /home/wnv9567/workspace/OpenDream/RobustToolbox/Robust.Shared/Utility/DebugTools.cs:line 215
   at Robust.Client.Graphics.Clyde.Clyde.CreateRenderTarget(Vector2i size, RenderTargetFormatParameters format, Nullable`1 sampleParameters, String name) in /home/wnv9567/workspace/OpenDream/RobustToolbox/Robust.Client/Graphics/Clyde/Clyde.RenderTargets.cs:line 195
   at Robust.Client.Graphics.Clyde.Clyde.GLContextWindow.CreateWindowRenderTexture(WindowData reg) in /home/wnv9567/workspace/OpenDream/RobustToolbox/Robust.Client/Graphics/Clyde/GLContext/GLContextWindow.cs:line 329
   at Robust.Client.Graphics.Clyde.Clyde.GLContextWindow.WindowResized(WindowReg reg, Vector2i oldSize) in /home/wnv9567/workspace/OpenDream/RobustToolbox/Robust.Client/Graphics/Clyde/GLContext/GLContextWindow.cs:line 130
   at Robust.Client.Graphics.Clyde.Clyde.SendWindowResized(WindowReg reg, Vector2i oldSize) in /home/wnv9567/workspace/OpenDream/RobustToolbox/Robust.Client/Graphics/Clyde/Clyde.Events.cs:line 118
[...]

PR #2409, opened 2025-10-10
DMF Button Images

### LINKED ISSUE #1660: Button Control needs to support images
```
elem "observe"
		type = BUTTON
		pos = 18,136
		size = 200x40
		anchor1 = none
		anchor2 = none
		font-family = "Verdana"
		font-size = 10
		font-style = "bold"
		saved-params = "is-checked"
		text = ""
		image = 'icons\\ui\\observe.png'
		command = ".observe_round"
```
currently renders a blank button, should instead render an image on the button.

amylizzle:
Fixes #1660 

image

I'd take a picture from goon but I can't actually launch it due to #2396 :sob: 

This implements clients being able to request resources by path, provided they have been loaded as a resource by the server already. It does not allow the client to request arbitrary files, but it does allow the client to request _any_ loaded resource by file path. This isn't particularly distinct from being able to request any loaded resource by id.

The client essentially does a path -> id lookup, then gets the resource by id in the usually manner, and displays it as an image on the button control.

Also the server does some DMF parsing now to extract the images. DMF parsing has been moved to the OpenDreamShared namespace.

TODOs (probably in another PR):
- it should be possible to winset the button's image to a `\ref[icon]`  
- path normalisation would probably be a good idea for resources in general
- some other controls use images, adding support for those should be relatively trivial

### DIFF
--- a/OpenDreamClient/Resources/DreamResourceManager.cs
+++ b/OpenDreamClient/Resources/DreamResourceManager.cs
@@ -295,10 +295,10 @@ internal sealed class DreamResourceManager : IDreamResourceManager {
 
             var timeout = _cfg.GetCVar(OpenDreamCVars.DownloadTimeout);
             Robust.Shared.Timing.Timer.Spawn(TimeSpan.FromSeconds(timeout), () => {
-                if (_pendingResourceLookups.ContainsKey(resourcePathOrRef)) {
+                if (_pendingResourceLookups.TryGetValue(resourcePathOrRef, out var pendingLookup)) {
                     _sawmill.Warning(
                         $"Resource id {resourcePathOrRef} lookup was requested, but is still not received {timeout} seconds later.");
-                    foreach (var failureCallback in _pendingResourceLookups[resourcePathOrRef].FailureCallbacks)
+                    foreach (var failureCallback in pendingLookup.FailureCallbacks)
                         failureCallback.Invoke();
                 }
             });


OpenDreamClient/Resources/DreamResourceManager.cs
```diff
@@ -144,7 +150,21 @@ private void RxResourceUpdateNotification(MsgNotifyResourceUpdate message) {
         }
     }
 
-    public void LoadResourceAsync<T>(int resourceId, Action<T> onLoadCallback) where T:DreamResource {
+    private void RxLookupResourceResponse(MsgLookupResourceResponse message) {
+        if (_pendingResourceLookups.TryGetValue(message.ResourcePathOrRef, out var pendingResourceLookup)) {
```
wixoaGit: These should be removed from the dictionary once they're no longer necessary.
```suggestion
        if (_pendingResourceLookups.TryGetValue(message.ResourcePathOrRef, out var pendingResourceLookup)) {
            _pendingResourceLookups.Remove(message.ResourcePathOrRef);
```

OpenDreamClient/Resources/DreamResourceManager.cs
```diff
@@ -252,8 +272,50 @@ public bool EnsureCacheFile(string filename, int timeoutSeconds = 5) {
...
+            _netManager.ClientSendMessage(msg);
+
+            var timeout = _cfg.GetCVar(OpenDreamCVars.DownloadTimeout);
+            Robust.Shared.Timing.Timer.Spawn(TimeSpan.FromSeconds(timeout), () => {
+                if (_pendingResourceLookups.ContainsKey(resourcePathOrRef)) {
+                    _sawmill.Warning(
+                        $"Resource id {resourcePathOrRef} lookup was requested, but is still not received {timeout} seconds later.");
+                    foreach (var failureCallback in _pendingResourceLookups[resourcePathOrRef].FailureCallbacks)
+                        failureCallback.Invoke();
+                    _pendingResourceLookups.Remove(resourcePathOrRef);
```
wixoaGit: I don't think considering the request failed is a good idea here, it could just be a delay in the network.

OpenDreamRuntime/Resources/DreamResourceManager.cs
```diff
@@ -18,14 +19,15 @@ public sealed class DreamResourceManager {
     [Dependency] private readonly IServerNetManager _netManager = default!;
     [Dependency] private readonly IStatusHost _statusHost = default!;
     [Dependency] private readonly IDependencyCollection _dependencyCollection = default!;
-
+    [Dependency] private readonly ISerializationManager _serializationManager = default!;
```
wixoaGit, line 22:
```suggestion
    [Dependency] private readonly ISerializationManager _serializationManager = default!;
    
```

OpenDreamRuntime/Resources/DreamResourceManager.cs
```diff
@@ -102,6 +107,7 @@ DreamResource GetResource() {
             return resource;
         }
 
+
```
wixoaGit:
```suggestion
```

OpenDreamRuntime/Resources/DreamResourceManager.cs
```diff
@@ -111,9 +117,20 @@ DreamResource GetResource() {
             _resourcePathToId.Add(resourcePath, resourceId);
         }
 
+        ProcessQueuedResourceLoads();
```
wixoaGit, line 120: This ends up being a recursive call, instead of the loop you intend it to be.

Maybe instead make `ParseAndLoadResources()` public and call it here? You could add `[Access(typeof(DreamResourceManager), Friend = AccessPermissions.Execute)]` to it.

OpenDreamRuntime/Resources/DreamResourceManager.cs
```diff
@@ -202,8 +225,11 @@ public IconResource CreateIconResource(byte[] data) {
 
     public void RxRequestResource(MsgRequestResource pRequestResource) {
         if (TryLoadResource(pRequestResource.ResourceId, out var resource)) {
+            if(resource.ResourceData is null) {
+                throw new Exception($"Attempted to send a bad resource {resource.ResourcePath} with ID {resource.Id}");
+            }
             var msg = new MsgResource {
-                ResourceId = resource.Id, ResourceData = resource.ResourceData
+                ResourceId = resource.Id, ResourceData = resource.ResourceData!
```
wixoaGit: This isn't necessary because of the null check above.
```suggestion
                ResourceId = resource.Id, ResourceData = resource.ResourceData
```

OpenDreamRuntime/Resources/DreamResourceManager.cs
amylizzle: Recursive was the intent - basically loading a resource (like a DMF) can trigger other resources to be created, but if you create them inside the `DreamResource` constructor it throws off all the ID numbers so you end up pointing to the wrong thing. Instead, `DreamResource` constructors add the resource to be created to a queue, and then process that at the end of registering the resource. Recursive in the event that somehow a resource loads a resource which loads a resource...

Issue #2412, opened 2025-10-13
Parity break in `LValue.EmitPushInitial()`


ike709:
Right now all `LValue`s that do not override `EmitPushInitial()` currently compiletime error on `initial()`. However, BYOND in at least some of these cases (such as `args`) returns the current value instead of an error.

I didn't go through and test all of the lvalues but we should likely just convert the method from `virtual` to `abstract` and audit the behavior of every `initial(LValue)` in BYOND.

PR #2413, opened 2025-10-13
Tests for `LValue.cs`

ike709:
- Codecov for most of `LValue.cs`
- Fixed a negative stack size error
- Uncovered a parity issue: https://github.com/OpenDreamProject/OpenDream/issues/2412
- Learned that we still don't actually set BYOND's default values for `global.vars`. I'll fix that in a follow-up PR.

### DIFF
--- a/Content.Tests/DMProject/Tests/Tree/Global/Var/global_reference.dm
+++ b/Content.Tests/DMProject/Tests/Tree/Global/Var/global_reference.dm
@@ -2,5 +2,5 @@
 
 /proc/RunTest()
 	var/list/L = list()
-	for(global in L)
+	for(global in L) // PARITY BREAK: In BYOND this is actually a `Cannot modify null.Northeast` runtime. Compiletime error is probably fine for OD.
 		return


Content.Tests/DMProject/Tests/Tree/Global/Var/global_reference.dm
```diff
@@ -0,0 +1,6 @@
+// COMPILE ERROR OD0011
+
+/proc/RunTest()
+	var/list/L = list()
+	for(global in L)
```
ike709: BYOND actually attempts to run this until it hits a `Cannot modify null.Northeast` runtime. I think keeping this a compiletime error is probably fine.

ike709:
```suggestion
	for(global in L) // PARITY BREAK: In BYOND this is actually a `Cannot modify null.Northeast` runtime. Compiletime error is probably fine for OD.
```

Issue #2367, opened 2025-08-18
`list[key] = value` is much slower than in BYOND


wixoaGit:
```dm
var/list/L1 = list()
for (var/i in 1 to 100000)
    L1 += new /datum

var/start = world.timeofday
var/list/L2 = list()
for (var/i in 1 to 100000)
    var/key = L1[i]
    L2[key] = 1
world.log << world.timeofday - start
```
The second loop took 4.1 seconds in BYOND while OpenDream took 21.7 seconds (this ran on different hardware so not _directly_ comparable, but still a large difference).

This is causing a large increase in init time on pretty much all major codebases right now. The issue is that the entire list needs searched for `key` to know whether or not the value needs added to the list. ~~BYOND probably checks its red-black tree for the existence of the key. We don't keep any data structure like that currently, so we search the whole list.~~ BYOND might just be a lot faster at comparing values, see my comment below.

wixoaGit:
I experimented with removing DreamValue's `_refValue` and turning it into this:
```cs
[FieldOffset(0)] public readonly DreamValueType Type;

[FieldOffset(4)] private readonly int _intValue;
[FieldOffset(4)] private readonly float _floatValue;
```
Where `_intValue` is used as an ID for the DreamObject, allowing me to change the `Equals()` method:
```cs
public bool Equals(DreamValue other) {
    return Type == other.Type && _intValue == other._intValue;
}
```
This brought execution of the DM example from 24.3 seconds down to 3.6 seconds. Probably worth looking into as an overall optimization to both execution speed and memory.

Biggest issue would likely be it breaking an assumption required for our GC to work, since DreamValue no longer holds a reference to DreamObject.

Ruzihm:
I was curious about how much of the speed difference is attributable to datum equality checking vs list operation implementation and found that the following benchmark runs a little bit faster in OD (32-40) than Byond (56-58) :

```
	verb/test_equality_speed()
		var/list/L1 = list()
		for (var/i in 1 to 9000)
			L1 += new /datum
		
		var/start = world.timeofday
		var/x = 1
		for (var/i in 1 to 9000)
			for (var/o in i to 9000)
				var/test = L1[i]
				var/test2 = L1[o]
				if (test == test2)
					x = 1 - x
		world.log << x
		world.log << world.timeofday - start
```

To me, that suggests that the main difference in speed in the PR benchmark lies in the list operation, not in equality checking. Or maybe OD is faster here for some other reason.

PR #2418, opened 2025-10-14
Speeds up lists massively

redmoogle:
Partial fix for: #2367 

This is alot faster; went from 2.1 minutes with the code in the issue to ~490ms

I'm *fairly* certain this won't break anything because its just checking if the key exists and instead of manually searching the dictionary I just uses the dictionary search function thats faster (O(n) instead of O(n^2))


redmoogle:
> ```
> /proc/RunTest()
> 	var/list/L = list()
> 	for(var/i in 1 to 5)
> 		L.Add("[i]")
> 		L["[i]"] = "item [i]"
> 	ASSERT(length(L) == 5)
> 	ASSERT(L["3"] == "item 3")
> ```
> 
> should give `{"1":"item 1","2":"item 2","3":"item 3","4":"item 4","5":"item 5"}` instead gives `{"1":"item 1","1":"item 1","2":"item 2","2":"item 2","3":"item 3","3":"item 3","4":"item 4","4":"item 4","5":"item 5","5":"item 5"}`

I just found this out; I got another idea that should be pretty similar though when I get done coding it

redmoogle:
Im going to close this for now until I figure this out

redmoogle:
Should be good to rereview now

OpenDreamRuntime/Objects/Types/DreamList.cs
```diff
@@ -173,13 +188,33 @@ public virtual DreamValue GetValue(DreamValue key) {
...
                 _values[keyInteger - 1] = value;
+
+                if (!_reverseLookup.TryAdd(value, 1)) {
+                    _reverseLookup[value] += 1;
+                }
+
             }
         } else {
-            if (!ContainsValue(key)) _values.Add(key);
+            if (!_reverseLookup.TryAdd(key, 1)) {
```
Ruzihm, line 213: This should be: 

```
if (_reverseLookup.TryAdd(key, 1)) {
    _values.Add(key);
}
```

redmoogle: This is covered by the containsvalue function. also this is closed

PR #2432, opened 2025-10-26
Bring parity to num/`null` 

### LINKED ISSUE #2425: Math Parity: 10/null should equal 10
From running BYOND against unit tests. `Division.dm` test should be modified and OD behaviour should be modified

Note this is only for constant values - for variable values 10/null should be a runtime error. Because BYOND has never been anything approaching consistent.

amylizzle:
variable 10/null = runtime error
constant 10/null = compile error

fixes #2425

### DIFF
--- a/Content.Tests/DMProject/Tests/Operators/div_by_null.dm
+++ b/Content.Tests/DMProject/Tests/Operators/div_by_null.dm
@@ -1,3 +1,3 @@
-// COMPILE ERROR OD0502
+// COMPILE ERROR OD0011
 /proc/RunTest
 	var/test = 10/null


wixoaGit:
Lummox made this a compile error in the next release

DMCompiler/Compiler/CompilerError.cs
```diff
@@ -126,6 +127,7 @@ public struct CompilerEmission {
         {WarningCode.IAmATeaPot, ErrorLevel.Error},
         {WarningCode.HardConstContext, ErrorLevel.Error},
         {WarningCode.WriteToConstant, ErrorLevel.Error},
+        {WarningCode.DivideByNull, ErrorLevel.Error},
```
ike709: ...Do we need a new warning code specifically for divide by null? Isn't stuff like this normally just BadExpression?

wixoaGit: Yea, BadExpression would be more consistent here.

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -1143,8 +1143,7 @@ public static ProcStatus Divide(DMProcState state) {
                     state.Push(new DreamValue(0));
                     break;
                 case DreamValue.DreamValueType.Float when second.IsNull:
-                    state.Push(new DreamValue(first.MustGetValueAsFloat()));
-                    break;
+                    throw new Exception("Division by null");
```
wixoaGit: Can't hurt to have information about what we're trying to divide
```suggestion
                    throw new Exception($"Attempted to divide {first} by null");
```

DMCompiler/Compiler/CompilerError.cs
amylizzle: Will do

wixoaGit (approved):

Content.Tests/DMProject/Tests/Operators/div_by_null.dm
```diff
@@ -0,0 +1,3 @@
+// COMPILE ERROR OD0502
```
wixoaGit:
```suggestion
// COMPILE ERROR OD0011
```

PR #2454, opened 2025-11-19
Fix `for` body parsing

### LINKED ISSUE #2117: empty `for` body causes parsing errors
```
if (!contents_changed)
	return src.cached_producibility_data
for (var/obj/item/material_piece/M as anything in C)
// the body was not indented properly
var/list/output = list()
for (var/datum/manufacture/M as anything in ALL_BLUEPRINTS)
	var/M_ref = "\ref[M]"
	output[M_ref] = 1
```

this code threw unknown identifier errors on `output`. I'm pretty sure the empty for loop is eating the var def.

ideal output should probably be a warning on the empty for loop

wixoaGit:
A `for` will no longer treat an unindented line of code as its body.

Fixes #2117

### DIFF
--- a/.github/workflows/compiler-test.yml
+++ b/.github/workflows/compiler-test.yml
@@ -139,28 +139,28 @@ jobs:
     - name: Compile Paradise Master
       run: main/bin/DMCompiler/DMCompiler para/paradise.dme --suppress-unimplemented --suppress-unsupported
 
-#  compile-nebula:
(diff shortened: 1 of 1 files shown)


Drulikar:
https://github.com/cmss13-devs/cmss13/pull/11179 is merged.

MistakeNot4892:
Thanks for the report, our fix is here https://github.com/NebulaSS13/Nebula/pull/5202

Issue #2178, opened 2025-01-21
Relative-pathed var parsing issues


ike709:
The following code currently can't be parsed by OpenDream:
```dm
/datum/foo
	var
		tmp
			active = 0
			deleting = 0
			full_delete = 0
			delay = 5
		list
			connected = list()
```

PR #2463, opened 2025-11-25
Fix nested object var blocks

wixoaGit:
Moves object blocks to be prioritized over object vars, which gets the recursive parsing working.

Alternative fix to #2178.


Zonespace27:
Confirmed to work on eternia.

PR #2464, opened 2025-11-25
Make `macro` optional

### LINKED ISSUE #1996: Not setting `macro` causes an exception
This is from Eternia:
```
window "main"
    elem "main"
        type = MAIN
        pos = 281,0
        size = 1487x714
        anchor1 = -1,-1
        anchor2 = -1,-1
        is-default = true
        saved-params = "pos;size"
        on-size = "WindowResize"
        statusbar = false
        icon = 'icons\\interface\\titlescreen\\images\\meranthe_logo.png'
```
`macro` defaulting to an empty string causes a `KeyNotFound` exception in `ControlWindow.cs`: `public InterfaceMacroSet Macro => _interfaceManager.MacroSets[WindowDescriptor.Macro.AsRaw()];`

According to the ref, it seems this field is optional so we just need to support it being null or empty.

https://www.byond.com/docs/ref/#/{skin}/param/macro

ike709:
Resolves #1996 

I tested that this removes the exception by nuking the `macro` from testgame.

I did not test whether this fix requires move involved changes to the active macro set.

### DIFF
--- a/OpenDreamClient/Interface/Controls/ControlWindow.cs
+++ b/OpenDreamClient/Interface/Controls/ControlWindow.cs
@@ -20,7 +20,7 @@ public sealed class ControlWindow : InterfaceControl {
     public readonly List<InterfaceControl> ChildControls = new();
 
     public string Title => WindowDescriptor.Title.Value;
-    public InterfaceMacroSet Macro => _interfaceManager.MacroSets[WindowDescriptor.Macro.AsRaw()];
+    public InterfaceMacroSet? Macro => _interfaceManager.MacroSets.GetValueOrDefault(WindowDescriptor.Macro.AsRaw());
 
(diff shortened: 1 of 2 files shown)


OpenDreamClient/Interface/Controls/ControlWindow.cs
```diff
@@ -54,7 +54,7 @@ protected override void UpdateElementDescriptor() {
             UpdateWindowAttributes(_myWindow);
 
         if (WindowDescriptor.IsDefault.Value) {
-            Macro.SetActive();
+            Macro?.SetActive();
```
wixoaGit, line 57: This would leave the previous macro set as the active one if `macro` is removed through winset.

I would also like to test if BYOND has a default macro set for this case.

wixoaGit (changes requested):

PR #2470, opened 2025-11-26
Free speedups

redmoogle:
The changes in `DreamValues.cs` provide about 20-30% speedup on equals comparison (900-1000 ticks vs 1300)
The changes in `DreamList.cs` maybe provide 1-5% speedup (get_count was using about 1.83% cpu)

tested using
```
var/list/L1 = list()
for (var/i in 1 to 100000)
    L1 += new /datum

var/start = world.timeofday
var/list/L2 = list()
for (var/i in 1 to 100000)
    var/key = L1[i]
    L2[key] = 1
world.log << world.timeofday - start
```


OpenDreamRuntime/DreamValue.cs
```diff
@@ -455,11 +455,13 @@ public bool Equals(DreamValue other) {
...
-                if (_refValue != null && Unsafe.As<DreamObject>(_refValue).Deleted)
+                var thisObj = Unsafe.As<DreamObject>(_refValue);
+                var otherObj = Unsafe.As<DreamObject>(other._refValue);
+                if (thisObj != null && thisObj.Deleted == true)
                     _refValue = null;
-                if (other._refValue != null && Unsafe.As<DreamObject>(other._refValue).Deleted)
+                if (otherObj != null && otherObj.Deleted == true)
                     other._refValue = null;
-                break;
+                return thisObj == otherObj;
```
wixoaGit, line 470: Does this not break comparisons with deleted objects? It compares `_refValue` before the `Deleted` check.

wixoaGit (changes requested):

wixoaGit (approved):

PR #2473, opened 2025-12-14
Give animate its own opcode, fixes animate w/ null transform

### LINKED ISSUE #1790: `animate(transform=null)` doesn't work right
explicitly setting `transform=null` should set the transform to the identity matrix. Instead `animate()` currently treats it like transform is not passed as a param.

Ruzihm:
Refactors animate to be a separate opcode, and handles its arguments more directly to distinguish between provided null transform vs a missing one.

Changes `animate` to behave differently when an argument is null because it is omitted vs. a null included in the caller's arguments. Closes #1790 

Adds a unit test to test for the aforementioned animate/transform/null behavior.

### DIFF
--- a/DMCompiler/DMStandard/UnsortedAdditions.dm
+++ b/DMCompiler/DMStandard/UnsortedAdditions.dm
@@ -6,29 +6,15 @@
 /proc/eval(script)
 	set opendream_unsupported = "eval() is officialy deprecated"
 
-/proc/findlasttext_char(Haystack,Needle,Start=0,End=1)
-	set opendream_unimplemented = TRUE
-/proc/findlasttextEx_char(Haystack,Needle,Start=1,End=0)
-	set opendream_unimplemented = TRUE
-/proc/findtext_char(Haystack,Needle,Start=1,End=0)
-	set opendream_unimplemented = TRUE
-/proc/findtextEx_char(Haystack,Needle,Start=1,End=0)
-	set opendream_unimplemented = TRUE
 /proc/load_resource(File)
 	set opendream_unimplemented = TRUE
 proc/missile(Type, Start, End)
 	set opendream_unimplemented = TRUE
 /proc/obounds(Ref=src, Dist=0)
 	set opendream_unimplemented = TRUE
-/proc/replacetext_char(Haystack,Needle,Replacement,Start=1,End=0)
-	set opendream_unimplemented = TRUE
 /proc/run(File)
 	set opendream_unimplemented = TRUE
 /proc/shell(command)
-	set opendream_unimplemented = TRUE
-/proc/nonspantext_char(Haystack,Needles,Start=1)
-	set opendream_unimplemented = TRUE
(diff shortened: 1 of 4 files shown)


OpenDreamRuntime/DreamValue.cs
```diff
@@ -70,6 +76,7 @@ public static DreamValue EmptyString {
 
     private object? _refValue;
     private readonly float _floatValue;
+    private bool _defaultNull = false;
```
wixoaGit: Adding new fields to `DreamValue` is kind of a no-go since any increase to its size or complexity has heavy impacts on OpenDream's performance as a whole. This would greatly increase RAM usage as well as the time taken to copy / create these values, since it's done so often.

Ruzihm: I was afraid of that. Would it be reasonable to, say, add a new field to `DreamProcArguments` which indicates which argument indices (if any) correspond to provided nulls vs default nulls, and pass in that info via its constructors?

Currently the distinction would only be made in one `DreamProcArguments` constructor call, so maybe just make another constructor with the `providedNulls` parameter and just have it called there and leave it at that...

OpenDreamRuntime/Procs/ProcDecoder.cs
```diff
@@ -142,6 +142,8 @@ public ITuple DecodeInstruction() {
             case DreamProcOpcode.Gradient:
             case DreamProcOpcode.Rgb:
                 return (opcode, (DMCallArgumentsType)ReadByte(), ReadInt());
+            case DreamProcOpcode.Animate:
```
wixoaGit, line 144: The bodies are the same so you can put the two cases together instead.
```suggestion
            case DreamProcOpcode.Rgb:
            case DreamProcOpcode.Animate:
```

DMCompiler/DM/DMCodeTree.Vars.cs
```diff
@@ -86,6 +86,7 @@ private bool IsValidRightHandSide(DMCompiler compiler, DMObject dmObject, DMExpr
                 NewList => true,
                 NewPath => true,
                 Rgb => true,
+                Animate => true,
```
wixoaGit: BYOND errors on this:
```dm
/datum/var/foo = animate(new /obj)
```
so this isn't valid on the right-hand side.
```suggestion
```

DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -2757,6 +2757,9 @@ private void BracketWhitespace() {
 
                         return new DMASTRgb(identifier.Location, callParameters);
                     }
+                    case "animate": {
```
wixoaGit, line 2760: This should verify that there is at least 1 argument.
```suggestion
                    case "animate": {
                        if (callParameters.Length is < 1)
                            Emit(WarningCode.InvalidArgumentCount, callLoc,
                                "Expected at least 1 argument for animate()");
                        
```

PR #2475, opened 2025-12-14
Implement several  *_char function wrappers

Sorrowfulwinds:
Adds support for 8 *_char() procs from [#2194](https://github.com/OpenDreamProject/OpenDream/issues/2194#issue-2824014237)

It looks like the OpenDream project has settled on text manipulation procs indexing by character and not the byond-default byte indexing. This means that they all function identically to the *_char versions in byond. So these added functions are just wrappers for their non_char versions.

Adds the following proc support:
findtext_char
findtextEx_char
findlasttext_char
findlasttextEx_char
replacetext_char
replacetextEx_char
nonspantext_char
splittext_char

I did a simple test to confirm these share the same results in both this PR of OD and 516.1673 of byond. There is no actual new logic added with this PR so the test is purely for functionality confirmation.

### DIFF
--- a/DMCompiler/DMStandard/UnsortedAdditions.dm
+++ b/DMCompiler/DMStandard/UnsortedAdditions.dm
@@ -6,29 +6,15 @@
 /proc/eval(script)
 	set opendream_unsupported = "eval() is officialy deprecated"
 
-/proc/findlasttext_char(Haystack,Needle,Start=0,End=1)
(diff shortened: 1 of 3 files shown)


Sorrowfulwinds:
I think I have added the new functions to all the necessary lists. I removed the compiler warnings for these functions and they do compile and return the correct values on a simple test.

PR #2477, opened 2025-12-15
Client distinguish between eye and mob for drawing, context menu, and verbs

### LINKED ISSUE #1581: Hopping into a locker lets you see ghosts
15:57]wixoa: Hopping in a locker causes see_invisible to be ignored
[15:58]wixoa: Paradise changes your eye to be the locker you're in, and we get the see_invisible value from your eye instead of the mob like we should
[Screencast from 31-12-23 15:52:35.webm](https://github.com/OpenDreamProject/OpenDream/assets/2544323/78f01dd4-e506-4d60-ba60-3b634ef5bb19)

### LINKED ISSUE #2081: `client.eye` can't be set to a turf
ref says you can do 
```
client
  eye = locate(5,5,1)
```
but we currently error with "Cannot set eye to non-movable"

Ruzihm:
Adds a network message sent from the server to client to notify of mob/eye updates (currently, the client uses the attached entity which is ambiguous).  I wasn't sure of a better way to accomplish this but am looking for feedback.

The client then uses this info to distinguish between eye/mob for context menu, drawing, and verbs.

Closes #1581
Closes #2081


OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -339,6 +374,22 @@ private void RxUpdateClientInfo(MsgUpdateClientInfo msg) {
...
 
+    private void RxNotifyMobEyeUpdate(MsgNotifyMobEyeUpdate msg) {
+        var mob = _entityManager.GetEntity(msg.MobNetEntity);
+        if (mob.IsValid()) {
+            MobUid = mob;
+        } else {
+            MobUid = _playerManager.LocalSession?.AttachedEntity.GetValueOrDefault(EntityUid.Invalid) ?? EntityUid.Invalid;
+        }
+
+        var eye = _entityManager.GetEntity(msg.MobNetEntity);
```
ike709:
```suggestion
        var eye = _entityManager.GetEntity(msg.EyeNetEntity);
```
Bad copypasta?

Ruzihm: Yep, thanks! Fixed it if you would like to take another look.

OpenDreamShared/Network/Messages/MsgNotifyMobEyeUpdate.cs
```diff
@@ -0,0 +1,24 @@
...
+using Robust.Shared.Serialization;
+
+namespace OpenDreamShared.Network.Messages;
+
+//Server -> Client: Tell the client about the current entity UIDs of its mob and eye
+public sealed class MsgNotifyMobEyeUpdate : NetMessage {
+    public override MsgGroups MsgGroup => MsgGroups.EntityEvent;
+
+    public NetEntity MobNetEntity;
+    public NetEntity EyeNetEntity;
```
wixoaGit: While we don't currently support it, the eye can also be a turf. This should probably be a `ClientObjectReference` which can point to either a movable or a turf.

OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -1081,6 +1130,8 @@ public interface IDreamInterfaceManager {
     public bool ShowPopupMenus { get; }
     public int IconSize { get; }
     public CursorHolder Cursors { get; }
+    public EntityUid MobUid { get; }
+    public EntityUid EyeUid { get; }
```
wixoaGit: I think `DreamClientSystem` would be a better place for these and the message handlers.

OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -339,6 +372,22 @@ private void RxUpdateClientInfo(MsgUpdateClientInfo msg) {
...
+            MobUid = mob;
+        } else {
+            MobUid = _playerManager.LocalSession?.AttachedEntity.GetValueOrDefault(EntityUid.Invalid) ?? EntityUid.Invalid;
+        }
+
+        var eye = _entityManager.GetEntity(msg.EyeNetEntity);
+        if (eye.IsValid()) {
+            EyeUid = eye;
+        } else {
+            EyeUid = _playerManager.LocalSession?.AttachedEntity.GetValueOrDefault(EntityUid.Invalid) ?? EntityUid.Invalid;
```
wixoaGit: `MobUid` would be a better fallback.
```suggestion
            EyeUid = MobUid;
```

wixoaGit (changes requested):

OpenDreamShared/Network/Messages/MsgNotifyMobEyeUpdate.cs
wixoaGit: This would be best to do while also adding support for turf eyes actually

OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -131,9 +131,14 @@ public void Initialize() {
         _netManager.RegisterNetMessage<MsgLoadInterface>(RxLoadInterface);
         _netManager.RegisterNetMessage<MsgAckLoadInterface>();
         _netManager.RegisterNetMessage<MsgUpdateClientInfo>(RxUpdateClientInfo);
+        _netManager.RegisterNetMessage<MsgNotifyMobEyeUpdate>(RxNotifyMobEyeUpdate);
```
Ruzihm, line 138: I'm not sure why but when I tried to register this in `DreamClientSystem` I would get the following error & disconnect:

```
[WARN] net: 127.0.0.1:1212: Got net message with invalid ID 72.
[DEBG] net: "127.0.0.1:1212": Status changed to Disconnecting, reason: "Got NetMessage with invalid ID"
``` 

I don't know of a better way to handle this so I'm just leaving this here for now for proof of concept until a better method is suggested.

Ruzihm:
Setting eye to null will stop rendering - but since that's done by setting the eye position to be in nullspace, the viewport stops clearing the render target and you just get a frozen screen:

https://github.com/space-wizards/RobustToolbox/blob/aa5cca4c7f210f3ed7b7841735b5aadb533c9d88/Robust.Client/Graphics/Clyde/Clyde.HLR.cs#L494-L497

Does this warrant a pull request to RT to add a viewport option to allow target clearing if there is no eye/the eye is in null space?

OpenDreamClient/Interface/DreamInterfaceManager.cs
Ruzihm: I'm going to go ahead and mark this as resolved. if someone wants to reopen this discussion thats cool.

Ruzihm:
I tried my best to merge the refcounting change to Connection.Eye. I noticed that there was a redundant ref management outside of the `Eye` `set` at `DreamObjectClient.SetVar` so I removed it since it seems to be a duplicate of what happens inside the `Eye` set.

Ruzihm:
@wixoaGit I think this PR is good for another look when you get a moment

OpenDreamClient/Interface/DreamInterfaceManager.cs
wixoaGit: Net messages need to be registered before systems are initialized. If you want to send messages through a system, you need to create an entity event. That has some considerations to think about though, since it becomes a part of the game state and associated with a tick. That may be desirable since this has effects on your view of the game world, but it's not a big deal since we disable RT's network prediction.

OpenDreamClient/DreamClientSystem.cs
```diff
@@ -1,18 +1,89 @@
...
+    public EntityUid MobUid {
+        get {
+            // if entity of mob is invalid but net entity isn't, try referring to known net entities
+            if (! _mobUid.IsValid() && _mobNet.IsValid() && _entityManager.TryGetEntity(_mobNet, out var ent)) {
+                _mobUid = ent.GetValueOrDefault(EntityUid.Invalid);
+            }
+
+            return _mobUid;
+        }
+    }
```
wixoaGit: This is a lot of effort to avoid the single Dictionary lookup that `_entitySystem.GetEntity()` does. The same NetEntity also isn't guaranteed to always refer to the same EntityUid on the client due to PVS shenanigans, so the cached value could break.
```suggestion
    public EntityUid MobUid => _entityManager.GetEntity(_mobNet);
```

OpenDreamClient/DreamClientSystem.cs
```diff
@@ -1,18 +1,89 @@
...
     }
+
+    public void RxNotifyMobEyeUpdate(MsgNotifyMobEyeUpdate msg) {
+        var prevMobNet = _mobNet;
+        _mobNet = msg.MobNetEntity;
+
+        if (prevMobNet != _mobNet) {
+            // mark mob cache as dirty/invalid
+            _mobUid = EntityUid.Invalid;
+        }
```
wixoaGit:
```suggestion
        _mobNet = msg.MobNetEntity;
```

OpenDreamRuntime/DreamConnection.cs
```diff
@@ -66,14 +81,21 @@ [ViewVariables] public DreamObjectMob? Mob {
         }
     }
 
-    [ViewVariables] public DreamObjectMovable? Eye {
+    [ViewVariables] public ClientObjectReference? Eye {
```
wixoaGit: `ClientObjectReference` is intended to only be used on the client or in network communication with the client. Using it like this adds an annoying amount of indirection. This should be a `DreamObjectAtom?` and `DreamManager.GetClientReference()` should be used when creating the `MsgNotifyMobEyeUpdate`.

OpenDreamShared/Network/Messages/MsgNotifyMobEyeUpdate.cs
```diff
@@ -0,0 +1,55 @@
...
+namespace OpenDreamShared.Network.Messages;
+
+//Server -> Client: Tell the client about the current entity UIDs of its mob and eye
+public sealed class MsgNotifyMobEyeUpdate : NetMessage {
+    public override MsgGroups MsgGroup => MsgGroups.EntityEvent;
+
+    public NetEntity MobNetEntity;
+
+    // Type = Client -> null
+    public ClientObjectReference EyeRef;
```
wixoaGit: Would `ClientObjectReference?` work better here than treating `Client` as null?

OpenDreamShared/Network/Messages/MsgNotifyMobEyeUpdate.cs
```diff
@@ -0,0 +1,55 @@
...
+                int x = buffer.ReadInt32();
+                int y = buffer.ReadInt32();
+                int z = buffer.ReadInt32();
+                EyeRef = new(new(x, y), z);
+                break;
+            default:
+                // null eye
+                EyeRef = new();
+                break;
+        }
```
wixoaGit: I don't like having the process for `DreamClientReference` serialization in here. I'd rather this either be a helper method on `DreamClientReference`, use the `serializer` this method is given, or have this net message converted to an entity system event.

OpenDreamClient/Rendering/DreamPlane.cs
```diff
@@ -83,12 +83,15 @@ public void Draw(DreamViewOverlay overlay, DrawingHandleWorld handle, Box2 world
     /// <summary>
     /// Draws this plane's mouse map onto the current render target
     /// </summary>
-    public void DrawMouseMap(DrawingHandleWorld handle, DreamViewOverlay overlay, Vector2i renderTargetSize, Box2 worldAABB) {
+    public void DrawMouseMap(DrawingHandleWorld handle, DreamViewOverlay overlay, Vector2i renderTargetSize, Box2 worldAABB, bool onlyDrawScreenSprites = false) {
```
wixoaGit: `onlyDrawScreenSprites` is never used anywhere.

OpenDreamClient/Rendering/DreamViewOverlay.cs
```diff
@@ -145,25 +153,48 @@ protected override void Draw(in OverlayDrawArgs args) {
...
+                _mobSightQuery.TryGetComponent(mob, out eyeSight);
+                break;
+            case ClientObjectReference.RefType.Entity:
+                var eyeUid = _entityManager.GetEntity(eyeRef.Entity);
+                if (!_xformQuery.TryGetComponent(eyeUid, out var eyeTransform))
+                    return;
+                eyeCoords = _transformSystem.GetMapCoordinates(eyeUid, eyeTransform);
+                _mobSightQuery.TryGetComponent(eyeUid, out eyeSight);
+                break;
+        }
```
wixoaGit: Could this be turned into `IsNullEye`, `EyeCoords`, and `Sight` helper properties on `DreamClientEye`? It should be reachable from `Draw()`'s `args.Viewport.Eye`.

wixoaGit (changes requested):

Ruzihm:
@wixoaGit Okay, it should be good for another look when you're ready 👍

PR #2480, opened 2026-01-02
Makes EmitPushIsSaved virtual 

### LINKED ISSUE #1021: `EmitPushIsSaved` should be virtual
This has caused some confusion when compared to `EmitPushValue` and `EmitPushInitial` (both of which are virtuals on our Expression types) - all of these could maybe use minor changes to keep future code sane.

<https://discord.com/channels/847318834952405042/847318834952405045/1061715519310405783>

Ruzihm:
Closes #1021

### DIFF
--- a/DMCompiler/DM/DMExpression.cs
+++ b/DMCompiler/DM/DMExpression.cs
@@ -49,7 +49,7 @@ internal abstract class DMExpression(Location location) {
     }
 
     public virtual void EmitPushIsSaved(ExpressionContext ctx) {
-        ctx.Compiler.Emit(WarningCode.BadArgument, Location, $"can't get saved value of {this}");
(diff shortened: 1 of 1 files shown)


DMCompiler/DM/DMExpression.cs
```diff
@@ -48,6 +48,11 @@ public virtual DMReference EmitReference(ExpressionContext ctx, string endLabel,
         return DMReference.Invalid;
     }
 
+    public virtual void EmitPushIsSaved(ExpressionContext ctx) {
+        ctx.Compiler.Emit(WarningCode.BadArgument, Location, $"can't get saved value of {this}");
```
wixoaGit: This isn't related to a saved value, but whether the LValue itself is saved. The message could be clarified on that.
```suggestion
        ctx.Compiler.Emit(WarningCode.BadArgument, Location, $"can't get issaved() value of {this}");
```

PR #2481, opened 2026-01-03
makes .quit winset behave same as closing with windowbar x button

### LINKED ISSUE #1915: the `.quit` winset doesn't quit like pressing the x on the windowbar does
if you're connected through the launcher and `.quit` it causes the launcher to try a reconnect as if the client crashed. It doesn't do that if you close it with the window's title bar

Ruzihm:
Closes #1915

### DIFF
--- a/OpenDreamClient/Interface/DreamInterfaceManager.cs
+++ b/OpenDreamClient/Interface/DreamInterfaceManager.cs
@@ -452,7 +452,7 @@ internal sealed class DreamInterfaceManager : IDreamInterfaceManager {
     public void RunCommand(string fullCommand, bool repeating = false) {
         switch (fullCommand) {
             case not null when fullCommand.StartsWith(".quit"):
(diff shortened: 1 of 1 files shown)


OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -452,7 +452,7 @@ public void Prompt(DreamValueType types, string title, string message, string de
     public void RunCommand(string fullCommand, bool repeating = false) {
         switch (fullCommand) {
             case not null when fullCommand.StartsWith(".quit"):
-                IoCManager.Resolve<IClientNetManager>().ClientDisconnect(".quit used");
+                _gameController.Shutdown();
```
wixoaGit: Would be good to keep the reason.
```suggestion
                _gameController.Shutdown(".quit used");
```

wixoaGit (approved):

PR #2483, opened 2026-01-05
Support text box editing via setting text attribute

Ruzihm:
Adds the ability to set the text value of an input via winset.

**Test Setup**

1.  Starting with TestGame, add the following verb to mob:

```
	verb/set_text()
		set name = "Set Text"
		winset(src,"outputwindow.input","text=asdf")
```

2. Compile and start server
3. Click "Set Text" verb in Verbs tab (test case 1)
4. Clear input box of any content, then type `set-text` in input box and press enter (test case 2)
5. Clear input box of any content, then type `.winset "outputwindow.input.text="asdf""` in text box and press enter (test case 3)

**Test Results**

master OD:
The text box will be empty for all test cases

PR:
The text box will be set to `asdf` for all test cases

### DIFF
--- a/OpenDreamClient/Interface/Controls/ControlInput.cs
+++ b/OpenDreamClient/Interface/Controls/ControlInput.cs
@@ -9,11 +9,6 @@ namespace OpenDreamClient.Interface.Controls;
 internal sealed class ControlInput(ControlDescriptor controlDescriptor, ControlWindow window) : InterfaceControl(controlDescriptor, window) {
     private LineEdit _textBox = default!;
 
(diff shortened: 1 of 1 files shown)


wixoaGit:
Could you instead reset the text _before_ running the command?

Ruzihm:
> Could you instead reset the text _before_ running the command?

You know, I must have had a typo in an earlier test where I had something like:

`.winset "outputwindow.input.text="" ? outputwindow.input.text="empty" : outputwindow.input.text="not empty""`

because now when I try it in byond, it shows as empty. so yes that should work. 👍

PR #2484, opened 2026-01-06
shows results of removing client/mob verbs on client

### LINKED ISSUE #2342: Removing client verbs doesn't update the panels
Add a verb that does
```
/verb/thing()
    usr.verbs = list()
```

push button
note that verbs remain visible on the client but clicking them has no effect

Ruzihm:
Closes #2342.

The verb list classes didn't override `ContainsKey`, `ContainsValue`, or `RemoveValue`, which led to verb removal not working as expected. I also added an explicit verb refresh step to appearance updates on the client side. to ensure that atom verb updates are kept up to date on the client.

Also, while testing `listx.Remove(listx)` I had an issue where the list was being modified while iterated over so now list.remove will copy lists passed as arguments before iterating over them.

Testing code on top of testgame:

```
/mob 
	// ...

	verb/thing1()
		world.log << "thing1"
		for(var/i=src.verbs.len, i>=1, i--)
			src.verbs.Remove(src.verbs[i])
	
	verb/thing2()
		world.log << "thing2"
		src.verbs.Remove(src.verbs)

	verb/thing3()
		world.log << "thing3"
		for(var/i=src.verbs.len, i>=1, i--)
			src.verbs -= src.verbs[i]

	verb/thing4()
		world.log << "thing4"
		src.verbs -= src.verbs

/client		
	verb/thing5()
		world.log << "thing5"
		for(var/i=src.verbs.len, i>=1, i--)
			src.verbs.Remove(src.verbs[i])

	verb/thing6()
		world.log << "thing6"
		src.verbs.Remove(src.verbs)

	verb/thing7()
		world.log << "thing7"
		for(var/i=src.verbs.len, i>=1, i--)
			src.verbs -= src.verbs[i]

	verb/thing8()
		world.log << "thing8"
		src.verbs -= src.verbs
```

Results: `thing1-4` remove the mob verbs, `thing5-8` remove the client verbs.

### DIFF
--- a/OpenDreamRuntime/Procs/Native/DreamProcNativeList.cs
+++ b/OpenDreamRuntime/Procs/Native/DreamProcNativeList.cs
@@ -151,8 +151,8 @@ namespace OpenDreamRuntime.Procs.Native {
             foreach (var argument in args) {
                 if (argument.TryGetValueAsDreamList(out var argumentList)) {
                     if (argumentList ==  list) {
-                        list.Cut();
                         itemRemoved += list.GetLength();
+                        list.Cut();
                         break;
                     }
 


OpenDreamClient/Rendering/DreamIcon.cs
```diff
@@ -147,6 +147,7 @@ public void SetAppearance(uint? appearanceId, AtomDirection? parentDir = null) {
             }
 
             Appearance = appearance;
+            appearanceSystem.RefreshVerbs();
```
Ruzihm: This is causing way too much overhead when connected to a real server. Will need an alternative to this....

Ruzihm: I'm a little unsure what else could be done here. Ideally, we would only update verbs for clients which are gaining or losing executable verbs. 

But I'm wondering if it's acceptable to do something simple and just have the client periodically call `RefreshVerbs` on a timer of some kind...

OpenDreamClient/Rendering/ClientAppearanceSystem.cs
```diff
@@ -363,4 +370,11 @@ public string GetName(ClientObjectReference reference) {
     public Flick? GetMovableFlick(EntityUid entity) {
         return _movableFlicks.GetValueOrDefault(entity);
     }
+
+    private async Task StartVerbRefresher() {
+        while (true) {
+            await Task.Delay(TimeSpan.FromSeconds(timeToRefreshVerbs));
+            _taskManager.RunOnMainThread(_verbSystem.RefreshVerbs);
+        }
+    }
```
Ruzihm: This works but I'm not sure if this is the best way.

Ruzihm:
Also I noticed that when reading/clearing `client.verbs`, I can see/remove builtins like `/client/verb/Center`, `/client/verb/Northwest`, and so on. So clearing `client.verbs` actually makes the character stop responding to the arrow keys in `TestGame`. In Byond, they are not included in `client.verbs` and not removable. That's probably worthy of a different PR/issue.

OpenDreamRuntime/Procs/Native/DreamProcNativeList.cs
```diff
@@ -150,7 +151,11 @@ private static int ListRemove(IDreamList list, ReadOnlySpan<DreamValue> args) {
             var itemRemoved = 0;
             foreach (var argument in args) {
                 if (argument.TryGetValueAsDreamList(out var argumentList)) {
-                    foreach (DreamValue value in argumentList.EnumerateValues()) {
+                    // In case this is a "listx.Remove(listx)" situation, copy the contents here to avoid modification while enumerating
+                    // TODO: check for that case first to avoid unnecessary copy?
+                    var subtraction = argumentList.EnumerateValues().ToList();
```
wixoaGit: This will be a lot less performant with large lists. I would prefer instead doing a Cut() if it's the same list.

OpenDreamRuntime/Objects/Types/DreamList.cs
```diff
@@ -652,6 +652,21 @@ public override IEnumerable<DreamValue> EnumerateValues() {
             yield return new(verb);
     }
 
+    public override bool ContainsKey(DreamValue value) {
```
wixoaGit: This method is a little confusingly named. This looks for a key in the associative array, and an integer can never be one of those.

This should be removed.

OpenDreamRuntime/Objects/Types/DreamList.cs
```diff
@@ -716,6 +743,23 @@ public override IEnumerable<DreamValue> EnumerateValues() {
         }
     }
 
+    public override bool ContainsKey(DreamValue value) {
```
wixoaGit: This should also be removed.

OpenDreamClient/Rendering/ClientAppearanceSystem.cs
```diff
@@ -77,6 +83,8 @@ public override void Initialize() {
         SubscribeNetworkEvent<AnimationEvent>(OnAnimation);
         SubscribeNetworkEvent<FlickEvent>(OnFlick);
         SubscribeLocalEvent<DMISpriteComponent, WorldAABBEvent>(OnWorldAABB);
+
+        _ = StartVerbRefresher(new());
```
wixoaGit: I would prefer this to be a simple `_verbSystem.RefreshVerbs()` inside the `Update()` method, so it's a little more in tune to the server's update schedule.

[This](https://github.com/OpenDreamProject/OpenDream/blob/00d97ade9932b4269c926e9978f1c1993279dfdd/OpenDreamClient/ClientVerbSystem.cs#L245) can also be removed if periodically calling `RefreshVerbs()` is how we're going to go about it.

OpenDreamClient/Rendering/ClientAppearanceSystem.cs
```diff
@@ -99,6 +102,12 @@ public override void Update(float frameTime) {
             if (flick.GetAnimationFrame(_gameTiming) == -1)
                 _movableFlicks.Remove(flickKey);
         }
+
+        _refreshVerbRemainingTime -= frameTime;
+        if (_refreshVerbRemainingTime < 0) {
+            _refreshVerbRemainingTime = _refreshVerbPeriod;
+            _verbSystem.RefreshVerbs();
```
wixoaGit: This method is taking 0.002 to 0.01 milliseconds to execute on my machine. I don't think the 1 second timer is necessary.

OpenDreamRuntime/Procs/Native/DreamProcNativeList.cs
```diff
@@ -150,6 +150,11 @@ private static int ListRemove(IDreamList list, ReadOnlySpan<DreamValue> args) {
             var itemRemoved = 0;
             foreach (var argument in args) {
                 if (argument.TryGetValueAsDreamList(out var argumentList)) {
+                    if (argumentList ==  list) {
+                        list.Cut();
+                        continue;
```
wixoaGit: The removal count needs updated, and there's nothing left in the list to remove so we can break out of the loop early.
```suggestion
                        itemRemoved += list.GetLength();
                        break;
```

OpenDreamRuntime/Procs/Native/DreamProcNativeList.cs
```diff
@@ -150,6 +150,12 @@ private static int ListRemove(IDreamList list, ReadOnlySpan<DreamValue> args) {
             var itemRemoved = 0;
             foreach (var argument in args) {
                 if (argument.TryGetValueAsDreamList(out var argumentList)) {
+                    if (argumentList ==  list) {
+                        list.Cut();
+                        itemRemoved += list.GetLength();
```
wixoaGit: Whoops, gotta do the `GetLength()` _before_ the `Cut()`
```suggestion
                        itemRemoved += list.GetLength();
                        list.Cut();
```

Ruzihm:
@wixoaGit I think this may have caused an issue where now the verbs can be a little hard to click, like sometimes I click a verb and nothing happens. I didn't notice that when i was testing this while it included the sleep.

Have you noticed anything like that since this was merged?

PR #2495, opened 2026-01-23
Back macro won't interfere with holding backspace deleting multiple characters

### LINKED ISSUE #1583: Holding backspace doesn't remove more than one string
Issue on SS14 Paradise Test

Ruzihm:
Closes #1583

**Test case setup:**

Edit TestGame's TestInterface.dmf and add a new macro to `macro` list: 

* Action: `Back`
* Command: `say hi`

**Test case:**

1. Click on map then press backspace.
3. Click on text input then press backspace.
2. Click on text input, input `asdf`, then hold backspace for a few seconds.

Byond:
1. Command happens (`xyz says: "hi"` appears in output)
2. The command happens once.
3. Whole input gets emptied, then the command happens once.

OD Before PR:
1. Command happens
2. **The command does not happen.**
3. **Only `f` gets removed. Command does not happen.**

OD After PR:
1. Command Happens
2. **The command does not happen**
3. Whole input gets emptied. **Command does not happen.**

The remaining disparities here (in bold) basically require a couple RT changes. Making these changes to RT and rebuilding bring the test case to full parity.

1. A `LineEdit` that calls `args.Handle()` in `KeyBindDown` under different conditions - in this case, in backspace it would only be called when it actually removes character(s). Other controls would need similar treatment, for situations like writing on paper.
2. A change to `UserInterfaceManager.OnUIKeyBindStateChanged` where it can require `args.Handled` to also be true in order to return true.

### DIFF
--- a/OpenDreamClient/Interface/DreamInterfaceManager.cs
+++ b/OpenDreamClient/Interface/DreamInterfaceManager.cs
@@ -23,7 +23,6 @@ using Robust.Shared.Utility;
 using SixLabors.ImageSharp;
 using System.Linq;
 using Robust.Shared.Map;
-using Robust.Shared.Input;
 
 namespace OpenDreamClient.Interface;
 


OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -103,14 +104,31 @@ public void LoadInterfaceFromSource(string source) {
...
         // Set up the middle-mouse button keybind
         _inputManager.Contexts.GetContext("common").AddFunction(OpenDreamKeyFunctions.MouseMiddle);
         _inputManager.RegisterBinding(new KeyBindingRegistration() {
             Function = OpenDreamKeyFunctions.MouseMiddle,
             BaseKey = Keyboard.Key.MouseMiddle
         });
 
+        // Ensure macros can't prevent backspace from deleting text on repeat
+        // Remove the default backspace binding
+        _inputManager.RemoveBinding(_inputManager.GetKeyBinding(EngineKeyFunctions.TextBackspace));
```
wixoaGit: Does removing the bind from [`keybinds.yml`](https://github.com/OpenDreamProject/OpenDream/blob/afd59dcb2272b5406dd9d10ee739d4f6a0dbe793/Resources/keybinds.yml#L141) achieve the same thing?

Looks like you could also set its priority there instead.

Ruzihm: .....yes.  Thank you lol. I looked for this in vscode but of course resources isn't included in the search 😄

PR #2498, opened 2026-01-25
Overlays and underlays inherit from iconstate from parent when iconstate is null

### LINKED ISSUE #2291: /image objects with null icon_state don't inherit the icon_state from the atom they're overlaying
In Baystation, airlocks use `/image` objects as overlays without an icon_state to render airlock bolt/opening lights.
`/image` objects inherit the icon_state from the atom they're overlaying recursively, if the `/image` doesn't have an `icon_state` set.
I'm told the reasoning is this:
> [16:22] its done this way specifically for animated states, because updating the overlays when changing icon states is a lot slower than just having the inbuilt engine overlay manager automatically update all the overlays icon states to match the parent icon state. Otherwise all the overlays would be a frame late than the parent animated icon state
[16:23] cant remember what the function is called but when you call the animate proc it also does it for overlays
[16:23] isnt it like flick() or something
[16:23] I think so
[16:23] that way you can just flick("parent icon state name") and itll do all the overlays too

OpenDream does not follow this, and just assumes a null `icon_state` for the image.

Smallest reproducable example in TestGame:
1. Make a new file called `mob_overlays.dmi` and give it two icon states, one with no name with an X, and the other with the mob's `icon_state` name, in this case `mob`, and make it a nose or something.
2. Image
3. Run the verb
4. You have an X in OpenDream, but in DreamSeeker you have a nose.
```ts
/mob
    icon = 'icons/mob.dmi'
    icon_state = "mob"

/mob/verb/give_nose()
    usr << "Enjoy your new nose!"
    var/image/I = image('icons/mob_overlays.dmi', null)
    src.overlays += I
```

### Expected
Image
### Actual
Image

Another property that I didn't know about is that recursively overlayed `/image`s also inherit the icon_state from their parent.
```ts
/mob
    icon = 'icons/mob.dmi'
    icon_state = "mob"

/mob/verb/give_nose()
    usr << "Enjoy your new nose!"
    var/image/I = image('icons/mob_overlays.dmi', null)
    var/image/II = image('icons/mob_overlays.2.dmi', null)
    var/image/III = image('icons/mob_overlays.3.dmi', null)
    II.overlays += III
    I.overlays += II
    src.overlays = I
```

Ruzihm:
Closes #2291

### DIFF
--- a/OpenDreamClient/Rendering/DreamIcon.cs
+++ b/OpenDreamClient/Rendering/DreamIcon.cs
@@ -147,12 +147,7 @@ internal sealed class DreamIcon(RenderTargetPool renderTargetPool, IDreamInterfa
                 _direction = appearance.Direction;
             }
 
-            if (parentIconState != null && appearance.IconState == null) {
-                _iconState = parentIconState;
-            } else {
-                _iconState = appearance.IconState;
-            }
-
+            _iconState = appearance.IconState ?? parentIconState;
             Appearance = appearance;
         });
     }


OpenDreamClient/Rendering/DreamIcon.cs
```diff
@@ -146,6 +147,12 @@ public void SetAppearance(uint? appearanceId, AtomDirection? parentDir = null) {
                 _direction = appearance.Direction;
             }
 
+            if (parentIconState != null && appearance.IconState == null) {
+                _iconState = parentIconState;
+            } else {
+                _iconState = appearance.IconState;
+            }
+
```
wixoaGit:
```suggestion
            _iconState = appearance.IconState ?? parentIconState;
```

PR #2517, opened 2026-01-30
Implement `caller`

ike709:
https://github.com/OpenDreamProject/OpenDream/issues/2203

This is admittedly a bit awful but the alternative is every single proccall storing its `ProcState` on the called proc even when nobody utilizes or cares about `caller`.

### DIFF
--- a/Content.Tests/DMProject/Tests/Builtins/caller.dm
+++ b/Content.Tests/DMProject/Tests/Builtins/caller.dm
@@ -1,0 +1,16 @@
+/datum/meep/proc/bar()
+    ASSERT(caller.caller.caller.name == "ihateithere")
+
+/datum/proc/foo()
+    var/datum/meep/M = new
+    ASSERT(caller.name == "ihateithere")
+    M.bar()
+
+/proc/ihateithere()
+	var/datum/D = new
+	ASSERT(caller.name == "RunTest")
+	D.foo()
+
+/proc/RunTest()
+    ASSERT(caller.name == "New")
+    ihateithere()
--- a/Content.Tests/DMProject/Tests/Builtins/caller.dm
+++ b/Content.Tests/DMProject/Tests/Builtins/caller.dm
@@ -12,5 +12,5 @@
 	D.foo()
 
 /proc/RunTest()
-    ASSERT(caller.name == "New")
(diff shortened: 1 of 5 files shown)


Content.Tests/DMProject/Tests/Builtins/caller.dm
```diff
@@ -0,0 +1,16 @@
+/datum/meep/proc/bar()
+    ASSERT(caller.caller.caller.name == "ihateithere")
```
wixoaGit, line 2: Should this not be `"RunTest"`?

Could also be using `nameof(/proc/RunTest)` for these asserts instead.

OpenDreamRuntime/Objects/Types/DreamObjectCallee.cs
```diff
@@ -54,6 +64,19 @@ protected override bool TryGetVar(string varName, out DreamValue value) {
         }
     }
 
+    /// <summary>
+    /// Sets <see cref="_caller"/> if it hasn't been already and <see cref="ProcState"/> is not null
+    /// </summary>
+    private void SetCaller() {
+        if (ProcState is null || _caller is not null) return;
+        if (ProcState.Thread.PeekStack(1) is not DMProcState dmProcState) return;
```
wixoaGit, line 72: Always passing an index of 1 is incorrect

OpenDreamRuntime/Objects/Types/DreamObjectCallee.cs
```diff
@@ -3,39 +3,49 @@
...
-                value = new(ProcState.Proc);
+                value = ProcState.Proc != null ? new(ProcState.Proc) : DreamValue.Null;
                 return true;
             case "args":
                 value = new(new ProcArgsList(ObjectTree.List.ObjectDefinition, ProcState));
                 return true;
             case "caller":
-                // TODO
-                value = DreamValue.Null;
+                SetCaller(); // sometimes ProcState is null in the constructor?
```
wixoaGit, line 28: Why _wouldn't_ `ProcState` always be null in the constructor? It's not passed to the constructor or set in there.

wixoaGit (changes requested):

ike709:
will revisit when I have time

PR #2524, opened 2026-03-06
Add GetPixel() for Icon

Sorrowfulwinds:
Adds proc/GetPixel(x, y, icon_state, dir = 0, frame = 0, moving = -1) for #2194.

It has 2 TODOs that cannot be implemented currently:
-The moving var does nothing as OD does not currently track that variable from DMI's
-If icon_state is not in the icon BYOND will either pull the empty string/null icon_state by default or if one does not exist it will pull the first icon_state in the dmi. OD does not track this so it only defaults to checking for an empty string state before failing.

Vars:
X&Y - Must be >1 and less than the max dimension +1 of the icon_state.
icon_state - Must be a string.
dir - Must be a valid integer of dirs N E S W NE NW SE SW (1,2,4,5,6,8,9,10) or 0. 0 counts as South.
frame - Values less than 1 count as 1. Must be less than the total frames +1 of the icon_state.
moving - not implemented


OpenDreamRuntime/Procs/Native/DreamProcNativeIcon.cs
```diff
@@ -107,5 +107,76 @@ public static DreamValue NativeProc_Turn(NativeProc.Bundle bundle, DreamObject?
...
+            bundle.GetArgument(1, "y").TryGetValueAsInteger(out var yPos);
+            bundle.GetArgument(2, "icon_state").TryGetValueAsString(out var iconState);
+            bundle.GetArgument(3, "dir").TryGetValueAsInteger(out var dir);
+            bundle.GetArgument(4, "frame").TryGetValueAsInteger(out var frame);
+            //TODO: Implement moving var
+
+            DreamIcon iconObj = ((DreamObjectIcon)src!).Icon;
+            //TODO BYONDISM: A non-existent icon_state will default to the empty string icon_state,
+            //or if one doesn't exist it will default to the first icon_state in the DMI.
+            if(!iconObj.States.TryGetValue(iconState ?? string.Empty, out var state))
```
wixoaGit, line 129: Using `DreamIcon.States` is going to get you the textures before any sort of modifications like `/icon.Blend()` are applied. You need to use the result of `iconObj.GenerateDMI()` to get the final DMI.

OpenDreamRuntime/Procs/Native/DreamProcNativeIcon.cs
```diff
@@ -107,5 +107,76 @@ public static DreamValue NativeProc_Turn(NativeProc.Bundle bundle, DreamObject?
...
+                4 => AtomDirection.East,
+                5 => AtomDirection.Northeast,
+                6 => AtomDirection.Southeast,
+                8 => AtomDirection.West,
+                9 => AtomDirection.Northwest,
+                10 => AtomDirection.Southwest,
+                _ => AtomDirection.None
+            };
+            //BYONDISM: Bad dir values just crash instantly :)
+            if (atomDir == AtomDirection.None) return DreamValue.Null;
```
wixoaGit, line 156: You can cast the value instead of having a switch
```suggestion
            //BYONDISM: Bad dir values just crash instantly :)
            if (!Enum.IsDefined(typeof(AtomDirection), dir)) return DreamValue.Null;
            
            var atomDir = (AtomDirection)dir;
```

wixoaGit (changes requested):

amylizzle:
So I just PR'd this #2574 because I didn't see you'd already done it lol

Looks like we've each got bits the other missed, I suggest we combine them

wixoaGit:
Closing in favor of #2574

PR #2543, opened 2026-03-27
fix CanConstFold not being set by opendream_compiletimereadonly

### LINKED ISSUE #2540: opendream_compiletimereadonly doesn't work inside proc code
```
/turf/Entered(atom/movable/A, atom/OL, ignoreRest = FALSE)
    ..()
    src.loc = null
```

loc is marked as `opendream_compiletimereadonly`. It's caught if in an object definition, but not in that.

Toastical:
At some point, `CompiletimeReadonly` was moved from being a flag on `ValType.Type` to being a bool on `ValType`. It was checking the flag and always returned false. Fixes #2540

Tested by marking a variable as `opendream_compiletimereadonly` in DMStandard and trying to change it in DM.
```
Error OD0501 at test.dm:2:2: Cannot write to const var
Compilation failed with 1 errors and 1 warnings
```

### DIFF
--- a/DMCompiler/DM/DMVariable.cs
+++ b/DMCompiler/DM/DMVariable.cs
@@ -17,7 +17,7 @@ internal sealed class DMVariable {
     /// </remarks>
     public readonly bool IsConst;
 
-    private bool CanConstFold => (IsConst || ValType.Type.HasFlag(DMValueType.CompiletimeReadonly)) &&
+    private bool CanConstFold => (IsConst || ValType.IsCompileTimeReadOnly) &&
                                  !ValType.Type.HasFlag(DMValueType.NoConstFold);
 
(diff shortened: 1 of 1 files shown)


wixoaGit:
I'm only seeing one failing test in the CI, `List/vars_list.dm`:
```dm
var/datum/test/t = new()
var/varlist = t.vars
ASSERT(length(varlist) == 6) // normal_var, static_var, tag, type, parent_type, vars
```
The compiler is constant folding `t.vars` to null now that it believes it can be constant folded, since it's marked with `opendream_compiletimereadonly`.

I'm not sure why we even try to const-fold compiletimereadonly vars, since the point of it is to mark a var as unwriteable yet can still change.

PR #2545, opened 2026-03-30
diff verbs instead of recreating every time

### LINKED ISSUE #2520: The entire verb panel is recreated when refreshed
Right now [`RefreshVerbs()`](https://github.com/OpenDreamProject/OpenDream/blob/3b148fc70a477e577fa09c817887103373276864/OpenDreamClient/Interface/Controls/ControlInfo.cs#L220) removes every UI element and recreates them. This leads to some adverse effects that had to be worked around in #2513, and it would probably be better to just remove the buttons that are no longer in use.

hry-gh:
prevent flicker and fix #2520

### DIFF
--- a/OpenDreamClient/Interface/Controls/ControlInfo.cs
+++ b/OpenDreamClient/Interface/Controls/ControlInfo.cs
@@ -183,6 +183,7 @@ internal sealed class VerbPanel : InfoPanel {
     private readonly ClientVerbSystem? _verbSystem;
 
     private readonly VerbPanelGrid _grid;
+    private readonly Dictionary<(int VerbId, ClientObjectReference Src), Button> _verbButtons = new();
 
     private Color _highlightColor;
     private Color _textColor;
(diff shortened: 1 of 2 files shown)


OpenDreamClient/Interface/Controls/ControlInfo.cs
```diff
@@ -218,35 +219,53 @@ public override void UpdateElementDescriptor(ControlDescriptorInfo descriptor) {
     }
 
     public void RefreshVerbs(IEnumerable<(int, ClientObjectReference, VerbSystem.VerbInfo)> verbs) {
-        _grid.Children.Clear();
+        var panelVerbs = verbs
+            .Where(v => v.Item3.GetCategoryOrDefault(DefaultVerbPanel) == PanelName)
+            .Order(VerbNameComparer.OrdinalInstance)
+            .ToList();
+
+        var seenKeys = new HashSet<(int, ClientObjectReference)>(panelVerbs.Count);
```
wixoaGit: The `ToList()` forces an extra enumeration and allocates a new list, which slows things down.
```suggestion
            .Order(VerbNameComparer.OrdinalInstance);

        var seenKeys = new HashSet<(int, ClientObjectReference)>();
```

OpenDreamClient/Interface/Controls/ControlInfo.cs
```diff
@@ -218,35 +219,53 @@ public override void UpdateElementDescriptor(ControlDescriptorInfo descriptor) {
...
-            verbButton.OnMouseEntered += _ => {
-                verbButton.Label.FontColorOverride = _highlightColor;
-            };
+            gridIndex++;
+        }
 
-            verbButton.OnMouseExited += _ => {
-                verbButton.Label.FontColorOverride = _textColor;
-            };
+        foreach (var key in _verbButtons.Keys.ToList()) {
```
wixoaGit: You can `foreach` an IEnumerable directly instead of the extra enumeration and List allocation.
```suggestion
        foreach (var key in _verbButtons.Keys) {
```

OpenDreamClient/Interface/Controls/ControlInfo.cs
```diff
@@ -218,35 +219,53 @@ public override void UpdateElementDescriptor(ControlDescriptorInfo descriptor) {
...
+            if (seenKeys.Contains(key)) continue;
 
-            _grid.Children.Add(verbButton);
+            _grid.Children.Remove(_verbButtons[key]);
```
wixoaGit: This ends up calling the same code but I like the directness better.
```suggestion
            _grid.RemoveChild(_verbButtons[key]);
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #2546, opened 2026-03-30
ignore lack of byondapi dll if we don't have one

hry-gh:
it's a bit painful testing on macos which we can't compile our "byond"api dlls


wixoaGit:
What happens if a DLL tries calling a byondapi method after this fails?

There's also the repository at https://github.com/OpenDreamProject/OpenDreamByondApi if anyone wants to get this working on macOS and can test it.

Issue #2443, opened 2025-11-03
`/database/query.GetColumn()` should be 1-indexed, not 0-indexed


amylizzle:
`query.GetColumn(0)`  is bad, should be `query.GetColumn(1)`

PR #2547, opened 2026-03-30
database 1 indexing and query.Execute() runtimes

### LINKED ISSUE #2442: `/database/query.Execute()` should runtime if database is not specified at least once
This should succeed:
```
/proc/RunTest()	
	var/database/db = new("database.db")

	var/database/query/query = new("CREATE TABLE test (id int, name string, points float)")
	query.Execute(db)
```

This should fail:
```
/proc/RunTest()	
	var/database/db = new("database.db")

	var/database/query/query = new("CREATE TABLE test (id int, name string, points float)")
	query.Execute() //note the lack of arg
```

After the first use of `Execute()` with a db, any further uses of the query without the db argument should operate on the db specified in the first query use.

hry-gh:
fixes #2442 and #2443


OpenDreamRuntime/Objects/Types/DreamObjectDatabaseQuery.cs
```diff
@@ -98,12 +101,13 @@ public DreamValue GetColumn(int id) {
             return DreamValue.Null;
         }
 
+        id = Math.Max(--id, 0);
```
wixoaGit, line 99: So `<= 0` is equal to 1 here in byond?

Should be added to our tests if so

wixoaGit (changes requested):

wixoaGit:
This will need updated with reference counting.

Issue #2427, opened 2025-10-26
Actual parity issues from the BYOND tests


amylizzle:
- [ ] TEST FAILED: Tests/Operators/Division.dm -- appears to be an actual parity error 10/null should be 10
- [ ] TEST FAILED: Tests/Operators/valid_and_null.dm -- runtime error: Undefined operation: 2 / null
- [ ] TEST FAILED: Tests/List/ListFind.dm -- Tests/List/ListFind.dm:15:Assertion Failed: L.Find("b", 1, 2) == 2
- [ ] TEST FAILED: Tests/Sleeping/New.dm -- runtime error: Tests/Sleeping/New.dm:5:Assertion Failed: call_number == num
- [ ] TEST FAILED: Tests/Math/minmax.dm -- runtime error: Tests/Math/minmax.dm:8:Assertion Failed: min(0, null) == null
- [ ] TEST FAILED: Tests/Text/Splittext.dm -- runtime error: Tests/Text/Splittext.dm:9:Assertion Failed: test2 ~= test2_expected (probably due to 515 behaviour change)
- [ ] TEST FAILED: Tests/Text/HoleEmbeddedExpression.dm -- compile error: Tests/Text/HoleEmbeddedExpression.dm:3:error: "TEST": extra argument - parity break is us handling [ ]
- [ ] TEST FAILED: Tests/Builtins/Max.dm -- runtime error: Tests/Builtins/Max.dm:16:Assertion Failed: max("", null) == ""
- [ ] TEST FAILED: Tests/Savefile/ExportText.dm -- runtime error: Tests/Savefile/ExportText.dm:27:Assertion Failed: F.dir ~= list("dir","dir2","dir3","dir4","dir5","dir6","dir7","dirIcon","list") -- list equivalence operator cares about ordering now?
- [ ] TEST FAILED: Tests/Builtins/params2list.dm -- runtime error: Tests/Builtins/params2list.dm:6:Assertion Failed: params2list("a=1;a=2") ~= list(a="2") -- params2list with duplicate key now does key=list(value1, value2)
- [ ] TEST FAILED: Tests/Builtins/world_config.dm -- runtime error: Tests/Builtins/world_config.dm:20:Assertion Failed: world.GetConfig("env", env_var) == null -- should return empty string

ike709:
[...]

PR #2551, opened 2026-04-05
Fix `params2list()` duplicate key behavior

ike709:
~~At some point BYOND changed duplicate key behavior from "use the last value" to "make a list of all the values".~~

From [the ref](https://www.byond.com/docs/ref/#/proc/params2list):

> If you have multiple items with the same name, they will be combined into a list of text strings. For example, "key=value1;key=value2" would set list["key"] to a list containing "value1" and "value2", not necessarily in that order.

See also https://github.com/OpenDreamProject/OpenDream/pull/2444 and https://github.com/OpenDreamProject/OpenDream/issues/2427. This ticks a box on the latter.

### DIFF
--- a/OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
+++ b/OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
@@ -1886,9 +1886,11 @@ internal static class DreamProcNativeRoot {
                     }
                 }
             } else {
-                string queryValue = queryValues[^1]; //Use the last appearance of the key in the query
-
-                list.SetValue(new DreamValue(queryKey), new DreamValue(queryValue));
+                // NOTE: At some point BYOND's handling of duplicate keys changed from "use the last value" to "create a list of values for that key"
+                if (queryValues.Length > 1)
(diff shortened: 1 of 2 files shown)


OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
```diff
@@ -1886,9 +1886,11 @@ public static DreamList Params2List(DreamObjectTree objectTree, string queryStri
                     }
                 }
             } else {
-                string queryValue = queryValues[^1]; //Use the last appearance of the key in the query
-
-                list.SetValue(new DreamValue(queryKey), new DreamValue(queryValue));
+                // NOTE: At some point BYOND's handling of duplicate keys changed from "use the last value" to "create a list of values for that key"
```
wixoaGit: It looks like this never changed, and was just wrong from the beginning. I checked the ref up to BYOND v506 and it still said this was the case.

wixoaGit (changes requested):

OpenDreamRuntime/Procs/Native/DreamProcNativeRoot.cs
ike709:
```suggestion
```

wixoaGit (approved):

PR #2552, opened 2026-04-05
Fix `world.GetConfig()` parity

ike709:
Ticks a box on https://github.com/OpenDreamProject/OpenDream/issues/2427

See also #2444


ike709:
Looks like I need to do some deeper digging on how GetConfig works and I'm busy right now.

PR #2558, opened 2026-05-08
Control Flow Graph

ike709:
This is _extremely_ not ready yet to the point this PR will likely go stale and be closed. But I was tired of losing track of the branch and also it gives other people a chance to prod at and point out the (likely stupid) things I am doing.

I haven't even properly looked over this code in months.

PR #2561, opened 2026-05-09
CFG Part 0: Max Stack Size & Peephole Optimizer Fixes/Improvements

ike709:
This PR does a few things:
- Steals some redundant `Jump` optimizations from https://github.com/OpenDreamProject/OpenDream/pull/2558
- Fixes an issue with a case in which we did not rescan opcodes after calling `AttemptCurrentOpt()`
- Changes the peephole optimizer to keep rerunning until each pass reports no changes to the bytecode. This provides further gains in some operations involving stack pushes and list creations, and generally makes it less annoying than trying to spread the optimizations out across separate passes in the correct order. This could also lead to further simplification of the peephole optimizer in the future, but I didn't want to change too much all at once.
- Fixes some incorrect stack size/delta info
- Fixes max stack size not being recomputed after optimizations. This leads to reductions in stack size across thousands of procs.
- `switch()` cases no longer emit a `Jump` instruction if the last operation is a terminal change in control flow (e.g. `return`)
- `spawn()` blocks no longer emit `PushNull` and `Return` if the last operation is a terminal change in control flow (e.g. `return`)
- Peephole optimizer now ignores `AnnotatedBytecodeVariable` during opt scanning
- Pointless `Jump` instructions are more aggressively pruned by the bytecode optimizer
- Made some fixes to ProcDecoder that arose during testing

I did before & after diffs on /tg/station compilation & disassembly to confirm that this resulted in further optimizations to the bytecode without introducing any apparent regressions. The `NullRef` opt in particular was being left on the table in thousands of cases, while the other peephole optimizer changes are less dramatic but still present.

The fixes to max stack size are also a huge improvement. Thousands of procs have either had their max stack size reduced to 0 or have seen a reduction from double to single digits. I ran TG through round init to confirm nothing exploded.

This sufficiently beats the codegen and optimizer into submission to the point that we can now implement dead code detection (after CFG is finished) and run it on /tg/station and goonstation without getting false positives.


ike709:
I'm calling this PR done barring any review comments. I did another testing pass without any of my other non-committed experimental changes, and confirmed that TG and goon both compile, while also making sure I can start a round on TG.

DMCompiler/DM/DMProc.cs
```diff
@@ -856,6 +858,57 @@ public void Jump(string label) {
         WriteLabel(label);
     }
 
+    public bool LastInstructionTransfersControl() {
+        List<IAnnotatedBytecode> bytecode = AnnotatedBytecode.GetAnnotatedBytecode();
+        HashSet<string>? referencedLabels = null;
+        // Man sometimes it'd be nice if we had a list of just instructions and didn't need loops like this just to grab the last actual instruction
```
wixoaGit, line 864: You could also save the last opcode written by `WriteOpcode()`

DMCompiler/DM/DMProc.cs
```diff
@@ -856,6 +858,57 @@ public void Jump(string label) {
...
+        for (int i = bytecode.Count - 1; i >= 0; i--) {
+            switch (bytecode[i]) {
+                case AnnotatedBytecodeVariable:
+                    continue;
+                case AnnotatedBytecodeLabel label:
+                    referencedLabels ??= GetReferencedLabels(bytecode);
+                    if (referencedLabels is not null && referencedLabels.Contains(label.LabelName))
```
wixoaGit, line 871: Wouldn't this not work if the label is only referenced later on in the code?

DMCompiler/DM/Builders/DMProcBuilder.cs
```diff
@@ -242,7 +244,8 @@ private void ProcessStatementIf(DMASTProcStatementIf statement) {
             proc.StartScope();
             ProcessBlockInner(statement.Body);
             proc.EndScope();
-            proc.Jump(endLabel);
+            if (!proc.LastInstructionTransfersControl())
```
wixoaGit, line 247: Could `LastInstructionTransfersControl()` be put into `Jump()` instead of littering these checks everywhere?

DMCompiler/Optimizer/AnnotatedByteCodeWriter.cs
```diff
@@ -230,6 +245,27 @@ public int GetMaxStackSize() {
         return _maxStackSize;
     }
 
+    /// <summary>
+    /// Recomputes the maximum possible stack size from the current annotated bytecode.
+    /// Used after optimization because peephole rewrites can change the max stack size.
+    /// </summary>
+    public int RecalculateMaxStackSize() {
```
wixoaGit, line 252: Why bother tracking stack size while generating opcodes if we just recalculate it at the end?

DMCompiler/Optimizer/PeepholeOptimizer.cs
```diff
@@ -109,37 +109,64 @@ private void GetOptimizations() {
...
-        }
+        bool changed;
+
+        do {
+            changed = false;
+
+            foreach (var optPass in _passes) {
+                changed |= RunPass((byte)optPass, input);
+            }
+        } while (changed);
```
wixoaGit, line 120: This adds a second pass to practically every proc at minimum. I'd like to avoid that if possible, what additional optimizations is this catching?

DMCompiler/Optimizer/PeepholeOptimizer.cs
```diff
@@ -109,37 +109,64 @@ private void GetOptimizations() {
...
-            if (currentOpt == null)
+            if (currentOpt == null || currentOptStartIndex == -1)
                 return 0;
 
             int offset;
 
-            if (currentOpt.Optimization?.CheckPreconditions(input, i - optSize) is true) {
-                currentOpt.Optimization.Apply(_compiler, input, i - optSize);
+            int startIndex = currentOptStartIndex;
+            var skippedVariables = RemoveTransparentVariables(input, startIndex);
```
wixoaGit, line 136: Removing every `AnnotatedBytecodeVariable` surrounding every instruction, checking for a possible optimization, then inserting those variables back, seems ridiculous. Is there no better way to do this?

Not to mention all the `O(n)` operations this introduces.

DMCompiler/Optimizer/BytecodeOptimizer.cs
```diff
@@ -117,6 +128,76 @@ private void JoinAndForwardLabels(List<IAnnotatedBytecode> input) {
         }
     }
 
+    private void RemoveJumpsAfterTerminalInstructions(List<IAnnotatedBytecode> input) {
+        var labelAliases = RemoveJumpsAfterTerminalInstructions(input, null);
+        RewriteLabelAliases(input, labelAliases);
+    }
+
+    private Dictionary<string, string>? RemoveJumpsAfterTerminalInstructions(List<IAnnotatedBytecode> input, Dictionary<string, string>? labelAliases) {
```
wixoaGit, line 136: Could we just avoid emitting these in `DMProc` in the first place, instead of looping through every opcode and potentially running multiple O(n) removals?

wixoaGit (changes requested):

ike709:
Superceded by https://github.com/OpenDreamProject/OpenDream/pull/2635, see that PR for details.

tl;dr splitting this off into the successor "MVP for CFG" PR, and a separate max stack changes PR coming soon(TM).

Expensive peephole optimizer changes are effectively being abandoned.

PR #2572, opened 2026-05-15
Only send animation events to clients in PVS range

wixoaGit:
There's no need to send `animate()` and `flick()` events to every client, they don't do anything if the client has no knowledge of the relevant entity.


amylizzle:
I believe animations should play if an object that was animated comes into view, though idk if we handled that before

wixoaGit:
We didn't. The client would throw out the event because the entity didn't exist client-side.

PR #2573, opened 2026-05-17
Implements various operators for ALists

### LINKED ISSUE #2588: `DreamAssocList` missing `OperatorAppend`/`OperatorRemove`
running goonstation under opendream produces ~7200 of these per session:

```
System.InvalidOperationException: Cannot append ... to /alist
   at OpenDreamRuntime.Objects.DreamObject.OperatorAppend(DreamValue b)
```

`DreamAssocList` inherits from `DreamObject` not `DreamList` and never got operator overrides. goonstation's spatial hashmap uses `alist += entry` everywhere to track mob positions so everything was crashing constantly.

fix adds `OperatorAppend`/`OperatorRemove` to `DreamAssocList` and `TryGetValueAsIDreamList` to `DreamValue` since the existing `TryGetValueAsDreamList` cast fails for alists.

---

`OpenDreamRuntime/DreamValue.cs` — add after `TryGetValueAsDreamList`:

```csharp
public readonly bool TryGetValueAsIDreamList([NotNullWhen(true)] out IDreamList? list) {
    if (_refValue is IDreamList iList) {
        list = iList;
        return true;
    }
    list = null;
    return false;
}
```

`OpenDreamRuntime/Objects/Types/DreamAssocList.cs` — add before final `}`:

```csharp
[MustDisposeResource]
public override DreamValue OperatorAppend(DreamValue b) {
    if (b.TryGetValueAsIDreamList(out var bList)) {
        foreach (var value in bList.EnumerateValues()) {
            AddValue(value);
            if (bList.GetAssociativeValues().TryGetValue(value, out var assocValue))
                SetValue(value, assocValue);
        }
    } else {
        AddValue(b);
    }

    IncRef();
    return new(this);
}

[MustDisposeResource]
public override DreamValue OperatorRemove(DreamValue b) {
    if (b.TryGetValueAsIDreamList(out var bList)) {
        foreach (var value in bList.EnumerateValues())
            RemoveValue(value);
    } else {
        RemoveValue(b);
    }

    IncRef();
    return new(this);
}
```

*worked out with Claude Sonnet 4.6. I cannot write C#, forgiveness is begged for :p*

Laboredih123:
Implements +, +=, -, -=, |, |=, ~=, and &= for ALists.
needs tests, to be added
Fixes #2588

### DIFF
--- a/OpenDreamRuntime/Objects/Types/DreamAssocList.cs
+++ b/OpenDreamRuntime/Objects/Types/DreamAssocList.cs
@@ -149,6 +149,73 @@ public sealed class DreamAssocList(DreamObjectDefinition aListDef, int size) : D
         return _values.ContainsKey(value);
     }
 
+    public override DreamValue OperatorAdd(DreamValue b, DMProcState state) {
+        var listCopy = (DreamAssocList)CreateCopy();
+
+        if (b.TryGetValueAsIDreamList(out var bList)) {
+            foreach (var pair in bList.EnumerateAssocValues()) {
+                if (listCopy.ContainsKey(pair.Key)) {
+                    continue;
+                }
+
+                listCopy.SetValue(pair.Key, pair.Value);
+            }
+        } else {
+            listCopy.AddValue(b);
+        }
+
+        return new DreamValue(listCopy);
+    }
+
+    public override DreamValue OperatorSubtract(DreamValue b, DMProcState state) {
+        var listCopy = (DreamAssocList)CreateCopy();
+
+        if (b.TryGetValueAsIDreamList(out var bList)) {
+            foreach (DreamValue value in bList.EnumerateValues()) {
+                listCopy.RemoveValue(value);
+            }
+        } else {
+            listCopy.RemoveValue(b);
+        }
+
+        return new DreamValue(listCopy);
+    }
+
+    public override DreamValue OperatorOr(DreamValue b, DMProcState state) {
+        var listCopy = (DreamAssocList)CreateCopy();
+
+        if (b.TryGetValueAsIDreamList(out var bList)) {
+            foreach (var pair in bList.EnumerateAssocValues()) {
+                if (listCopy.ContainsKey(pair.Key)) {
+                    continue;
+                }
+
+                listCopy.SetValue(pair.Key, pair.Value);
+            }
+        } else {
+            listCopy.AddValue(b);
+        }
+
+        return new DreamValue(listCopy);
+    }
+
+    public override DreamValue OperatorCombine(DreamValue b) {
+        if (b.TryGetValueAsIDreamList(out var bList)) {
+            foreach (var pair in bList.EnumerateAssocValues()) {
+                if (ContainsKey(pair.Key)) {
+                    continue;
+                }
+
+                SetValue(pair.Key, pair.Value);
+            }
+        } else {
+            AddValue(b);
+        }
+
+        IncRef();
+        return new(this);
+    }
+
     public override DreamValue OperatorAppend(DreamValue b) {
         if (b.TryGetValueAsIDreamList(out var bList)) {
             foreach (var pair in bList.EnumerateAssocValues()) {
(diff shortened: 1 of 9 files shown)


OpenDreamRuntime/Objects/Types/DreamAssocList.cs
```diff
@@ -148,4 +148,142 @@ public void Swap(int index1, int index2) {
...
+                    continue;
+                }
+
+                listCopy.SetValue(pair.Key, pair.Value);
+            }
+        } else {
+            listCopy.AddValue(b);
+        }
+
+        return new DreamValue(listCopy);
```
wixoaGit: You could just replace this since they're the same thing.
```suggestion
        return OperatorAdd(b, state);
```

OpenDreamRuntime/Objects/Types/DreamAssocList.cs
```diff
@@ -148,4 +148,142 @@ public void Swap(int index1, int index2) {
...
+                }
+
+                SetValue(pair.Key, pair.Value);
+            }
+        } else {
+            AddValue(b);
+        }
+
+        IncRef();
+        return new(this);
```
wixoaGit:
```suggestion
        return OperatorAppend(b);
```

OpenDreamRuntime/Objects/Types/DreamAssocList.cs
```diff
@@ -148,4 +148,142 @@ public void Swap(int index1, int index2) {
...
+    }
+
+    public override DreamValue OperatorMask(DreamValue b) {
+        if (b.TryGetValueAsIDreamList(out var bList)) {
+            foreach (var value in EnumerateValues()) {
+                if (!bList.ContainsKey(value)) {
+                    RemoveValue(value);
```
wixoaGit: This is modification during enumeration, and will throw an exception.

OpenDreamRuntime/Objects/Types/DreamAssocList.cs
```diff
@@ -148,4 +148,142 @@ public void Swap(int index1, int index2) {
...
+                }
+            }
+        } else {
+            foreach (var value in EnumerateValues()) {
+                if (!value.Equals(b)) {
+                    RemoveValue(value);
```
wixoaGit: Modification during enumeration

wixoaGit (changes requested):

OpenDreamRuntime/Objects/Types/DreamAssocList.cs
```diff
@@ -148,4 +148,142 @@ public void Swap(int index1, int index2) {
...
+        }
+
+        IncRef();
+        return new(this);
+    }
+
+    public override DreamValue OperatorRemove(DreamValue b) {
+        if (b.TryGetValueAsIDreamList(out var bList)) {
+            foreach (var value in bList.EnumerateValues()) {
+                RemoveValue(value);
```
wixoaGit: This could also be a modification during enumeration if you do `alist -= alist`. This should just do a `Cut()` in that case.

OpenDreamRuntime/Objects/Types/DreamAssocList.cs
```diff
@@ -148,4 +148,132 @@
...
+            RemoveValue(b);
+        }
+
+        IncRef();
+        return new(this);
+    }
+
+    public override DreamValue OperatorMask(DreamValue b) {
+        if (Equals(b)) {
+            ;
```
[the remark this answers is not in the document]
tontyGH: Why not just wrap everything in an if(!Equals) block? It'd definitely read a lot better

OpenDreamRuntime/Objects/Types/DreamAssocList.cs
```diff
@@ -148,4 +148,132 @@ public void Swap(int index1, int index2) {
...
+            }
+        } else {
+            listCopy.AddValue(b);
+        }
+
+        return new DreamValue(listCopy);
+    }
+
+    public override DreamValue OperatorSubtract(DreamValue b, DMProcState state) {
+        if (Equals(b)) {
```
wixoaGit: `DreamObject.Equals()` doesn't do comparisons with `DreamValue` so this will always be false. You should do `bList == this` further down.

OpenDreamRuntime/Objects/Types/DreamAssocList.cs
```diff
@@ -148,4 +148,132 @@ public void Swap(int index1, int index2) {
...
+            return new DreamValue(new DreamAssocList(ObjectDefinition, null));
```
wixoaGit: You should do `ObjectTree.CreateAssocList()` instead of creating one directly.

OpenDreamRuntime/Objects/Types/DreamAssocList.cs
```diff
@@ -148,4 +148,132 @@ public void Swap(int index1, int index2) {
...
+        } else {
+            AddValue(b);
+        }
+
+        IncRef();
+        return new(this);
+    }
+
+    public override DreamValue OperatorRemove(DreamValue b) {
+        if (Equals(b)) {
```
wixoaGit: Always false

OpenDreamRuntime/Objects/Types/DreamAssocList.cs
```diff
@@ -148,4 +148,132 @@ public void Swap(int index1, int index2) {
...
+        } else {
+            RemoveValue(b);
+        }
+
+        IncRef();
+        return new(this);
+    }
+
+    public override DreamValue OperatorMask(DreamValue b) {
+        if (!Equals(b)) {
```
wixoaGit: `Equals()` is always false

wixoaGit (changes requested):

wixoaGit (approved):

PR #2582, opened 2026-05-21
Fixes num2text incorrectly interpreting 'a' as hex in r-10 case

tontyGH:
Previous coder accidentally did `<=` instead of `<` when checking for hex digits, which made the compiler interpret 'a' as valid hex when radix == 10.


wixoaGit:
Wouldn't hurt to [add a test](https://github.com/OpenDreamProject/OpenDream/blob/e12a5337c344721eeef364e17be6d4ae8b7ec94a/Content.Tests/DMProject/Tests/Builtins/text2num.dm) for this.

tontyGH:
Oh, true. Give me a minute.

wixoaGit (approved):

PR #2585, opened 2026-05-23
Implements DMProc.LocalCount, adds unit test for localvar allocs

### LINKED ISSUE #2583: Compiler fails to emit the correct local variable count
DMProc.LocalCount holds incorrect information about the variable count of a given proc, causing it to bloat considerably.

Image

This can be fatal in procs that use a lot of scoped variables, like in [BeeStation](https://github.com/BeeStation/BeeStation-Hornet/blob/abe2df36f54b46df278fe45136c9ceced2800006/code/modules/mob/living/living.dm#L1642-L1695)'s case where macros conceal a lot of them.

tontyGH:
LocalCount is meant to reflect the max amount of variables a proc COULD allocate during its execution, but it used to just naively reflect Proc.LocalNames.Count which doesn't do Any of That.

Fixes #2583

I had to move some stuff around in the compiler to expose the Json object to the unit test, but I triple checked and it should be equivalent function, and consumers can just ignore the version with output.


Content.Tests/DMProject/SpecialTests/Procs/local_count.dme
```diff
@@ -0,0 +1,42 @@
+#define TEST_PROC(name) /proc/Test___##name()
```
wixoaGit, line 3: Something to let people know where this test is used
```suggestion
// Used in the EnsureLocalCountIsValid test

#define TEST_PROC(name) /proc/Test___##name()
```

Content.Tests/RuntimeTests.cs
```diff
@@ -29,4 +34,32 @@ public void EnsurePragmaDefaults() {
...
+    [Test]
+    public void EnsureLocalCountIsValid() {
+        const string testPath = "Procs/local_count.dme";
+        const string testIdentifier = "Test___";
+
+        var compiler = new DMCompiler.DMCompiler();
+        compiler.Compile(new() {
+            Files = [Path.Join(Directory.GetCurrentDirectory(), SpecialTestsDirectory, testPath)]
+        }, out DreamCompiledJson? compiledDream);
+        Assert.That(compiledDream, Is.Not.Null, "Environment failed to compile");
```
wixoaGit: Use the boolean return of `compiler.Compile()` to determine success instead.

DMCompiler/DMCompiler.cs
```diff
@@ -44,7 +44,9 @@ public DMCompiler() {
         _errorConfig = new Dictionary<WarningCode, ErrorLevel>(CompilerEmission.DefaultErrorConfig);
     }
 
-    public bool Compile(DMCompilerSettings settings) {
+    public bool Compile(DMCompilerSettings settings) => Compile(settings, out _);
+
+    public bool Compile(DMCompilerSettings settings, out DreamCompiledJson? compiledDream) {
```
wixoaGit: Lets the compiler know `compiledDream` won't be null if this returns `true`. This will prevent warnings when it's used without a null check first.
```suggestion
    public bool Compile(DMCompilerSettings settings, [NotNullWhen(true)] out DreamCompiledJson? compiledDream) {
```

DMCompiler/DMCompiler.cs
wixoaGit, line 6: For the `NotNullWhen` attribute
```suggestion
using System.Diagnostics.CodeAnalysis;
using System.Globalization;
```

wixoaGit (changes requested):

wixoaGit (approved):

Issue #2580, opened 2026-05-20
Items sometimes disapear into backpack


mudzbe:
The item cant be clicked while inside the backpack, but emptying the whole backpack removes it

(sound removed because of music)

https://github.com/user-attachments/assets/34bc1758-b262-43ca-917a-3057d66e8997

PR #2586, opened 2026-05-23
Fixes some list wrappers not correctly incrementing refs

tontyGH:
fixes? https://github.com/OpenDreamProject/OpenDream/issues/2580

### DIFF
--- a/OpenDreamRuntime/Objects/Types/DreamList.cs
+++ b/OpenDreamRuntime/Objects/Types/DreamList.cs
@@ -568,7 +568,6 @@ internal sealed class DreamListVars(DreamObjectDefinition listDef, DreamObject d
             return new DreamValue(DreamObject.GetVariableNames().ElementAt(keyInteger - 1)); //1-indexed
         } else if (key.TryGetValueAsString(out var varName)) {
             if (DreamObject.TryGetVariable(varName, out var objectVar)) {
-                objectVar.IncRef();
                 return objectVar;
             }
 
@@ -1121,7 +1120,6 @@ public sealed class DreamFilterList(DreamObjectDefinition listDef, DreamObject o
 
         DreamObjectFilter filterObject = ObjectTree.CreateObject<DreamObjectFilter>(ObjectTree.Filter);
         filterObject.Filter = appearance.Filters[filterIndex - 1];
-        filterObject.IncRef();
         return new DreamValue(filterObject);
     }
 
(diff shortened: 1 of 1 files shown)


OpenDreamRuntime/Objects/Types/DreamList.cs
```diff
@@ -568,6 +568,7 @@ public override DreamValue GetValue(DreamValue key) {
             return new DreamValue(DreamObject.GetVariableNames().ElementAt(keyInteger - 1)); //1-indexed
         } else if (key.TryGetValueAsString(out var varName)) {
             if (DreamObject.TryGetVariable(varName, out var objectVar)) {
+                objectVar.IncRef();
```
wixoaGit: This actually shouldn't have its reference count incremented here. `TryGetVariable()` already does that, this is more of a reference transfer than a reference creation.
```suggestion
```

OpenDreamRuntime/Objects/Types/DreamList.cs
```diff
@@ -1116,9 +1119,9 @@ public override DreamValue GetValue(DreamValue key) {
         if (filterIndex < 1 || filterIndex > appearance.Filters.Length)
             throw new Exception($"Atom only has {appearance.Filters.Length} filter(s), cannot index {filterIndex}");
 
-        DreamFilter filter = appearance.Filters[filterIndex - 1];
         DreamObjectFilter filterObject = ObjectTree.CreateObject<DreamObjectFilter>(ObjectTree.Filter);
-        filterObject.Filter = filter;
+        filterObject.Filter = appearance.Filters[filterIndex - 1];
+        filterObject.IncRef();
```
wixoaGit: This also isn't necessary to do here. DreamObject starts off with a reference count of 1 because it assumes the code creating it is now holding a reference to it. This means `GetValue()` is transferring the reference here too.
```suggestion
```

OpenDreamRuntime/Objects/Types/DreamList.cs
```diff
@@ -1525,8 +1528,10 @@ public override DreamValue GetValue(DreamValue key) {
             childEnumerator.MoveNext(out EntityUid child);
 
             if (index == 1) {
-                if (AtomManager.TryGetMovableFromEntity(child, out var childObject))
+                if (AtomManager.TryGetMovableFromEntity(child, out var childObject)) {
+                    childObject.IncRef();
                     return new DreamValue(childObject);
+                }
                 else
```
```suggestion
                } else
```

wixoaGit (changes requested):

wixoaGit:
> fixes? https://github.com/OpenDreamProject/OpenDream/issues/2580

Probably not, I remember running into that issue before ref counting was added.

tontyGH:
Double checked the others

wixoaGit (approved):

PR #2587, opened 2026-05-24
Implements a handful of contents lists' ContainsValue methods

tontyGH:


OpenDreamRuntime/Objects/Types/DreamList.cs
```diff
@@ -1511,6 +1536,21 @@
...
+
+    public override bool ContainsValue(DreamValue value) {
+        DreamValue current = value;
+
+        while(current.TryGetValueAsDreamObject<DreamObjectAtom>(out var dreamAtom)) {
+            if(current.TryGetValueAsDreamObject<DreamObjectTurf>(out var dreamTurf)) {
+                return dreamTurf.Cell.Area == area;
+            }
+
+            dreamAtom.TryGetVariable("loc", out current);
```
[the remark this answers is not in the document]
tontyGH: I think that's fine?

OpenDreamRuntime/Objects/Types/DreamList.cs
```diff
@@ -1379,6 +1386,17 @@ public override int GetLength() {
...
     }
+
+    public override bool ContainsValue(DreamValue value) {
+        // world.contents is a list of every atom that exists,
+        // so this should be equivalent across the board
+        if(!value.TryGetValueAsDreamObject<DreamObjectAtom>(out var dreamObject))
+            return false;
+
+        if(dreamObject.Deleting || dreamObject.Deleted) // ??
+            return false;
```
wixoaGit: `TryGetValueAsDreamObject()` will return false and give you null if the object is deleted.

OpenDreamRuntime/Objects/Types/DreamList.cs
```diff
@@ -1511,6 +1536,21 @@ public override int GetLength() {
...
+    public override bool ContainsValue(DreamValue value) {
+        DreamValue current = value;
+
+        while(current.TryGetValueAsDreamObject<DreamObjectAtom>(out var dreamAtom)) {
+            if(current.TryGetValueAsDreamObject<DreamObjectTurf>(out var dreamTurf)) {
+                return dreamTurf.Cell.Area == area;
+            }
+
+            dreamAtom.TryGetVariable("loc", out current);
+            current.Dispose(); // TryGetVar auto-increments
```
wixoaGit: While its ref count wouldn't ever hit 0 here so it doesn't matter, this isn't correct since `current` is still holding a reference.

A better `ContainsValue()` implementation here would be to instead get the map cell at the atom's position. This avoids looping through `atom.loc` vars and needing to ref count.
```cs
if (!value.TryGetValueAsDreamObject<DreamObjectAtom>(out var atom))
    return false;

var pos = AtomManager.GetAtomPosition(atom);
if (!DreamMapManager.TryGetCellAt((pos.X, pos.Y), pos.Z, out var cell))
    return false;

return cell.Area == area;
```

wixoaGit (changes requested):

OpenDreamRuntime/Objects/Types/DreamList.cs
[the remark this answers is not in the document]
wixoaGit: Using `Dispose()` when you've got a DreamValue is better because it helps the analyzers catch mistakes.

I see how `DecRef()` also being public could be confusing. That should be made private / moved into `Dispose()`, but that's for another PR.

tontyGH:
I got baited...

wixoaGit:
> [I HONESTLY SHOULD BE ALLOWED TO DO THAT THOUGH](https://github.com/OpenDreamProject/OpenDream/pull/2587/commits/ededbc01b4e08279ef807546ecdc8a3630289b22)

You could, they just have to be lowercase if they're gonna be local variables.

tontyGH:
Oh.

OpenDreamRuntime/Objects/Types/DreamList.cs
```diff
@@ -1511,6 +1531,20 @@ public override int GetLength() {
     public override int FindValue(DreamValue value, int start = 1, int end = 0) {
         throw new NotImplementedException($".Find() is not yet implemented on {GetType()}");
     }
+
+    public override bool ContainsValue(DreamValue value) {
+        if (!value.TryGetValueAsDreamObject<DreamObjectAtom>(out var atom))
+            return false;
+
+        if (value.TryGetValueAsDreamObject<DreamObjectArea>(out _)) // areas do not contain themselves
```
wixoaGit: More direct
```suggestion
        if (atom is DreamObjectArea) // areas do not contain themselves
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #2590, opened 2026-05-25
Implement Remaining /icon.Blend() functions

VMSolidus:
As a follow up to #141 

/icon.Blend() now supports ICON_AND and ICON_OR.


wixoaGit:
Have you tested that this is how `ICON_AND` and `ICON_OR` work in BYOND? The `&` isn't correct according to the DM ref's example.

amylizzle:
I made a test #2591

wixoaGit (approved):

PR #2597, opened 2026-05-26
Remove a few unnecessary `List<>` creations

ike709:
As far as I could tell there was no reason to be converting this array to a list?

### DIFF
--- a/DMCompiler/Bytecode/OpcodeMetadata.cs
+++ b/DMCompiler/Bytecode/OpcodeMetadata.cs
@@ -33,8 +33,7 @@ public static class OpcodeMetadataCache {
     private static readonly OpcodeMetadata[] MetadataCache = new OpcodeMetadata[256];
 
     static OpcodeMetadataCache() {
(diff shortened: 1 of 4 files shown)


DMCompiler/Bytecode/OpcodeMetadata.cs
```diff
@@ -33,7 +33,8 @@ public static class OpcodeMetadataCache {
     private static readonly OpcodeMetadata[] MetadataCache = new OpcodeMetadata[256];
 
     static OpcodeMetadataCache() {
-        foreach (DreamProcOpcode opcode in Enum.GetValues(typeof(DreamProcOpcode))) {
+        var allOpcodes = Enum.GetValues(typeof(DreamProcOpcode));
+        foreach (DreamProcOpcode opcode in allOpcodes) {
```
ike709:
```suggestion
        foreach (DreamProcOpcode opcode in Enum.GetValues(typeof(DreamProcOpcode))) {
```

wixoaGit (approved):

PR #2598, opened 2026-05-27
Makes Client/Mob Login/out behavior match BYOND way more closely

tontyGH:
This is how I atone for making client.mob regress slightly.

### DIFF
--- a/DMCompiler/DMStandard/Types/Atoms/Mob.dm
+++ b/DMCompiler/DMStandard/Types/Atoms/Mob.dm
@@ -16,5 +16,10 @@
 	layer = MOB_LAYER
 
 	proc/Login()
+		if(!loc)
+			var/turf/fallback_spawn = locate(1, 1, 1) // TODO: Find nearest non-dense turf
+			if(fallback_spawn)
+				loc = fallback_spawn
 
 	proc/Logout()
+		del client
--- a/OpenDreamRuntime/DreamConnection.cs
+++ b/OpenDreamRuntime/DreamConnection.cs
@@ -45,7 +45,7 @@ public sealed partial class DreamConnection {
             SetClientMob(value);
 
             if(oldConnection is not null) {
-                oldConnection.Client?.Delete();
+                oldConnection.Client?.Delete(); // TODO: This should tell you why you disconnected
                 _mob!.SpawnProc("Logout").Dispose();
             }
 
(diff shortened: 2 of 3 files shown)


DMCompiler/DMStandard/Types/Atoms/Mob.dm
```diff
@@ -16,5 +16,10 @@
 	layer = MOB_LAYER
 
 	proc/Login()
+		if(!loc)
```
wixoaGit, line 19: So this happens again in BYOND if you re-log while your mob is in nullspace?

DMCompiler/DMStandard/Types/Atoms/Mob.dm
```diff
@@ -16,5 +16,10 @@
...
+			var/turf/fallback_spawn = locate(1, 1, 1) // TODO: Find nearest non-dense turf
+			if(fallback_spawn)
+				loc = fallback_spawn
 
 	proc/Logout()
+		del client
```
wixoaGit, line 25: This is already forced by the engine itself in `DreamConnection.HandleDisconnection()`.

OpenDreamRuntime/DreamConnection.cs
```diff
@@ -37,32 +37,20 @@ public sealed partial class DreamConnection {
...
+            var oldMob = _mob;
+            var oldConnection = value?.Connection;
+            SetClientMob(value);
 
-                    _mob.Connection = this;
-                    _mob.Key = Key;
-                    _mob.SpawnProc("Login", usr: _mob).Dispose();
-                }
+            if(oldConnection is not null) {
+                oldConnection.HandleDisconnection();
```
wixoaGit: `HandleDisconnection()` is for handling the player disconnecting from the server. This has the side effect of deleting the `/client` as well as preventing future network communication.

wixoaGit (changes requested):

OpenDreamRuntime/DreamConnection.cs
tontyGH: Yes, BYOND behaviour for having your mob stolen is being disconnected from the server. This is intentional.

Would I need to implement my own disconnect method or is there something else I should be using? I looked around but nothing else seemed appropriate.

DMCompiler/DMStandard/Types/Atoms/Mob.dm
tontyGH: This is for the case where mob.Logout() is called directly, BYOND base behaviour is to disconnect the client. I think we talked about this before.

DMCompiler/DMStandard/Types/Atoms/Mob.dm
tontyGH: That is correct.

OpenDreamRuntime/DreamConnection.cs
wixoaGit: ~~You'd want to use `Session.Channel.Disconnect()` instead of leaving the player in that limbo state~~ deleting the `/client` does this anyway but that's still a little indirect. Maybe delete the client object instead.

I'm curious if BYOND still disconnects you if you gain another mob before the tick ends?

tontyGH: Done.
> I'm curious if BYOND still disconnects you if you gain another mob before the tick ends?

It does. BYOND disconnects the client the moment the mob is overriden, even before the first Logout is invoked.

OpenDreamRuntime/DreamConnection.cs
```diff
@@ -125,21 +113,43 @@ public void HandleDisconnection() {
...
-            // Don't null out the ckey here
-            _mob.SpawnProc("Logout").Dispose();
-
-            if (_mob != null) { // Logout() may have removed our mob
-                _mob.Connection = null;
-                _mob = null;
-            }
-        }
-
+        Mob = null;
```
tontyGH: Clients still hold a reference to their mob on Del(), which is only wiped after they get deleted. This breaks relogging outright.

wixoaGit (approved):

PR #2599, opened 2026-05-27
Fix warning on DV overlay

amylizzle:
protomanager.Index doesn't like string literals

### DIFF
--- a/OpenDreamClient/Rendering/DreamViewOverlay.cs
+++ b/OpenDreamClient/Rendering/DreamViewOverlay.cs
@@ -80,6 +80,19 @@ internal sealed partial class DreamViewOverlay : Overlay {
         M22 = -1
     };
 
+    //Used for supressing the "No Literals" warning on shader index lookups
+    private static class ODShaderID
+    {
+        public const string
+            BlockColor = "blockcolor",
+            Color = "color",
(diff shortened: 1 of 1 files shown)


ike709:
Robust Toolbox's more "correct" way would be for this to be a `ProtoId<OdShaderId>` iirc, but I'm also not sure how much we care especially given that this isn't a `[Datafield]` and also we don't have YAML to validate.

amylizzle:
I suppose we could add it as a prototype but honestly it is not worth the effort of figuring out how to do that and make it work with the filter system, for the sake of just avoiding a potential typo on an ID that's used once in one file.

OpenDreamClient/Rendering/DreamViewOverlay.cs
```diff
@@ -80,6 +80,19 @@ internal sealed partial class DreamViewOverlay : Overlay {
         M22 = -1
     };
 
+    //Used for supressing the "No Literals" warning on shader index lookups
+    private static class OdShaderId
+    {
```
wixoaGit:
```suggestion
    private static class OdShaderId {
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #2603, opened 2026-05-28
Recoup of #2517 "Implement `caller`"

tontyGH:
I implemented the RefType for callees because locating a stale callee with the debugger would crash the debugger.


tontyGH:
Okay, I think we're good.

wixoaGit:
> locating a stale callee with the debugger would crash the debugger.

The debugger should be able to handle a variable throwing an exception if you try to read it. That can be another PR though, if you want to open an issue for it.

OpenDreamRuntime/Objects/Types/DreamObjectCallee.cs
```diff
@@ -7,35 +8,67 @@ public sealed class DreamObjectCallee(DreamObjectDefinition objectDefinition) :
...
                 value = new(new ProcArgsList(ObjectTree.List.ObjectDefinition, ProcState));
                 return true;
             case "caller":
-                // TODO
-                value = DreamValue.Null;
+                if(ProcState.Caller is DMProcState callerState) {
+                    callerState.CalleeObject.IncRef();
+                    value = new(callerState.CalleeObject);
+                }
+                else if(ProcState.Caller?.Caller is DMProcState superCallerState) { // init case
```
wixoaGit: Shouldn't this loop up `ProcState.Caller` until it reaches a `DMProcState`, rather than only checking a depth of 2?

OpenDreamRuntime/Procs/DMProc.cs
```diff
@@ -370,13 +370,25 @@ public sealed class DMProcState : ProcState {
     public readonly IDreamValueEnumerator?[] Enumerators = new IDreamValueEnumerator?[16];
 
     public int ProgramCounter => _pc;
-    public override DMProc Proc => _proc;
+    public override DMProc? Proc => (DMProc?)_proc; // spawn() procs can make this null for a sec, don't worry about it
```
wixoaGit: How so? `_proc` is set fairly early in the constructor that spawn() uses.

OpenDreamRuntime/Procs/DMProc.cs
```diff
@@ -370,13 +370,25 @@ public sealed class DMProcState : ProcState {
...
+
+    public DreamObjectCallee CalleeObject { get {
+            if(Thread is null)
+                throw new InvalidOperationException("Attempted to get the callee of a disposed proc");
+            if(_callee is null) {
+                _callee = DreamObjectCallee.FromDMProcState(this);
+                _callee.IncRef(); // the state's ref to the callee
```
wixoaGit: `DreamObjectCallee.FromDMProcState()` is already returning the object with 1 ref count.
```suggestion
```

OpenDreamRuntime/Procs/DMProc.cs
```diff
@@ -932,16 +948,20 @@ public DreamValue GetReferenceValue(DreamReference reference, bool peek = false)
...
+                // BYOND seems to reuse the same object. At least, callee == callee
+                CalleeObject.IncRef();
+                return new(CalleeObject);
             }
             case DMReference.Type.Caller: {
-                // TODO
-                return DreamValue.Null;
+                // Note that the ref says that caller still returns a "/callee" object, just with the caller's info
+                if(Caller is not DMProcState dmCaller) {
+                    if(Caller?.Caller is not DMProcState realDmCaller) // init case
```
wixoaGit: Shouldn't this also loop up `Caller?.Caller`

wixoaGit (changes requested):

OpenDreamRuntime/Objects/Types/DreamObjectCallee.cs
tontyGH: From what I understand, I assumed the only case where this would be relevant would be if we check Caller from inside a new() call, which we'd only need to check a depth of 2 to handle. Making it recursive would just be for future-proofing.

OpenDreamRuntime/Procs/DMProc.cs
tontyGH: I'm confused on why this is no longer an issue, but it is no longer an issue. Fixed.

OpenDreamRuntime/Procs/DMProc.cs
tontyGH: Resolving this since the conversation is happening in the other comment.

OpenDreamRuntime/Objects/Types/DreamObjectCallee.cs
wixoaGit: Future-proofs it, as well as just makes the code easier to read/follow.
```cs
case "caller":
    var caller = ProcState.Caller;
    while (caller is not (DMProcState or null))
        caller = caller.Caller;

    value = caller is DMProcState dmCaller ? new DreamValue(dmCaller.CalleeObject) : DreamValue.Null;
    value.IncRef();
    return true;
```
```cs
case DMReference.Type.Caller: {
    // Note that the ref says that caller still returns a "/callee" object, just with the caller's info
    var caller = Caller;
    while (caller is not (DMProcState or null))
        caller = caller.Caller;

    var value = caller is DMProcState dmCaller ? new(dmCaller.CalleeObject) : DreamValue.Null;
    value.IncRef();
    return value;
}
```

The test should probably also include this behavior.

wixoaGit (approved):

PR #2604, opened 2026-05-28
Re-implement range() and orange() to match BYOND

### LINKED ISSUE #2584: Any proc that uses default view range throws exception if world.view is not 1:1
ViewRange uses a default value of 0 as its range if it isn't square-sized or centerable, which completely breaks default behaviour.
Image
BYOND behavior is to match world.view exactly when doing range checks.

### LINKED ISSUE #2601: range() does not include found areas
Snippet: 
```dm
/world
  maxx = 7
  maxy = 7
  maxz = 1


/proc/main()
  var/turf/center = locate(7, 7, 1)
  for(var/atom/A as anything in range(0, center))
    world.log << "([A.x], [A.y]) [A.type]"
```

BYOND (via Wrench):

Image

OpenDream (via Sleepyhead):

Image

tontyGH:
Fixes #2601 
Fixes #2584

Re-implements the ViewSpiral algorithm to handle rectangular cases. It's definitely not optimized at all and I'm probably the last person you should ask to write optimized code.

Unless there's some niche interaction I'm unaware of, this implementation should match BYOND 1-to-1
I may have overengineered the integration test


ZeWaka:
orange!!!!! 🍊

PowerfulBacon:
Nice, this is probably one of the biggest things I see erroring out when I try and run SS13 codebases locally, since this gets hit every time you click anything.

wixoaGit (approved):

PR #2611, opened 2026-06-04
Fixes DreamListFilters not incrementing refs in GetValue

tontyGH:


tontyGH:
Whoops.

OpenDreamRuntime/Objects/Types/DreamList.cs
```diff
@@ -1144,6 +1144,7 @@ public override DreamValue GetValue(DreamValue key) {
 
         DreamObjectFilter filterObject = ObjectTree.CreateObject<DreamObjectFilter>(ObjectTree.Filter);
         filterObject.Filter = appearance.Filters[filterIndex - 1];
+        filterObject.IncRef();
```
wixoaGit: I don't think this is the problem either. `filterObject` starts off with a ref count of 1, since it was just created. Ownership of that reference immediately gets transferred to the code calling `GetValue()`.

tontyGH: Oh. Huh...

PR #2617, opened 2026-06-05
Coloured console errors

PowerfulBacon:
Creates a new helper function on emission which prints to the console, applying foreground colour for the warning/error/notice part of the message.

Replaces console writes of emissions with this new method.

While console colours aren't going to appear in most contexts (mainly, if you are running from the VS code extension), they will appear when you are running from the command line.

image

### DIFF
--- a/DMCompiler/Compiler/CompilerError.cs
+++ b/DMCompiler/Compiler/CompilerError.cs
@@ -224,6 +224,7 @@ public struct CompilerEmission {
             ErrorLevel.Error => ConsoleColor.Red,
             _ => null
         };
+        
         if (usedColor is null) {
             return;
         }
@@ -233,5 +234,4 @@ public struct CompilerEmission {
         Console.ResetColor();
         Console.WriteLine($" at {Location.ToString()}: {Message}");
     }
-
 }
--- a/DMCompiler/Compiler/CompilerError.cs
+++ b/DMCompiler/Compiler/CompilerError.cs
@@ -227,6 +227,7 @@ public struct CompilerEmission {
         if (usedColor is null) {
             return;
         }
+        
         Console.ForegroundColor = usedColor.Value;
         Console.Write($"{Level.ToString()} OD{(int)Code:d4}");
         Console.ResetColor();
--- a/DMCompiler/Compiler/CompilerError.cs
+++ b/DMCompiler/Compiler/CompilerError.cs
@@ -213,4 +213,24 @@ public struct CompilerEmission {
         ErrorLevel.Error => $"Error OD{(int)Code:d4} at {Location.ToString()}: {Message}",
         _ => ""
     };
+
+    /// <summary>
+    /// Write to the console with coloured formatting.
+    /// </summary>
+    public void WriteConsole() {
+        ConsoleColor? usedColor = Level switch {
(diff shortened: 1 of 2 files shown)


DMCompiler/Compiler/CompilerError.cs
```diff
@@ -213,4 +213,24 @@ public CompilerEmission(ErrorLevel level, WarningCode code, Location? location,
...
+    public void WriteConsole() {
+        ConsoleColor? usedColor = Level switch {
+            ErrorLevel.Notice => ConsoleColor.Cyan,
+            ErrorLevel.Warning => ConsoleColor.Yellow,
+            ErrorLevel.Error => ConsoleColor.Red,
+            _ => null
+        };
+        if (usedColor is null) {
+            return;
+        }
```
PowerfulBacon, line 230:
```suggestion
        }
        
```

wixoaGit:
The VS Code extension [uses a regex](https://github.com/OpenDreamProject/vscode-opendream/blob/e62bcb8defb6fff274a4682662d347bfc7669b96/package.json#L60) to parse the errors from the compiler's output. I wonder if this would break that?

wixoaGit:
> The VS Code extension [uses a regex](https://github.com/OpenDreamProject/vscode-opendream/blob/e62bcb8defb6fff274a4682662d347bfc7669b96/package.json#L60) to parse the errors from the compiler's output. I wonder if this would break that?

Tested it, it does not. TG's lint annotator has this in it, so I figure it doesn't break that either:
```python
def annotate(raw_output):
    # Remove ANSI escape codes
    raw_output = re.sub(r'(\x9B|\x1B\[)[0-?]*[ -\/]*[@-~]', '', raw_output)
```

DMCompiler/Compiler/CompilerError.cs
```diff
@@ -213,4 +213,25 @@ public CompilerEmission(ErrorLevel level, WarningCode code, Location? location,
...
+        
+        Console.ForegroundColor = usedColor.Value;
+        Console.Write($"{Level.ToString()} OD{(int)Code:d4}");
+        Console.ResetColor();
+        Console.WriteLine($" at {Location.ToString()}: {Message}");
+    }
+
```
wixoaGit:
```suggestion
```

DMCompiler/Compiler/CompilerError.cs
```diff
@@ -213,4 +213,25 @@ public CompilerEmission(ErrorLevel level, WarningCode code, Location? location,
...
+    /// Write to the console with coloured formatting.
+    /// </summary>
+    public void WriteConsole() {
+        ConsoleColor? usedColor = Level switch {
+            ErrorLevel.Notice => ConsoleColor.Cyan,
+            ErrorLevel.Warning => ConsoleColor.Yellow,
+            ErrorLevel.Error => ConsoleColor.Red,
+            _ => null
+        };
+        if (usedColor is null) {
```
wixoaGit, line 228:
```suggestion
        };
        
        if (usedColor is null) {
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #2618, opened 2026-06-05
Supports the titlebar DMF attribute

PowerfulBacon:
This PR implements the ability to create a borderless window in the Clyde window engine (Requires https://github.com/space-wizards/RobustToolbox/pull/6618), and implements the titlebar property, which hides the title bars when set to 0.

Because most TGUI windows set this property through browse(), I have also implemented support for passing options through to browse as right now only the size option is being handled.

These TGUI windows cannot yet be moved, as there is no support for moving windows from code in the Clyde engine, and I suspect that there may be some trouble getting that in if there are security concerns about having arbitrarily moving windows.

image

image

### DIFF
--- a/OpenDreamRuntime/DreamConnection.cs
+++ b/OpenDreamRuntime/DreamConnection.cs
@@ -1,6 +1,3 @@
-using System.Threading.Tasks;
-using System.Web;
-using DMCompiler.Bytecode;
 using OpenDreamRuntime.Objects;
 using OpenDreamRuntime.Objects.Types;
 using OpenDreamRuntime.Procs.Native;
@@ -11,6 +8,8 @@ using OpenDreamShared.Network.Messages;
 using Robust.Shared.Enums;
 using Robust.Shared.Player;
 using SpaceWizards.Sodium;
+using System.Threading.Tasks;
+using System.Web;
 
 namespace OpenDreamRuntime;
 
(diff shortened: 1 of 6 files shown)


OpenDreamRuntime/DreamConnection.cs
```diff
@@ -462,7 +464,8 @@ public void HandleBrowseResourceRequest(string filename) {
 
     public void Browse(string? body, string? options) {
         string? window = null;
-        Vector2i size = (480, 480);
+
+        List<string> unhandledOptions = new List<string>();
```
PowerfulBacon, line 468: Now that we no longer are using winset, it doesn't matter if we sent the window param across so we can just send the raw options string and not worry about populating/filtering this list

PowerfulBacon:
I could probably split this up so that the titlebar support which depends on a robust toolbox PR can come later.

ike709:
> These TGUI windows cannot yet be moved, as there is no support for moving windows from code in the Clyde engine, and I suspect that there may be some trouble getting that in if there are security concerns about having arbitrarily moving windows.

When things calm down we can shoot PJB a message about the security concerns.

based PR, the titlebars have been annoying me for a while

PowerfulBacon:
Im worried that there will be concerns about the ability to set the position of windows and remove the borders which will be needed for fancy TGUI. Im hoping that since they can be closed if misbehaving by just closing all from the task bar then it won't be a huge issue.

For now ill atomize out the browse() handling options and then deal with things needing RT changes when the PRs can be merged

amylizzle:
is this cross platform? can you test in on windows, wayland and X?

OpenDreamRuntime/DreamConnection.cs
PowerfulBacon: I actually think this is better practice

PowerfulBacon:
I believe it will be cross-platform since its a standard SDL call, but I'm trying to find a good way of testing this

PowerfulBacon:
I'll let the robust toolbox PR progress before re-opening this

PR #2619, opened 2026-06-07
Fixes HtmlParser handling of XML comments and values containing spaces

PowerfulBacon:
The HTML parser currently expects closing tags to match opening tags. This is a problem with comment nodes, as it pushes `!--` to the stack, and expects `!--/>`, while the closing tag for comments is actually `-->`.

When a comment tag is encountered, the reader will totally ignore it and it won't be pushed to the formatted message at all. By default, the reader will read the entire comment tag `<!-- this text will be included -->`, however if the comment tag contains the `>` symbol, it won't be fully captured so I added `SkipComment` to handle that scenario.

The result is that the warnings in the console no longer occur when parsing HTML with comments.

image

image

image

(The empty nodes are new-line tags, which should probably be limited to a single one without <br /> tags, but that's a different issue).

Also handling the situation that might break it nicely:

image

image

Since the chat uses a browser window anyway, this has no visible effect on the chat but it stops an error in the console and this might become more important for maptext parsing, which is where the custom HTML parsing will have the most impact I think.

### DIFF
--- a/OpenDreamClient/Interface/Html/HtmlParser.cs
+++ b/OpenDreamClient/Interface/Html/HtmlParser.cs
@@ -1,6 +1,8 @@
 ﻿using OpenDreamShared.Dream;
 using Robust.Shared.Utility;
+using System.Linq;
 using System.Text;
+using System.Text.RegularExpressions;
 
 namespace OpenDreamClient.Interface.Html;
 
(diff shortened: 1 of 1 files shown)


PowerfulBacon:
Found another bug here, the string `img class=' icon icon-floor' src='asset.e58c1d7b547c4efe06f91a20a3f7266f.png'` is incorrectly handled

image

Edit: Fixed

Note that attributes not wrapped in " or ' are not valid in HTML, but will be matched by this anyway (example `foo=5`)

image

image

image

OpenDreamClient/Interface/Html/HtmlParser.cs
```diff
@@ -11,6 +13,9 @@ public static class HtmlParser {
     private static readonly ISawmill Sawmill;
     private static readonly HashSet<string> WarnedAttributes = new();
 
+    private static Regex? _attributeMatchRegex = null;
```
PowerfulBacon:
```suggestion
    private static Regex? _attributeMatchRegex;
```

wixoaGit (approved):

PR #2620, opened 2026-06-07
Support options in browse

PowerfulBacon:
Implements support for options being passed through to browse.

The scope of this change is currently quite limited and won't have any practical effect because there is little support for most of the window descriptors, but this will be necessary for TGUI to work properly (as options are passed through in browse instead of through winset).

I tested this by modifying the options of TGUI to pass through the title of 'hello world', as title is a supported parameter:

image

image

This does not work with the same code on master:

image

Prior to this change, the only supported option was `size`.

Support for explicitly declared file in browse has also been added.

### DIFF
--- a/OpenDreamClient/Interface/DreamInterfaceManager.cs
+++ b/OpenDreamClient/Interface/DreamInterfaceManager.cs
@@ -206,11 +206,14 @@ internal sealed partial class DreamInterfaceManager : IDreamInterfaceManager {
         } else if (pBrowse.HtmlSource != null) {
             Dictionary<string, string>? options = null;
 
-            if (pBrowse.Options is not null
-                && ParseDmfParams(pBrowse.Options, out var checkParserErrors) is DMFParser parser
-                && parser.AttributesValues() is Dictionary<string, string> attributes
-                && !checkParserErrors()) {
-                options = attributes;
+            if (pBrowse.Options is not null) {
+                DMFParser? parser = ParseDmfParams(pBrowse.Options, out var checkParserErrors);
+                if (parser is not null) {
+                    Dictionary<string, string> attributes = parser.AttributesValues();
+                    if (!checkParserErrors()) {
+                        options = attributes;
+                    }
+                }
             }
 
             string htmlFileName;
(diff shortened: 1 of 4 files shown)


OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -203,7 +204,25 @@ private void RxBrowse(MsgBrowse pBrowse) {
...
+                    Dictionary<string, string> attributes = parser.AttributesValues();
+                    if (!checkParserErrors()) {
+                        options = attributes;
+                    }
+                }
+            }
+
+            string htmlFileName;
+            if (options?.TryGetValue("file", out var fileName) ?? false) {
+                htmlFileName = fileName;
```
PowerfulBacon, line 221: This needs to remove the extension from the filename, as we add .html ourselves.

OpenDreamRuntime/DreamConnection.cs
```diff
@@ -158,7 +159,8 @@ public void UpdateStat() {
                 }
 
                 return DreamValue.Null;
-            } finally {
+            }
+            finally {
```
wixoaGit:
```suggestion
            } finally {
```

OpenDreamClient/Interface/DreamInterfaceManager.cs
```diff
@@ -203,7 +204,25 @@ private void RxBrowse(MsgBrowse pBrowse) {
...
         } else if (pBrowse.HtmlSource != null) {
-            var htmlFileName = $"browse_{pBrowse.Window}_{_random.Next()}"; // TODO: Possible collisions and explicit file names
+            Dictionary<string, string>? options = null;
+
+            if (pBrowse.Options is not null) {
+                DMFParser? parser = ParseDmfParams(pBrowse.Options, out var checkParserErrors);
+                if (parser is not null) {
+                    Dictionary<string, string> attributes = parser.AttributesValues();
+                    if (!checkParserErrors()) {
+                        options = attributes;
```
wixoaGit, line 214: The DM ref says `You may use commas (,), ampersands (&), or semicolons (;) as the delimiter` which is probably worth addressing.

The DM ref also only mentions the ability to modify a few of these, so those should be whitelisted. Some of the properties are named slightly differently as well (`-` versus `_`).

wixoaGit (changes requested):

OpenDreamClient/Interface/DreamInterfaceManager.cs
wixoaGit: You'd use `Path.GetFileNameWithoutExtension()` for that, assuming the sandbox is happy with it.

PR #2621, opened 2026-06-07
Autolint

amylizzle:
commenting `!autolint` on PR should automatically apply the linter suggestions

### DIFF
--- a/.github/workflows/autolint.yml
+++ b/.github/workflows/autolint.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Setup Resharper
         run: dotnet tool install -g JetBrains.ReSharper.GlobalTools
       - name: Run Linter
-        run: jb cleanupcode ${git diff --diff-filter=ACMR --name-only origin/master...HEAD}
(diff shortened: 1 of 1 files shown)


amylizzle:
!autolint

amylizzle:
yeah I didn't think it'd work till it was merged into master, but eh...

PowerfulBacon:
Does this require that you are the owner of the PR/a maintainer?

amylizzle:
it does not, but it probably should

amylizzle:
works (see final attempt): https://github.com/amylizzle/OpenDream/pull/14

doesn't quite do what I wanted tho, corrected a lot of formatting issues but not the reported unused vars

ike709:
based PR

And yeah I agree with only PR author and maintainers being allowed to use it.

wixoaGit:
Is this much use as-is? It doesn't look like it actually takes care of anything that ReSharper points out. It just reformats whitespace/line length, some of which I think looks worse.

amylizzle:
yeah it's not really the intended effect I was going for. It's supposed to apply the project settings for resharper but I'm not sure it's actually doing that properly.

PR #2622, opened 2026-06-07
Make error recovery a toggleable error code

### LINKED ISSUE #2616: Tests should fail if "Error recovery had to skip" lines
It's just a warning by default, it should probably be an error - at least on tests, but probably globally

amylizzle:
Also removes a bad test which does not compile on BYOND or OD, it just hits our error recovery

Closes #2616


wixoaGit:
Not really sure how I feel about this. This warning was added really early on, when the compiler tripped on all sorts of things. Its existence was more to give an idea of how much of the code was being skipped because of parsing issues.

If something is causing this message, but there's no error for it, then that's a bug.

amylizzle:
I've only seen it on bad indent, but also at this point it should definitely be a "failed to compile" imo.

wixoaGit:
I suppose. Really this message shouldn't exist though, and the goal should be for anything that triggers error recovery to actually produce an error.

wixoaGit (approved):

PR #2629, opened 2026-06-17
Fixes misc Move() bugs, adds integration test for Move chain

tontyGH:


DMCompiler/DMStandard/_Standard.dm
```diff
@@ -167,7 +167,19 @@ proc/winset(player, control_id, params)
...
+		Dir &= ~(EAST | WEST)
+	if((Dir & (UP | DOWN)) == (UP | DOWN))
+		Dir &= ~(UP | DOWN)
+	if(!Dir)
+		return 0
+
+	var/atom/step = get_step(Ref, Dir)
+	if(!step || step == get_step(Ref, 0))
+		return 0
+	return Ref.Move(step, Dir)
```
wixoaGit: `Dir` gets passed to `Move()` unchanged:
Image

I think this is closer to BYOND behavior:
```suggestion
    var/NewLoc = get_step(Ref, Dir)
    if (!NewLoc || NewLoc == loc)
        return 0
        
	return Ref.Move(NewLoc, Dir)
```

wixoaGit (changes requested):

DMCompiler/DMStandard/_Standard.dm
wixoaGit: Probably worth testing how this behaves when the movable is inside another movable instead of a turf

DMCompiler/DMStandard/Types/Atoms/Movable.dm
```diff
@@ -23,39 +23,55 @@
 
 		if (Dir != 0)
 			dir = Dir
+		
+		var/area/oldarea = isarea(loc?.loc) ? loc.loc : null
+		var/area/newarea = isarea(NewLoc.loc) ? NewLoc.loc : null
```
wixoaGit, line 28: Wouldn't you want this to work for a depth greater than 1? `get_step(loc, 0)?.loc` could get you the area instead.

DMCompiler/DMStandard/Types/Atoms/Movable.dm
```diff
@@ -23,39 +23,55 @@
...
+				if (exiting != src)
+					if (!exiting.Uncross(src))
+						return FALSE
 
-		if (NewLoc.Enter(src, loc))
+		var/loc_entered = NewLoc.Enter(src, loc)
+		var/area_entered = !consider_area || isnull(newarea) || newarea.Enter(src, loc)
+		if (!loc_entered || !area_entered)
+			if (!area_entered)
+				src.Bump(newarea)
```
wixoaGit, line 48: You can bump into an area?

wixoaGit (changes requested):

DMCompiler/DMStandard/Types/Atoms/Movable.dm
tontyGH: If you fail to enter it, yeah.

DMCompiler/DMStandard/Types/Atoms/Movable.dm
tontyGH: No, cause the only time the area is relevant is if the old/new loc is a turf. Otherwise, the area check is skipped entirely.

Here's a case where a nullspace mob is moved into another mob, the area is ignored: 
image

And the inverse case of leaving an area into another mob is in the unit test.

wixoaGit (approved):

PR #2631, opened 2026-06-19
Makes replacetextEx and replacetext share a common method

tontyGH:
Should be more maintainable like this, having them split up created a bug that was fixed in one proc but unfixed in the other.


OpenDreamRuntime/Procs/Native/DreamProcNativeHelpers.cs
```diff
@@ -316,6 +316,63 @@ public static DreamValue HandleOviewersOhearers(NativeProc.Bundle bundle, DreamO
         return new DreamValue(view);
     }
 
+    public static DreamValue HandleReplaceText(DreamValue haystackValue, DreamValue needleValue, DreamValue replacementValue, int start, int end, bool exact) {
+        // TODO: _char support if/when we support char
```
tontyGH:
```suggestion
        // TODO: byte support if/when we support working with bytes
```

wixoaGit (approved):

PR #2638, opened 2026-06-23
Fix `<init>` proc for type trees compiled out of order

### LINKED ISSUE #2630: Type trees compiled out of order can have bogus <init> procs
Broken code block in question: 
```dm
/obj
  var/list/type_list = null

/obj/subtype/broken
  type_list = null

/obj/subtype
  type_list = list(/datum, /atom, /obj,)

/proc/main()
  var/obj/O = new /obj/subtype/broken()
  world.log << isnull(O.type_list)
```

When `/obj/subtype/broken` has to build its init vars, it correctly identifies that `type_list` is a null constant and checks the parent to see if it should still override its <init> proc. However, because it's being built **before** its parent type was ready, it will see that `/obj/subtype` has no init vars and simply assume that it doesn't need to make one.

ike709:
Fixes https://github.com/OpenDreamProject/OpenDream/issues/2630

This just effectively defers the `IsRuntimeInitialized()` check on the first pass if it's an override. Doesn't seem to horribly break anything?

### DIFF
--- a/DMCompiler/DM/DMCodeTree.Vars.cs
+++ b/DMCompiler/DM/DMCodeTree.Vars.cs
@@ -30,7 +30,7 @@ internal partial class DMCodeTree {
             return true;
         }
 
-        protected void SetVariableValue(DMCompiler compiler, DMObject dmObject, DMVariable variable, DMExpression value, bool isOverride) {
+        protected bool TrySetVariableValue(DMCompiler compiler, DMObject dmObject, DMVariable variable, DMExpression value, bool isOverride) {
             // Typechecking
             if (!variable.ValType.MatchesType(compiler, value.ValType) && !variable.ValType.IsUnimplemented && !variable.ValType.Type.HasFlag(DMValueType.NoConstFold)) {
                 if (value is Null && !isOverride) {
@@ -44,17 +44,20 @@ internal partial class DMCodeTree {
             if (value.TryAsConstant(compiler, out var constant)) {
                 variable.Value = constant;
 
+                // Overrides shouldn't succeed while the parent init proc is undefined
+                if (isOverride && dmObject.Parent?.InitializationProc == null && IsFirstPass) return false;
+
                 // We want to continue with putting this in the init proc if a base type initializes it to another value
                 if (!isOverride || !dmObject.IsRuntimeInitialized(variable.Name)) {
-                    return;
+                    return true;
                 }
             } else if (variable.IsConst) {
                 compiler.Emit(WarningCode.HardConstContext, value.Location, "Value of const var must be a constant");
-                return;
+                return false;
             } else if (!IsValidRightHandSide(compiler, dmObject, value)) {
                 compiler.Emit(WarningCode.BadExpression, value.Location,
                     $"Invalid initial value for \"{variable.Name}\"");
-                return;
+                return false;
             }
 
             var initLoc = value.Location;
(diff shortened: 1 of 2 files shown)


DMCompiler/DM/DMCodeTree.Vars.cs
```diff
@@ -95,7 +99,7 @@ private bool IsValidRightHandSide(DMCompiler compiler, DMObject dmObject, DMExpr
     }
 
     private class ObjectVarNode(DreamPath owner, DMASTObjectVarDefinition varDef) : VarNode {
-        private string VarName => varDef.Name;
+        public string VarName => varDef.Name;
```
ike709:
```suggestion
        private string VarName => varDef.Name;
```

DMCompiler/DM/DMCodeTree.Vars.cs
```diff
@@ -44,17 +44,20 @@ protected void SetVariableValue(DMCompiler compiler, DMObject dmObject, DMVariab
             if (value.TryAsConstant(compiler, out var constant)) {
                 variable.Value = constant;
 
+                // Overrides that are defined out of order need an opportunity for the parent to perform runtime initializations
+                if (isOverride && !dmObject.IsRuntimeInitialized(variable.Name) && IsFirstPass) return false;
```
ike709:
```suggestion
                if (isOverride && IsFirstPass && !dmObject.IsRuntimeInitialized(variable.Name)) return false;
```

DMCompiler/DM/DMCodeTree.Vars.cs
```diff
@@ -44,17 +44,20 @@ protected void SetVariableValue(DMCompiler compiler, DMObject dmObject, DMVariab
...
+
                 // We want to continue with putting this in the init proc if a base type initializes it to another value
                 if (!isOverride || !dmObject.IsRuntimeInitialized(variable.Name)) {
-                    return;
+                    return true;
                 }
             } else if (variable.IsConst) {
                 compiler.Emit(WarningCode.HardConstContext, value.Location, "Value of const var must be a constant");
-                return;
+                return false;
```
wixoaGit: Returning false from TryDefineVar() will lead to the code tree trying again to define it in another pass. The variable isn't going to become non-const, so there's no point in that.

DMCompiler/DM/DMCodeTree.Vars.cs
```diff
@@ -44,17 +44,20 @@ protected void SetVariableValue(DMCompiler compiler, DMObject dmObject, DMVariab
...
             } else if (!IsValidRightHandSide(compiler, dmObject, value)) {
                 compiler.Emit(WarningCode.BadExpression, value.Location,
                     $"Invalid initial value for \"{variable.Name}\"");
-                return;
+                return false;
```
wixoaGit: Similar here

wixoaGit (changes requested):

DMCompiler/DM/DMCodeTree.Vars.cs
ike709:
```suggestion
                return true;
```

DMCompiler/DM/DMCodeTree.Vars.cs
ike709:
```suggestion
                return true;
```

wixoaGit (approved):

PR #2647, opened 2026-06-25
Fixes /atom/var/color having an incorrect initial value

tontyGH:
I feel like we should have some kind of system to automatically insert appearance vars into types so we only need to do this in one place

### DIFF
--- a/DMCompiler/DMStandard/Types/Atoms/_Atom.dm
+++ b/DMCompiler/DMStandard/Types/Atoms/_Atom.dm
@@ -32,7 +32,7 @@
 	var/layer = 2.0
 	var/plane = 0
 	var/alpha = 255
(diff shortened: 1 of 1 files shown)


DMCompiler/DMStandard/Types/Atoms/_Atom.dm
```diff
@@ -32,7 +32,7 @@
 	var/layer = 2.0
 	var/plane = 0
 	var/alpha = 255
-	var/color = "#FFFFFF"
+	var/color = null as text|null
```
wixoaGit: This can also be a list
```suggestion
	var/color = null as /list|text|null
```

wixoaGit (approved):

PR #2651, opened 2026-06-29
ProcDecoder Refactor Redux

ike709:
This PR has a few main changes:
- `ProcDecoder` was moved from the runtime to the compiler.
- As a side effect of that, `DMDisassembler` now depends on the compiler instead of the runtime.
- The `ITuple` is now a `ProcInstruction` comprised of `ProcOperand`s that store the argument values.
- `Format()` and `DecodeInstruction()` now utilize opcode metadata to figure out how to handle each opcode rather than using user-defined `switch()` cases for each opcode.

Most of that handling diverges depending on whether the opcode has a fixed-length argument count or a variable-length argument count (`N`-variant opcodes). There's also a fair bit of code repetition for the tradeoff of not having to do a ton of typechecking on what types the arguments actually are.

Tested with DM disassembly of /tg/station before and after this PR. Didn't see any non-whitespace diffs.

See my first comment below as well.


DMCompiler/Bytecode/ProcInstruction.cs
```diff
@@ -0,0 +1,272 @@
...
+using System.Text;
+using DMCompiler.DM;
+
+// ReSharper disable UnusedMember.Global
+
+namespace DMCompiler.Bytecode;
+
+/// <summary>
+/// A typed bytecode operand. Array operands are used only by compact variable-argument opcodes
+/// TODO: When .NET 11 is out, revisit this with C# 15 unions and OneOrMany
```
ike709, line 12: This is a really important TODO that will be very useful for these changes, but I'm not going to delay the PR by several months to wait for it.

I will open an issue for this when the PR gets merged.

DMCompiler/Bytecode/ProcInstruction.cs
```diff
@@ -0,0 +1,272 @@
...
+                {
+                    text.Append(' ');
+                    AppendFloat(text, t);
+                }
+
+                break;
+            }
+            case ProcOperandShape.RepeatedString: {
+                string[] values = operand0.Strings!;
+                for (int i = 0; i < values.Length; i++) {
```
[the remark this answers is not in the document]
ike709: Do we care?

DMCompiler/Bytecode/ProcDecoder.cs
```diff
@@ -0,0 +1,234 @@
...
+    public IEnumerable<(int Offset, ProcInstruction Instruction)> Disassemble() {
+        while (Remaining) {
+            yield return (Offset, DecodeInstruction());
+        }
+    }
+
+    /// <summary>
+    /// Stringifies a <see cref="ProcInstruction"/> and its operands
+    /// </summary>
+    public static string Format(ProcInstruction instruction, Func<int, string> getTypePath) {
```
wixoaGit, line 92: Can't the caller call `Format()` on the instruction directly?

DMCompiler/Bytecode/ProcDecoder.cs
```diff
@@ -0,0 +1,234 @@
...
+        return instruction.Format(getTypePath);
+    }
+
+    public static int? GetJumpDestination(ProcInstruction instruction) {
```
wixoaGit, line 96: This too?

DMCompiler/Bytecode/ProcInstruction.cs
```diff
@@ -0,0 +1,272 @@
...
+public readonly struct ProcInstruction(
+    int offset,
+    int endOffset,
+    DreamProcOpcode opcode,
+    int operandCount = 0,
+    ProcOperandShape operandShape = ProcOperandShape.Fixed,
+    ProcOperand operand0 = default,
+    ProcOperand operand1 = default,
+    ProcOperand operand2 = default,
+    ProcOperand operand3 = default) {
```
wixoaGit, line 78: Could this be an array instead? I don't like how you have to write out behavior for every possible argument length.

DMCompiler/Bytecode/OpcodeVerifier.cs
```diff
@@ -17,9 +17,23 @@ public static string GetOpcodesHash() {
             byte[] nameBytes = Encoding.ASCII.GetBytes(value.ToString()!);
             opcodesBytes.AddRange(nameBytes);
             opcodesBytes.Add((byte)value);
+
+            var metadata = OpcodeMetadataCache.GetMetadata((DreamProcOpcode)value);
+            opcodesBytes.AddRange(BitConverter.GetBytes(metadata.StackDelta));
+            opcodesBytes.Add((byte)(metadata.VariableArgs ? 1 : 0));
+            AddArgTypes(opcodesBytes, metadata.RequiredArgs);
+            AddArgTypes(opcodesBytes, metadata.RepeatedArgs);
```
wixoaGit, line 25: I would rather override `OpcodeMetadata.GetHashCode()` and use that here, so that it's clearer that this needs updated if the metadata's fields change.

DMCompiler/Bytecode/ProcInstruction.cs
```diff
@@ -0,0 +1,272 @@
+using System.Globalization;
+using System.Runtime.InteropServices;
+using System.Text;
+using DMCompiler.DM;
+
+// ReSharper disable UnusedMember.Global
```
wixoaGit, line 6: Is this doing anything? Removing it isn't showing me any new warnings.

DMCompiler/Bytecode/ProcInstruction.cs
```diff
@@ -0,0 +1,272 @@
...
+
+        return labelIndex == -1 ? null : GetInt(labelIndex);
+    }
+
+    private void AppendRepeatedOperands(StringBuilder text) {
+        switch (operandShape) {
+            case ProcOperandShape.RepeatedFloat: {
+                float[] values = operand0.Floats!;
+                foreach (var t in values)
+                {
```
wixoaGit, line 156:
```suggestion
                foreach (var t in values) {
```

DMCompiler/Bytecode/ProcInstruction.cs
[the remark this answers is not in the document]
wixoaGit: I don't see why you wouldn't use a `foreach` here, though I'd be fine with disabling the inspection if others don't like it.

wixoaGit (changes requested):

PR #2652, opened 2026-06-30
Fixes findtext() negative search bug

### LINKED ISSUE #2634: `findtext()` throws an exception if negative search is longer than haystack
Example code:
```dm
world.log << findtext("pipe gun", " (Wielded)", -10)
```

ozgurerturk:
Fixes #2634.

Prevents an error when an overly negative input start value is given for findtext() proc. 

Added a regression test for the reported case.


wixoaGit (changes requested):
This should be done to `findtextEx()` as well. Looks like it's missing the `start < 0` behavior entirely.

ozgurerturk:
I’ve added the same negative Start handling to findtextEx().
I noticed that findtext() also supports negative End values, while findtextEx() currently returns 0 for them.
Should I preserve the End value check behaviour or should negative End handling be made consistent with findtext() as well?

wixoaGit:
Yea, looks like it's missing that too:
```dm
world.log << findtextEx("pipe gun (Wielded)", "ed", -100)
world.log << findtextEx("pipe gun (Wielded)", "ed", -100, -1)
world.log << findtextEx("pipe gun (Wielded)", "ed", -100, -2)
```
Should output:
```
16
16
0
```

ozgurerturk:
I've applied the changes, also updated test findtext.dm

wixoaGit (approved):

PR #2653, opened 2026-07-01
Fixes void expressions in ternary not giving a compile-time error

### LINKED ISSUE #2643: `()` should be a compile failure as of 516.1683
`Tests/Tests/Operators/void_in_ternary.dm` should be a compile failure

note that `as()` is still supported

ozgurerturk:
() in ternary now gives compile error.

Also updated void ternary test.

Fixes #2643


wixoaGit:
This was the only reason DMASTVoid existed. That class can be removed now, and [this](https://github.com/OpenDreamProject/OpenDream/blob/0c62aeae5bab6eb49345b0e343e8809291849672/DMCompiler/Compiler/DM/DMParser.cs#L2249) replaced with [an error like this](https://github.com/OpenDreamProject/OpenDream/blob/0c62aeae5bab6eb49345b0e343e8809291849672/DMCompiler/DM/Builders/DMExpressionBuilder.cs#L478).

wixoaGit (approved):

PR #2655, opened 2026-07-06
Fixes browse() pages allowing illegal characters in filenames

tontyGH:
Required for Baystation12 to function on OD


wixoaGit:
I think a better solution would be to decode the URL in the browser's request handler instead. These methods aren't necessarily just for the browser files.

wixoaGit (approved):

PR #2657, opened 2026-07-06
Refactors particles to allow multiple owners

tontyGH:
This is probably extremely incorrect but that's what the PR process is for, right?

Made a ParticleData object to minimize juggling values between components


OpenDreamRuntime/Objects/Types/DreamObjectParticles.cs
```diff
@@ -30,31 +21,72 @@ public DreamObjectParticles(DreamObjectDefinition objectDefinition) : base(objec
...
+        }
+    }
+
     protected override void SetVar(string varName, DreamValue value) {
         //good news, these only update on assignment, so we don't need to track the generator, list, or matrix objects
         switch (varName) {
             case "width": //num
-                _particlesComponent.Width = (int)value.UnsafeGetValueAsFloat();
+                _particlesData.Width = (int)value.UnsafeGetValueAsFloat();
+                MarkDirty();
```
wixoaGit: I'd rather just have one `MarkDirty()` before the `base.SetVar()` at the end of this method. Non-`/particles` vars probably aren't being set often enough for it to be a problem.

wixoaGit (commented):
> This is probably extremely incorrect but that's what the PR process is for, right?

It's a reasonable approach, works fine.

OpenDreamRuntime/Objects/Types/DreamObjectParticles.cs
```diff
@@ -30,69 +21,103 @@ public DreamObjectParticles(DreamObjectDefinition objectDefinition) : base(objec
...
+            using var boundYValue = vectorList.GetValue(new(2));
+            using var boundZValue = vectorList.GetValue(new(3));
+            var boundX = boundXValue.UnsafeGetValueAsFloat();
+            var boundY = boundYValue.UnsafeGetValueAsFloat();
+            var boundZ = boundZValue.UnsafeGetValueAsFloat();
+
+            newVec = new Vector3(boundX, boundY, boundZ);
+            return true;
+        }
+        else if (value.TryGetValueAsDreamObject<DreamObjectVector>(out var vectorVector /*lol*/)) {
```
wixoaGit:
```suggestion
        } else if (value.TryGetValueAsDreamObject<DreamObjectVector>(out var vectorVector /*lol*/)) {
```

OpenDreamShared/Rendering/DreamParticlesComponent.cs
```diff
@@ -1,47 +1,54 @@
...
 using Robust.Shared.Analyzers;
 using Robust.Shared.GameObjects;
 using Robust.Shared.GameStates;
 using Robust.Shared.Maths;
+using Robust.Shared.Serialization;
 using Robust.Shared.ViewVariables;
 
 namespace OpenDreamShared.Rendering;
 
 [RegisterComponent, NetworkedComponent, AutoGenerateComponentState(true, fieldDeltas: true)]
```
wixoaGit:
```suggestion
[RegisterComponent, NetworkedComponent, AutoGenerateComponentState(true)]
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #2661, opened 2026-07-09
Fixes static indexing prioritizing globals

tontyGH:


tontyGH:
I have no idea where I'm supposed to put this test.

wixoaGit (approved):

PR #2662, opened 2026-07-11
Implements all* of the icon operations

tontyGH:
*except for Crop


Content.Tests/DMProject/Tests/Icon/_helpers.dm
```diff
@@ -0,0 +1,11 @@
+// IGNORE
+
+/proc/CompareIcons(icon/generated, icon/expected)
+	for(var/y in 1 to expected.Height())
+		for(var/x in 1 to expected.Width())
```
wixoaGit, line 8: Should check `generated.Height() == expected.Height()` as well. Same with width.

OpenDreamShared/Dream/ColorHelpers.cs
```diff
@@ -1,15 +1,19 @@
...
 public static class ColorHelpers {
+    public const string Transparent = "transparent";
+
     public static readonly Dictionary<string, Color> Colors = new() {
         {"black", new Color(00, 00, 00)},
         {"silver", new Color(192, 192, 192)},
         {"gray", new Color(128, 128, 128)},
         {"grey", new Color(128, 128, 128)},
         {"white", new Color(255, 255, 255)},
+        {Transparent, default}, // NOT equivalent to Color.Transparent, this matters!
```
wixoaGit, line 16: This doesn't show in the DM Ref's [html colors page](https://www.byond.com/docs/ref/#/{{appendix}}/html-colors). Does it work if you do `atom.color = "transparent"`?

OpenDreamRuntime/Procs/NativeProc.cs
```diff
@@ -52,6 +52,7 @@ public readonly ref struct Bundle(NativeProc proc, DreamProcArguments arguments)
         public readonly NativeProc Proc = proc;
 
         // NOTE: Deliberately not using DreamProcArguments here, tis slow.
+        // FIXME: Null arguments at the end of the list should be discarded
```
wixoaGit: Why? That doesn't match with how I've seen BYOND handle proc arguments. Some native procs may be different though, due to them being opcodes that can have special behavior in BYOND.
Image

tontyGH: I actually didn't know that, I made that comment since MapColors seems to internally trim null args when considering its form and I just figured it did that for all procs

OpenDreamShared/Dream/ColorHelpers.cs
tontyGH: It does, and it correctly makes the atom transparent. Not sure if this is replicated in OD, though.

wixoaGit: This line would be how you add that to OD.

wixoaGit (approved):

PR #2666, opened 2026-07-18
Adds `PreIncrement` & `PreDecrement` opcodes; fixes comparison operators

ike709:
The new unit tests passed in BYOND but not OpenDream. 

This fixes `++`/`--` the parity issue by creating dedicated `PreIncrement`/`PreDecrement` opcodes with matching behavior. Deduplicated the opcode handlers with a shared helper method.

This also fixes the greater than / less than comparison operators (and the "or equals to" variants) with some types.

### DIFF
--- a/OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
+++ b/OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
@@ -1488,6 +1488,11 @@ namespace OpenDreamRuntime.Procs {
             using var second = state.Pop();
             using var first = state.Pop();
 
+            if (TryGetReferenceComparisonResult(first, second, out DreamValue referenceResult)) {
+                state.Push(referenceResult);
+                return ProcStatus.Continue;
+            }
+
             state.Push(new DreamValue(IsGreaterThan(first, second) ? 1 : 0));
             return ProcStatus.Continue;
         }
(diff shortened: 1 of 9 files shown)


OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
```diff
@@ -2963,6 +2983,26 @@ private static bool IsEquivalent(DreamValue first, DreamValue second) {
...
+            result = default;
+            return false;
+        }
+
+        private static bool IsReferenceComparisonType(DreamValue.DreamValueType type) {
+            return type is DreamValue.DreamValueType.DreamObject or
+                DreamValue.DreamValueType.DreamProc or
+                DreamValue.DreamValueType.DreamResource or
+                DreamValue.DreamValueType.DreamType;
+        }
```
wixoaGit: I don't like the maintainability issue this creates. BYOND doesn't make this sort of "reference type" distinction, would it be more accurate to say this applies to anything that isn't a number, null, or string?

wixoaGit (changes requested):

OpenDreamRuntime/Procs/DMOpcodeHandlers.cs
ike709: Probably? I'd need to find time this weekend for further testing.

wixoaGit (approved):

PR #2668, opened 2026-07-19
Replaces IDreamList with an abstract BaseDreamList type

tontyGH:
BaseDreamList is a bad name, but I'm not really sure what else to call DreamList.

I plan on moving the list classes into their own subfolder, but this would have to touch the entire codebase (to fix usings) so it should probably be its own PR

ClientImagesList and DreamAssocList are migrated to this


tontyGH:
Not sure if I'm going in the right direction.

tontyGH:
I wasn't. I'll revisit this someday.

PR #2670, opened 2026-07-23
Less Strict InvalidVarType

VMSolidus:
InvalidVarType is essentially useless as a top level error otherwise, since it's not possible to static type icons. I would have expected it to be more useful if I was permitted to opt-in to it by declaring types for vars rather than globally checking every dynamic typed var. 

As an aside, it's seemingly not possible for me to correct icons not being statically typed anyways.

image

image

### DIFF
--- a/DMCompiler/DM/DMCodeTree.Vars.cs
+++ b/DMCompiler/DM/DMCodeTree.Vars.cs
@@ -36,7 +36,7 @@ internal partial class DMCodeTree {
                 if (value is Null && !isOverride) {
                     compiler.Emit(WarningCode.ImplicitNullType, value.Location, $"{dmObject.Path}.{variable.Name}: Variable is null but not explicitly typed as nullable, append \"|null\" to \"as\". Implicitly treating as nullable.");
                     variable.ValType |= DMValueType.Null;
-                } else if (!variable.ValType.IsAnything) {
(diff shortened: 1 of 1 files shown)


ike709:
Is this not effectively what `--skip-anything-typecheck` already does?

ike709:
Yeah after looking at the code this should instead just be checking for `SkipAnythingTypecheck` and that it's `anything`.

DMCompiler/DM/DMCodeTree.Vars.cs
```diff
@@ -36,7 +36,7 @@ protected bool TrySetVariableValue(DMCompiler compiler, DMObject dmObject, DMVar
                 if (value is Null && !isOverride) {
                     compiler.Emit(WarningCode.ImplicitNullType, value.Location, $"{dmObject.Path}.{variable.Name}: Variable is null but not explicitly typed as nullable, append \"|null\" to \"as\". Implicitly treating as nullable.");
                     variable.ValType |= DMValueType.Null;
-                } else {
+                } else if (compiler.Settings.SkipAnythingTypecheck && !variable.ValType.IsAnything || !compiler.Settings.SkipAnythingTypecheck) {
```
wixoaGit:
```suggestion
                } else if (!compiler.Settings.SkipAnythingTypecheck || !variable.ValType.IsAnything) {
```

wixoaGit (approved):

PR #2669, opened 2026-07-20
Misc Parity Fixes

ike709:
- Fixes misc builtins not properly handling some bad args
- Fixes incorrect list size handling, particularly for multidimensional lists
- Fixes some bitwise operator behavior on lists

All of the new unit tests were passing in BYOND but not in OpenDream prior to these changes.

PR #2671, opened 2026-07-23
Fix `list() | list()` not including the first list's associated values

wixoaGit:
`json_encode(list("a"=1, "b"=2) | list("c"=3, "d"=4))`
was giving
`{"a":null,"b":null,"c":3,"d":4}`


wixoaGit:
#2669 has a more correct fix

PR #2677, opened 2026-07-24
Fix unimplemented `==` operations

### LINKED ISSUE #2675: Unimplemented == comparison between numbers and resources
Aurora station was erroring on a comparison between 0 and `'sound/weapons/metalhit.ogg'`.

mumencoder:
Fixes #2675

## Problem
Comparing a number with a resource (e.g. `0 == 'sound/weapons/metalhit.ogg'`)
threw `NotImplementedException: Equal comparison for 0 and '...' is not implemented`
at runtime instead of evaluating to `FALSE`.

## Fix
In DM, values of different types are never `==`-equal, so each inner switch now
falls back to `return false` (mirroring the existing `DreamResource`/`Appearance`
branches) instead of throwing. This covers the whole class of cross-type
comparisons, and `!=` / `~=` inherit the fix since they route through `IsEqual`.

### DIFF
--- a/Content.Tests/DMProject/Tests/Operators/equality_disparate_types.dm
+++ b/Content.Tests/DMProject/Tests/Operators/equality_disparate_types.dm
@@ -1,0 +1,27 @@
+// Comparing values of different types with == / != should never error;
+// disparate types are simply never equal.
+// Regression test for https://github.com/OpenDreamProject/OpenDream/issues/2675
+
+/datum/equality_test
+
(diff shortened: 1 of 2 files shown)


wixoaGit:
> In DM, values of different types are never ==-equal

If this is true then you could replace all these switches with an `if (first.Type != second.Type) return false;` early on in the method.

wixoaGit (approved):

PR #2678, opened 2026-07-24
Ignore empty path elements, like BYOND does

### LINKED ISSUE #2649: Preprocessor Parity Error w/ Slashes in Macros
This goonstation code:
```dm
/// Radio colours, CSS classes, and frequencies.
CREATE_NAMESPACE(RADIO)


//------------ Colours ------------//
/// Radio colours.
CREATE_NAMESPACE(RADIO, COL)

ADD_TO_NAMESPACE(RADIO, COL)(var/const/BRIG = "#FF5000")
```
using these cursed macros: https://github.com/goonstation/goonstation/blob/5ad07ab1ca1d05d1c382a27fbb8c91d28c02d4de/_std/namespaces/_namespace.dm#L31

gets preprocessed into:
```dm
var/datum/namespace/RADIO/RADIO = /datum/namespace/RADIO
/datum/namespace/RADIO/var//datum/namespace/RADIO__COL/COL = /datum/namespace/RADIO__COL
/datum/namespace/RADIO__COL/var/const/BRIG = "#FF5000"
```
(note the `var//`)

which OpenDream then errors on. BYOND does not error.

Goonstation fixed this by removing a slash from the start of the path in one of the macros here: https://github.com/goonstation/goonstation/commit/9da86136cdd1158a8ea8c2bffc93b20af58e068a

I don't know off the top of my head when BYOND will/won't tolerate `var//path/example` so this will probably need further testing?

mumencoder:
A macro that expands to a path can be pasted directly after a slash, producing an empty path element. BYOND ignores those, we errored out. This is only reachable through macro expansion, because a literal // in source is a comment.


mumencoder:
fix for #2649

wixoaGit (approved):

Issue #2646, opened 2026-06-25
`/generator + generator` and other operations are unimplemented


tontyGH:
Reference page for generator operations: https://ref.dm-lang.org/notes/generators

This is necessary for Baystation12 to work on OpenDream, since they use this RNG for planet effects in some cases.

PR #2680, opened 2026-07-25
Basic implementation of generators for issue #2646

mumencoder:


wixoaGit:
We already have generators. The issue is about math operators on generators.

wixoaGit:
Oh I see, this did just do the operators

PR #2683, opened 2026-07-27
`image()` requires at least 1 arg

ike709:
BYOND errors on zero-arg `image()`. I checked to see if there was an upper limit past the ref's 5 args but it took 8 args just fine.

### DIFF
--- a/DMCompiler/Compiler/DM/DMParser.cs
+++ b/DMCompiler/Compiler/DM/DMParser.cs
@@ -2755,6 +2755,14 @@ namespace DMCompiler.Compiler.DM {
 
                         return new DMASTRgb(identifier.Location, callParameters);
                     }
+                    case "image": {
+                        if (callParameters.Length < 1)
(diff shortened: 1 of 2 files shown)


DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -2755,6 +2755,14 @@ private void BracketWhitespace() {
 
                         return new DMASTRgb(identifier.Location, callParameters);
                     }
+                    case "image": {
+                        if (callParameters.Length < 1)
+                            Emit(WarningCode.InvalidArgumentCount, callLoc,
+                                "Expected at least 1 argument for image()");
+
+                        return new DMASTProcCall(callLoc,
+                            new DMASTCallableProcIdentifier(callLoc, identifier.Identifier), callParameters);
```
wixoaGit:
```suggestion
                        goto default;
```

wixoaGit (changes requested):

wixoaGit (approved):

PR #2688, opened 2026-07-29
Implements vis_contents on images

tontyGH:
I think. Required for holomaps to work on Baystation.

### DIFF
--- a/OpenDreamRuntime/Objects/Types/DreamList.cs
+++ b/OpenDreamRuntime/Objects/Types/DreamList.cs
@@ -978,18 +978,18 @@ public sealed class DreamOverlaysList(DreamObjectDefinition listDef, DreamObject
 }
 
 // atom.vis_contents list
-// Operates on an atom's appearance
+// Operates on an object's appearance
 public sealed class DreamVisContentsList : DreamList {
     private readonly PvsOverrideSystem? _pvsOverrideSystem;
 
     private readonly List<DreamObjectAtom> _visContents = new();
-    private readonly DreamObject _atom;
+    private readonly DreamObject _owner; // atom or image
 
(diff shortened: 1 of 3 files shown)


OpenDreamRuntime/Objects/Types/DreamObjectImage.cs
```diff
@@ -203,6 +210,22 @@ protected override void SetVar(string varName, DreamValue value) {
...
             }
+            case "vis_contents": {
+                value.TryGetValueAsDreamList(out var valueList);
+
+                _visContents.Cut();
+
+                if (valueList != null) {
+                    // TODO: This should postpone UpdateAppearance until after everything is added
+                    foreach (DreamValue underlayValue in valueList.EnumerateValues()) {
+                        _underlays.AddValue(underlayValue);
```
wixoaGit: This is adding to `_underlays` and not `_visContents`

wixoaGit (changes requested):
The comments and `_atom` field on `DreamVisContentsList` should be updated to reflect that it operates on more than just atoms now.

OpenDreamRuntime/Objects/Types/DreamObjectImage.cs
tontyGH: Whoops.

OpenDreamRuntime/Objects/Types/DreamList.cs
```diff
@@ -978,18 +978,18 @@ private ImmutableAppearance[] GetOverlaysArray(ImmutableAppearance appearance) =
 }
 
 // atom.vis_contents list
```
```suggestion
// atom or image vis_contents list
```

wixoaGit (approved):

Issue #2523, opened 2026-03-02
Alpha is ignored on plane masters that are render relay'd


LemonInTheDark:
If you change the alpha of a plane master which is being render relayed somewhere else using "*" format, the alpha modification is ignored and it continues to render at full opacity. 

I'd expect the alpha modification to carry down, but it doesn't.

LemonInTheDark:
adding a line to plane rendering that mimics how normal textures use alpha seemingly resolves this fine, but it does reintroduce #2548

Issue #2509, opened 2026-01-28
Flashlight creates a shadow/reduces light


mudzbe:
On TG SS14 server

Flashlight creates a shadow/reduces light
<img width="191" height="211" alt="Image" src="https://github.com/user-attachments/assets/5ea38156-4a73-48ee-9869-9d953c0911bb" />

ZeWaka:
bad blend mode?

wixoaGit:
Observable on our test server's current commit, https://github.com/ike709/tgstation/commit/15f1d42d8c30dd101f53664644a1830f0eefab71

PR #2689, opened 2026-07-29
Appearance color and alpha is determined solely by ColorMatrix

### LINKED ISSUE #2548: Color Matrix Alpha is not handled correctly
In BYOND, the raw alpha channel in color matrixes (a -> a) is treated as partially linked to the alpha var. Updating the aa channel will mirror alpha to match, and updating alpha will set the aa channel to 1.

In opendream they appear to be treated as two different, separate things. In addition, it seems as if the aa channel straight up does not DO anything. Other components of the color matrix are allowed to muck with the alpha of the world, but aa is seemingly totally ignored.

tontyGH:
Fixes #2548 
Fixes? #2523 
Fixes? #2509

Should make it much easier to keep color/alpha state consistent.

ColorMatrix was cleaned up a bit so that it's easier to read.


tontyGH:
Jesus.

wixoaGit:
Flashlights no longer show at all on /tg/station
screenshot

PR #2690, opened 2026-07-29
Fix disassembling lateral proc overrides

### LINKED ISSUE #2562: `DMDisassembler` does not handle lateral proc overrides
Example: https://github.com/tgstation/tgstation/blob/e86689907d5296020b396ea46e721acffe8ed17d/code/game/objects/structures/guillotine.dm

`/obj/structure/guillotine/attackby()` is defined twice, but the disassembler only decompiles the last definition.

After I noticed the issue, wixoa pointed out the problematic code:

Image

```
if (Json.Procs != null) {
    foreach (List<int> procIds in Json.Procs) {
        DMProc proc = Program.Procs[procIds[^1]];

        Procs.Add(proc.Name, proc);
    }
}
```

snowyukitty:
## Summary

- Preserve every lateral proc definition in compiler-emitted source order instead of keeping only the final definition.
- Decompile all matching definitions together and print a notice explaining the multi-definition output.
- Add a compiler-to-CLI regression test and run the dedicated test project in CI.

Fixes #2562

## Testing

- `dotnet build DMDisassembler/DMDisassembler.csproj --configuration Release --no-restore`
- `dotnet test DMDisassembler.Tests/DMDisassembler.Tests.csproj --configuration Release --no-build --no-restore`
- `dotnet test DMDisassembler.Tests/DMDisassembler.Tests.csproj --configuration Release --no-build --no-restore --filter FullyQualifiedName~DMDisassembler.Tests.LateralProcTests` — 1/1 passed.
- `jb inspectcode OpenDream.slnx -o="output.json" --project="DMDisassembler.Tests" --no-swea` with JetBrains InspectCode 2026.2.0.2 — 0 inspection results.
- Compiled a three-definition reproduction and confirmed the CLI prints the notice followed by definitions returning 1, 2, and 3 in source order.


DMDisassembler/Program.cs
```diff
@@ -368,8 +368,14 @@ private static void Decompile(string[] args) {
             } else {
                 Console.WriteLine("Selected type does not have an init proc");
             }
-        } else if (_selectedType.Procs.TryGetValue(name, out DMProc? proc)) {
-            Console.WriteLine(proc.Decompile());
+        } else if (_selectedType.Procs.TryGetValue(name, out DMProc[]? procs)) {
+            if (procs.Length > 1) {
+                Console.WriteLine($"Notice: Found {procs.Length} definitions of {name}(); decompiling all in source order.");
```
ike709, line 373: Personally, I'd like a second message afterwards too since I may not always scroll up past the text wall to see this one. Just a `Finished decompiling all {proc.Length} definitions of {name}()` or similar.

snowyukitty:
Added a completion message after the last definition and extended the regression test to verify it appears after the final decompiled body. Thanks!

DMDisassembler/Program.cs
snowyukitty: Implemented in e1d9dfd: the disassembler now prints `Finished decompiling all {procs.Length} definitions of {name}().` after the loop. I also reran the exact-head `LateralProcTests` regression test locally; it passes and verifies that the message appears once after the last definition.

wixoaGit (approved):

PR #2693, opened 2026-08-09
Update RobustToolbox to version 288.1.0

### LINKED ISSUE #2571: Closing browser windows with the close button shuts down the client
Closing with the tgui close button is fine, but using the system window's close button shuts down the client.

Noticed on goon (closing the start popups) on ubuntu

### LINKED ISSUE #2650: Tracy can be updated to 0.13.1
The Tracy-CSharp package we use [has been updated to 0.13.1](https://github.com/clibequilibrium/Tracy-CSharp/pull/66). There _probably_ haven't been large enough changes to the API to make this a difficult upgrade, though it would need tested.

DamianX:
The non-trivial change is the refactoring of DreamFilter that previously used serialization features that were changed in RT 287. I opted to write something similar to what particles do, instead of trying to hold on to ISerializationManager, which seemed inefficient and round-about either way.

The lifetimes and handling of DreamFilters in general seems highly suspicious to me but my refactoring should have kept it as suspicious as it was before, I believe.

Closes #2650 as I confirmed Tracy Just Works with the new version RT uses, though I'll admit I had never used Tracy before this so something may have slipped past me.

Closes #2571


DamianX:
The latest RT versions provide an updated SDL3 so #2571 is also fixed

wixoaGit (approved):

PR #2695, opened 2026-08-12
Add shared DMException for DM runtime errors

### LINKED ISSUE #2685: A new DMException error type
It would be nice to have a separate exception type for errors originating from bad DM, rather than an unexpected C# exception. These errors could also exclude the C# stack trace if the server isn't compiled with DEBUG.

ahmdkaml:
## Context

While investigating the use of generic `Exception` in the DM runtime, I found that this is substantially larger than a single-file change.

`throw new Exception(...)` is used across OpenDream for several different categories of failures: DM-facing runtime errors, runtime/VM invariants, resource handling, compiler/bytecode validation, and other internal errors. Determining which cases should become `DMException` therefore requires distinguishing DM-originated failures from OpenDream implementation failures.

I initially focused on `DMOpcodeHandlers`, where several cases were straightforward DM errors. I also verified this behavior with faulty DM programs: for example, invalid arithmetic operations and division by zero now produce `OpenDreamShared.DMException` while preserving the existing DM error message and behavior.

## Current change

This PR introduces the shared `DMException` type and applies it to the confirmed DM-facing cases in `DMOpcodeHandlers`.

The intention here is **not** to change how these errors behave, but to give DM-originated runtime failures a distinct exception type.

## Question about the intended design

Before expanding this conversion further, I'd like to confirm the intended architecture.

Once `DMException` exists, should it:

1. Simply replace generic `Exception` at DM-facing failure sites while being handled by the existing runtime exception path?
2. Have dedicated handling somewhere in the runtime, such as catching `DMException` and converting it into a DM-visible runtime error?
3. Be handled differently from ordinary C# exceptions at a higher level?
4. Follow another pattern already used elsewhere in OpenDream?

The distinction matters because there are many additional `throw new Exception(...)` sites, and some are clearly DM-facing while others appear to represent OpenDream/VM invariants. Converting everything mechanically could incorrectly classify internal failures as DM errors.

I'd therefore appreciate guidance on the intended handling and boundary for `DMException` before continuing with the remaining cases.

If the intended approach is confirmed, I think the remaining conversion would be better split into smaller, independently reviewable changes rather than making one very large cross-cutting PR.

### DIFF
--- a/OpenDreamRuntime/AtomManager.cs
+++ b/OpenDreamRuntime/AtomManager.cs
@@ -11,7 +11,6 @@ using OpenDreamRuntime.Resources;
 using OpenDreamShared.Dream;
 using Robust.Shared.Map;
 using Dependency = Robust.Shared.IoC.DependencyAttribute;
-using OpenDreamShared;
 namespace OpenDreamRuntime;
 
 public sealed partial class AtomManager {
@@ -330,12 +329,12 @@ public sealed partial class AtomManager {
                 value.TryGetValueAsInteger(out appearance.MouseOverPointer);
                 break;
             case "appearance":
-                throw new DMException("Cannot assign the appearance var on an appearance");
+                throw new Exception("Cannot assign the appearance var on an appearance");
 
             // These should be handled by the DreamObject if being accessed through that
             case "overlays":
             case "underlays":
-                throw new DMException($"Cannot assign the {varName} var on an appearance");
+                throw new Exception($"Cannot assign the {varName} var on an appearance");
 
             // TODO: filters
             //       It's handled separately by whatever is calling SetAppearanceVar currently
@@ -483,7 +482,7 @@ public sealed partial class AtomManager {
             DreamObjectTurf turf => turf.Appearance,
             DreamObjectMovable movable => movable.SpriteComponent.Appearance!,
             DreamObjectImage image => image.IsMutableAppearance ? AppearanceSystem!.AddAppearance(image.MutableAppearance!, registerAppearance: false) : image.SpriteComponent!.Appearance!,
-            _ => throw new DMException($"Cannot get appearance of {atom}")
+            _ => throw new Exception($"Cannot get appearance of {atom}")
         };
     }
 
@@ -764,7 +763,6 @@ public sealed partial class AtomManager {
 
     [MethodImpl(MethodImplOptions.NoInlining)]
     private static (int X, int Y, int) ThrowCantGetPosition(DreamObjectAtom atom) {
-        throw new DMException($"Cannot get the position of {atom}");
+        throw new Exception($"Cannot get the position of {atom}");
     }
 }
-
--- a/OpenDreamRuntime/AtomManager.cs
+++ b/OpenDreamRuntime/AtomManager.cs
@@ -1,4 +1,4 @@
-﻿using System.Diagnostics.CodeAnalysis;
+using System.Diagnostics.CodeAnalysis;
 using System.Runtime.CompilerServices;
 using JetBrains.Annotations;
 using OpenDreamRuntime.Map;
@@ -11,7 +11,7 @@ using OpenDreamRuntime.Resources;
 using OpenDreamShared.Dream;
 using Robust.Shared.Map;
 using Dependency = Robust.Shared.IoC.DependencyAttribute;
-
+using OpenDreamShared;
 namespace OpenDreamRuntime;
 
 public sealed partial class AtomManager {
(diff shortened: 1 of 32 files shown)


ahmdkaml:
The goal of this PR is not necessarily to be merged as-is, but to demonstrate and discuss a possible workflow for handling DM-originated exceptions.

The proposed workflow has three stages:

Create a shared DMException type so DM-originated errors have a distinct type from internal C# exceptions.
Replace the relevant throw new Exception(...) sites with throw new DMException(...), without changing the underlying behavior or error messages.
Add top-level exception handling that can distinguish between normal, debug, and other execution modes and decide what should be exposed to the user versus what should be logged/debugged.

This PR currently focuses on demonstrating the first two stages. I would like feedback on whether this is the intended direction, particularly on how DMException should be handled at the top level and what information should be shown in normal versus debug mode.

If this direction makes sense, the remaining conversion can be split into smaller, focused PRs rather than attempting to change the entire exception hierarchy in one large change.

wixoaGit:
The introduction of `DMException` is a good start here. I'm not sure how far up you mean by top-level exception handling though, and I wouldn't expect #2685's requested behavior to be as large a change as you say.

Any uncaught exception within executed DM code is handled by [DreamThread's HandleException method](https://github.com/OpenDreamProject/OpenDream/blob/95e45172d733b2e88bc4fb0ba408097df0f34290/OpenDreamRuntime/DreamThread.cs#L484). Here you can check if the exception is of type `DMException`, and exclude the C# stacktrace if the `DEBUG` compile flag isn't present.

ahmdkaml:
> The introduction of `DMException` is a good start here. I'm not sure how far up you mean by top-level exception handling though, and I wouldn't expect #2685's requested behavior to be as large a change as you say.
> 
> Any uncaught exception within executed DM code is handled by [DreamThread's HandleException method](https://github.com/OpenDreamProject/OpenDream/blob/95e45172d733b2e88bc4fb0ba408097df0f34290/OpenDreamRuntime/DreamThread.cs#L484). Here you can check if the exception is of type `DMException`, and exclude the C# stacktrace if the `DEBUG` compile flag isn't present.

thanks for the quick response, I will start working u can expect the change soon.

ahmdkaml:
Hey! I’ve pushed the dmexception-shared branch with the changes to use DMException for the DM-facing runtime errors. I also verified that the project builds successfully and all tests pass.

If there’s anything I missed or anything that needs to be adjusted, I’m happy to do any follow-ups.

ahmdkaml:
all the changes are strightforward dm exceptions I might have missed some but I didn't incorrectly include any

OpenDreamShared/DMException.cs
wixoaGit, line 1: This would be a better fit for the `OpenDreamRuntime` project, the client should never need to use this.

OpenDreamRuntime/Procs/NativeProc.cs
```diff
@@ -30,7 +31,7 @@ public static (string, Dictionary<string, DreamValue>?, List<string>) GetNativeI
                     int intValue => new(intValue),
                     float floatValue => new(floatValue),
                     string stringValue => new(stringValue),
-                    _ => throw new Exception($"Invalid default value {parameterAttribute.DefaultValue}")
+                    _ => throw new DMException($"Invalid default value {parameterAttribute.DefaultValue}")
```
wixoaGit: This one is triggered by an invalid value in a C# method's definition, not by DM code.
```suggestion
                    _ => throw new Exception($"Invalid default value {parameterAttribute.DefaultValue}")
```

OpenDreamRuntime/AtomManager.cs
wixoaGit, line 1: The exceptions in here are much more likely to be triggered by bad C# code, so these should be kept `Exception`.

OpenDreamShared/DMException.cs
```diff
@@ -0,0 +1,7 @@
+using System;
+
+namespace OpenDreamShared;
+
+public sealed class DMException : Exception {
```
wixoaGit: It would be nice to have a quick documentation comment on this explaining how it should be used and its difference in behavior.

wixoaGit (changes requested):

ahmdkaml:
all 4 changes requested are done. thanks for the quick review @wixoaGit

OpenDreamRuntime/DMException.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+
+namespace OpenDreamRuntime;
```
wixoaGit:
```suggestion
namespace OpenDreamRuntime;
```

OpenDreamRuntime/DMException.cs
```diff
@@ -0,0 +1,12 @@
...
+
+/// <summary>
+/// Represents an error caused by invalid DM code execution.
+/// Unlike regular <see cref="Exception"/> instances, this exception is handled
+/// as a runtime error originating from DM code.
+/// </summary>
+public sealed class DMException : Exception {
+    public DMException(string message) : base(message) { }
+}
```
wixoaGit:
```suggestion
public sealed class DMException(string message) : Exception(message);
```

wixoaGit (approved):

wixoaGit:
The failing test is unrelated, I think #2677 made it flaky.

PR #2696, opened 2026-08-12
Fix SQLite database queries

DamianX:
Changed the following in order to make vgstation migrations and preferences actually work.

Made `GetColumn()` 1-indexed, as it is in BYOND.
Implemented `/database/query.Execute(filename)` (as opposed to `Execute(database/db)`)
`Execute()` now returns 1 on success and 0 on failure.
Invalid `Execute()` arguments raise Bad database.
`NextRow()` now returns 1 when it advances to a row and 0 when exhausted.
Re-executing a query closes its prior reader and restarts the result set.

An existing comment mentioned that /icons could be turned into BLOBs in SQLite. That turned out to be true, but BYOND, at least in the latest version, does not seem to allow *retrieving* BLOBs from SQLite.
So:
- File resources and /icon values are stored as BLOBs.
- BLOB values read through DM query APIs return null. I guess, man.


DamianX:
I didn't realize #2547 existed. It got some things right that I did not think about. I'm going to incorporate those changes here.

PR #2708, opened 2026-09-11
`.Find()` for special lists

### LINKED ISSUE #2302: `.Find()` is not implemented on most special list types
See #2300 

for example, `thing.vars.Find()` will fail.

To fix, go to DreamList.cs and implement `public override int FindValue(DreamValue value, int start = 1, int end = 0) {}` for each list that has it marked as unimplemented

ZeWaka:
also implemented `DreamGlobalVars.GetLength()` since my enumeration stuff needs it and it was missing

tested pretty thoroughly in the DM tests, only ones not covered are the 3 client ones (verbs, screen, images)

fixes #2302

### DIFF
--- a/DMCompiler/Compiler/CompilerError.cs
+++ b/DMCompiler/Compiler/CompilerError.cs
@@ -87,6 +87,7 @@ public enum WarningCode {
     // 4000 - 4999 are reserved for runtime configuration. (TODO: Runtime doesn't know about configs yet!)
     ListNegativeSizeException = 4000, // When a list's length is decremented below zero, raise an exception.
     InitialVarOnPrimitiveException = 4001, // initial(foo.var) where foo is a variable containing a non-datum value.
+    ListFindOutOfBoundsException = 4002, // list.Find() with an End past the end of the list, otherwise ignored on some special lists.
 }
 
 public enum ErrorLevel {
@@ -194,6 +195,7 @@ public struct CompilerEmission {
         //4000-4999
         {WarningCode.ListNegativeSizeException, ErrorLevel.Disabled},
         {WarningCode.InitialVarOnPrimitiveException, ErrorLevel.Disabled},
+        {WarningCode.ListFindOutOfBoundsException, ErrorLevel.Disabled},
     };
 
     public CompilerEmission(ErrorLevel level, Location? location, string message) {
--- a/OpenDreamRuntime/Objects/Types/DreamList.cs
+++ b/OpenDreamRuntime/Objects/Types/DreamList.cs
@@ -1,6 +1,7 @@
 using System.Diagnostics;
 using System.Linq;
 using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
 using JetBrains.Annotations;
 using OpenDreamRuntime.Map;
 using OpenDreamRuntime.Procs;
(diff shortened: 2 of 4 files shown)


OpenDreamRuntime/Objects/Types/DreamList.cs
```diff
@@ -259,6 +261,55 @@ public virtual int FindValue(DreamValue value, int start = 1, int end = 0) {
...
+        if (start < 0) // A negative Start doesn't search at all, unlike one past the end
+            return false;
+        if (start == 0)
+            start = 1;
+
+        return start < end;
+    }
+
+    /// <returns>The 1-indexed position of the value, or 0 if it isn't in the [start, end) range</returns>
+    protected static int FindValueInSpan<T>(ReadOnlySpan<T> values, T value, int start, int end) where T : IEquatable<T> {
```
ZeWaka, line 292: only used a handful of times but figured it was best to abstract this out

OpenDreamRuntime/Objects/Types/DreamList.cs
```diff
@@ -259,6 +261,55 @@ public virtual int FindValue(DreamValue value, int start = 1, int end = 0) {
         return 0;
     }
 
+    /// <summary>
+    /// Turns Find()'s Start/End into a [start, end) range of 1-indexed positions.
+    /// End 0 is just past the last element, Start 0 is 1, a negative Start finds nothing.
```
ZeWaka, line 266: unsure if negative start should also be a `ListFindOutOfBoundsException`

OpenDreamRuntime/Objects/Types/DreamList.cs
```diff
@@ -1087,7 +1171,12 @@ public override int GetLength() {
     }
 
     public override int FindValue(DreamValue value, int start = 1, int end = 0) {
-        throw new NotImplementedException($".Find() is not yet implemented on {GetType()}");
+        if (!NormalizeFindRange(_visContents.Count, ref start, ref end))
+            return 0;
+        if (!value.TryGetValueAsDreamObject<DreamObjectAtom>(out var atom))
+            return 0;
+
+        return _visContents.IndexOf(atom, start - 1, end - start) + 1;
```
ZeWaka, line 1179: indexof is 0-indexed

OpenDreamRuntime/Objects/Types/DreamList.cs
```diff
@@ -259,6 +261,55 @@
         return 0;
     }
 
+    /// <summary>
+    /// Turns Find()'s Start/End into a [start, end) range of 1-indexed positions.
+    /// End 0 is just past the last element, Start 0 is 1, a negative Start finds nothing.
+    /// </summary>
+    /// <param name="lenientEnd">Whether an out-of-range End searches to the end instead of erroring</param>
+    /// <returns>false if the range is empty</returns>
+    protected bool NormalizeFindRange(int length, ref int start, ref int end, bool lenientEnd = false) {
```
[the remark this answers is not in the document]
ZeWaka: im good dog

OpenDreamRuntime/Objects/Types/DreamList.cs
ike709: I think it should. I can think of a lot more reasons for catching unintended side effects from the no-op behavior than reasons for no-op to be desirable behavior. If people have a usecase to want to have `Find()` do nothing if `start<0` then they can just insert a guard before the `Find()` to check the start value and avoid calling `Find()` at all.

DMCompiler/Compiler/CompilerError.cs
```diff
@@ -87,6 +87,7 @@ public enum WarningCode {
     // 4000 - 4999 are reserved for runtime configuration. (TODO: Runtime doesn't know about configs yet!)
     ListNegativeSizeException = 4000, // When a list's length is decremented below zero, raise an exception.
     InitialVarOnPrimitiveException = 4001, // initial(foo.var) where foo is a variable containing a non-datum value.
+    ListFindOutOfBoundsException = 4002, // list.Find() with an End past the end of the list, otherwise ignored on some special lists.
```
ike709, line 90: If you do change the behavior please update the comment.

PR #2710, opened 2026-09-18
Fix mis-parsing `else:` in `switch()` statements

ike709:
Knocks an error off FTL13.


wixoaGit:
~~A test & fix for a normal if/else would be good too, a switch's `else` goes down a very different code path.~~ Nevermind, didn't expect us to already support that.

DMCompiler/Compiler/DM/DMParser.cs
```diff
@@ -1547,6 +1547,7 @@ private DMASTProcStatementSwitch.SwitchCase[] SwitchInner() {
 
                 return new DMASTProcStatementSwitch.SwitchCaseValues(expressions.ToArray(), body);
             } else if (Check(TokenType.DM_Else)) {
+                Check(TokenType.DM_Colon); // Someone wrote "else:" instead of "else"
```
wixoaGit, line 1551: There can be whitespace before.
```suggestion
                Whitespace();
                Check(TokenType.DM_Colon); // Someone wrote "else:" instead of "else"
```

wixoaGit (approved):

